Issue #1, opened 2019-01-14
Establish the RFC process


foolip:
# Summary

With the [wpt core team](https://lists.w3.org/Archives/Public/public-test-infra/2019JanMar/0000.html) now in place, this repo will hold the process for making substantive changes. This first RFC is about establishing the process itself, to give a chance for comments on that before beginning to use it.

# Motivation

https://github.com/web-platform-tests/wpt/issues/11473 is the background for how the core team was established, and why we need a governance project. 

# Details

Please review these files:
- https://github.com/web-platform-tests/rfcs/blob/master/README.md
- https://github.com/web-platform-tests/rfcs/blob/master/ISSUE_TEMPLATE.md

jgraham:
I think 1 week for an initial decision may be too short? Possibly it should be 2 weeks without any signal that times out the process, but with the possibility of wide agreement in less time? I'm not sure.

foolip:
It's certainly not uncommon for someone to be away for a week so that they'd miss RFCs if the period is just one week, but I wonder if we can rely on others to spot when something is likely controversial and extend the comment period for those cases, rather than shortening it for the trivial?

foolip:
I just added "As a last resort, the core team may decide using a 2/3 majority vote." to README.md based on previous discussion.

foolip:
@web-platform-tests/wpt-core-team please review.

zcorpan:
If something is non-controversial and trivial, does it need to use the RFC process?

I think 2 weeks from the time the issue is created is reasonable, and people can ask to extend if they need more time.

foolip:
> If something is non-controversial and trivial, does it need to use the RFC process?

The idea is that the conditions in https://github.com/web-platform-tests/rfcs/blob/master/README.md#when-to-use-the-rfc-process should filter out the non-controversial and trivial.

gsnedders:
> The idea is that the conditions in https://github.com/web-platform-tests/rfcs/blob/master/README.md#when-to-use-the-rfc-process should filter out the non-controversial and trivial.

I think that:

 * The "substantial changes" comment should be under the "When to use the RFC process" heading.
 * Some examples of what *doesn't* need an RFC would be useful.

foolip:
[...]

PR #2, opened 2019-01-15
Expand on "When to use the RFC process"

foolip:
Addresses https://github.com/web-platform-tests/rfcs/issues/1#issuecomment-454103641.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,14 +5,12 @@
 Most changes in web-platform-tests can be made directly in pull requests
 following the usual review process.
 
-The RFC (request for comments) process is used to request wider review for
-substantial changes, which might impact other stakeholders or users of
-web-platform-tests.
+This RFC (request for comments) process is used to request wider review.
 
 ## When to use the RFC process
 
-A few examples where the process is
-likely to be useful:
+Use this process for substantial changes which might impact other stakeholders
+or users of web-platform-tests. Examples of where it is likely to be useful:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,6 +20,13 @@ likely to be useful:
...
  - Changing review workflows or policies.
 
+Cases where the RFC process need *not* be used:
+
+ - Introducing a new top-level directory for a new specification.
+ - Minor changes in behavior in where all call sites are known and accounted
+   for.
+ - Behavior-preserving refactoring with a low risk of regressions.
+
 As a rule of thumb, please use the RFC process when it would be useful and err
```
jgraham: I don't really understand this rule of thumb. It doesn't seem to be a simple but imprecise test one can apply to figure out the correct decision (i.e. a "rule of thumb"), but a general request to use the process when appropriate. 

The correct rule of thumb is probably closer to "please use the RFC process for changes likely to have an impact on multiple users of wpt".

foolip: This section leads with "Use this process for substantial changes which might impact other stakeholders or users of web-platform-tests", that's roughly the same as "likely to have an impact on multiple users of wpt".

You're right that this isn't a rule of thumb, I've pushed an additional commit that I think is an improvement, but please review: https://github.com/web-platform-tests/rfcs/pull/2/commits/56fc69c820798cb419292f43aab9da61565314ec

PR #11, opened 2019-02-18
Delete ISSUE_TEMPLATE.md

foolip:


gsnedders:
Do we want to keep this around, maybe just rename it to `RFC_TEMPLATE` or something?

foolip:
@jgraham did you see @gsnedders' question?

jgraham:
I'm not sure this matches the description in the README at this point, and I'm not sure if anyone would find the template file. I suggest we recreate something later or add a simple script to create a new RFC if people usually work by cloning the repo rather than just editing in the GH UI.

PR #19, opened 2019-03-20
Always use the working copy for the manifest

gsnedders:
[Rendered](https://github.com/gsnedders/rfcs/blob/manifest-working-copy/rfcs/manifest-working-copy.md)

(so GitHub picks it up: references https://github.com/web-platform-tests/wpt/issues/7206)

### DIFF
--- a/rfcs/manifest-working-copy.md
+++ b/rfcs/manifest-working-copy.md
@@ -1,0 +1,33 @@
+# Always use the working copy for the manifest
+
+## Summary
+
+Instead of defaulting to building a manifest using what's in the git HEAD,
+unconditionally use what's in the working copy.
+
+## Details
+
+`./wpt manifest` supports two modes, the default (based on the `git` tree at
+HEAD) and `--work` (which uses the working copy). We supported the former mode
+mostly as a trivial performance optimization, though I believe it is now
+slower.
+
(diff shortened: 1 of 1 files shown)


jgraham:
This lgtm.

Can't we conspire to add the CSS build output to the ignore rules in the root?

gsnedders:
> Can't we conspire to add the CSS build output to the ignore rules in the root?

I don't think there's any stopping us from merging all the non-root non-third-party `.gitignore` files, but it still won't behave as git does (as it won't obey `.git/info/excludes` or `core.excludesFile`, nor will we handle tracked files correctly), which still isn't ideal.

It's also potentially surprising in the case of Chromium or Servo where it won't obey top-level `.gitignore` files (of course, in mozilla-central it's even harder, given we'd need some way to process `.hgignore`).

Hexcles:
FWIW, Chromium always uses `--work`: https://cs.chromium.org/chromium/src/third_party/blink/tools/blinkpy/w3c/wpt_manifest.py?l=209&rcl=b7338025b4a704fc08ee2348083fadfac4449a5e

The difference in default behaviours has caused some confusion when Chromium engineers occasionally use the upstream `./wpt` toolchain.

gsnedders:
No substantive disagreement, so considered accepted after 1 week

PR #29, opened 2019-09-16
testharness.js: Replace assert_throws and promise_rejects with more fine-grained APIs

Ms2ger:


jgraham:
This seems reasonable to me, but I'd like to check that e.g. @bzbarsky and @annevk are on board with the proposed changes.

bzbarsky:
I'm very much on board with this.  So much so that I filed the issue requesting this change (<https://github.com/web-platform-tests/wpt/issues/11726>) and wrote the testharness changes to implement it (<https://github.com/web-platform-tests/wpt/pull/19054>).

jgraham:
> I'm very much on board with this. So much so that I filed the issue requesting this change and wrote the testharness changes to implement it

Oh, hah I missed that bit of context.

Unrelated to that, in the risk section of the RFC it says:

> Updates to testharness.js haven't always been made together with the test updates in all browser engine repos.

We make those updates atomically in gecko and servo, so from our point of view no additional coordination is required to land this kind of change.

bzbarsky:
So what is the next step here?  Are we waiting for someone specific to look things over or make a call?

jgraham:
The process is that we allow a week for objections.

rfcs/assert_throws.md
```diff
@@ -0,0 +1,45 @@
...
+
+Six new APIs are proposed to handle each of the cases separately:
+
+- `promise_rejects_dom` and `assert_throws_dom`, called with a string representing a `DOMException` (as above);
+- `promise_rejects_js` and `assert_throws_js`, called with the Error constructor (e.g. `TypeError`);
+- `assert_throws_exactly` and `promise_rejects_exactly`, called with the expected object and using `Object.is`.
+
+The new APIs will be added first, in order to allow a well-organized transition.
+Once this is done and browser vendors have updated their copies of `testharness.js`, we can start working on a (mostly automated) conversion of the tests in wpt.
+The old API will only be removed once all callers have been removed from wpt and browser vendors have had a chance to update any callers in proprietary tests.
```
foolip: Is anyone signed up to do this work in each engine, or is this an ask for us to do it? Or would you do some engines but not all?

Looks like we have about 600 tests using `assert_throws` and `promise_rejects` in Chromium, about 300 of each.

foolip:
@youennf can you check how this would impact WebKit?

rfcs/assert_throws.md
Ms2ger: Not yet. I think it'd be best if the browser vendors handled this themselves; we can see how to deal if that doesn't happen.

foolip: If the timeline is quite lax I think we can agree to handle it on the Chromium side. Can you get a similar commitment for the other repos so that this doesn't get stuck in a half-transitioned state for months and months?

Ms2ger: @jdm for Servo
@jgraham for Gecko
@youennf for WebKit (if you have proprietary tests that use upstream th.js at all)

Would y'all be willing to deal with this? I imagine we could set a deadline about two months after the RFC and https://github.com/web-platform-tests/wpt/pull/19054 land.

bzbarsky: I just checked, and if I am not misreading things Gecko has 2 uses of `assert_throws` and 0 uses of promise_rejects in our non-upstreamed tests.  I think we can commit to fixing that very expeditiously.  ;)

foolip: Weird. @jgraham is it really the case that there's so little internal use of testharness.js in Gecko? What do people do with tests for a spec that they don't want to upstream yet but mean to in the future?

jgraham: Yes, I got the same results as @bzbarsky

I think this is a combination of two things:
* Gecko does a lot less work on speculative features
* We've been less successful at getting people to write wpt tests instead of gecko-specific tests unless they're specifically trying to contribute to wpt.

rfcs/assert_throws.md
```diff
@@ -0,0 +1,45 @@
...
+
+```js
+assert_throws(new TypeError(), () => { throw new DOMException("TypeError") });
+```
+
+Six new APIs are proposed to handle each of the cases separately:
+
+- `promise_rejects_dom` and `assert_throws_dom`, called with a string representing a `DOMException` (as above);
+- `promise_rejects_js` and `assert_throws_js`, called with the Error constructor (e.g. `TypeError`);
+- `assert_throws_exactly` and `promise_rejects_exactly`, called with the expected object and using `Object.is`.
```
bzbarsky: "expected value".  It doesn't have to be an object.

foolip: `Object.is` caught my eye here. I don't think we use this elsewhere and yet the benefits would be the same here and elsewhere. I don't think -0 and NaN are more common values to throw, so I wonder if we should use `===` and add in support of `Object.is` to various asserts separately?

bzbarsky: I'd be fine with using `===`.

Ms2ger: Or, actually, `assert_equals` for consistency?

foolip: Looks like `assert_class_string` uses it internally, but most asserts seemingly don't reuse other asserts like this. I think it's because the resulting error message then isn't clear, so this should be avoided?

jgraham: You can catch the error and rethrow with whatever message you like. That's a bit more verbose of course, but `assert_equals` does take some care around edge cases.

foolip: I sent https://github.com/web-platform-tests/wpt/pull/19219 using `same_value` for `assert_class_string`, see if you like that?

If so I guess let's just use `same_value` here too?

bzbarsky: We can do that, but `same_value` just implements the `Object.is` semantics (but presumably predates it), right?  So maybe we should consider removing `same_value` altogether and using `Object.is` consistently.  In any case, I'm happy to do whatever here, but I'd really like to get this landed ASAP; at this point I have quite a large number of other changes already written and reviewed but blocked on this...

jgraham:
I think this has been a week and there aren't substantive objections.

I think it's reasonable to land this with `same_value` and consider moving to `Object.is` as a followup; traditionally we've been slower to adopt new ES features in `testharness.js` since it's occasionally used to test relatively outdated browsers; I don't know if this makes the cut or not, but let's not block this issue on that discussion.

rfcs/assert_throws.md
foolip: `same_value` is indeed very old, I followed it back to https://github.com/w3c/testharness.js/commit/1780607fc4582b65b7fcdded3e39e79e95c8d915 in 2012 but that was a copy.

Replacing it with `Object.is` would be cool, but not necessary as part of this.

jgraham: For historical interest, despite the fact that I'd forgotten `same_value` existed in the comment above, I'm pretty sure it was code I wrote at Opera, informed by the experience of working on Carakan (a rewrite of the js engine), where the edge cases it handles are very important to get right. So I predict it was in the earliest versions of `testharness.js` when ES3 was still the contemporaneous version.

rfcs/assert_throws.md
bzbarsky: >  I think we can commit to fixing that very expeditiously

https://bugzilla.mozilla.org/show_bug.cgi?id=1607307 tracks that.

foolip: That was fast indeed! :)

bzbarsky: Well, terrible given the actual amount of work involved, but.... ;)

foolip:
@Ms2ger @bzbarsky what's the status of implementing this RFC? Will we need to audit use of `assert_throws` and `promise_rejects` in Chromium soon?

cc @stephenmcgruer

bzbarsky:
The new functions are landed.  New tests should use them.  I'm not sure there's any specific plan or timeline for converting the existing shared-test uses so far.

bzbarsky:
That said, there are ~4300 `assert_throws` uses in wpt so far.  Of those, I think ~2100 are very easy to rewrite automatically to `assert_throws_js` without even messing up indentation and another ~72 are auto-rewritable with some indentation messup.  I'm going to give that a show and see how things look....

Another ~1400 should be auto-rewritable to `assert_throws_dom`,  I think.

The remaining ones, (~700 for those keeping count), I guess we see what they look like.

bzbarsky:
I'm going to take a stab at the auto-rewriting and fixing the Gecko bugs it exposes... it was on my list anyway.

foolip:
Thanks @bzbarsky! When the time to remove `assert_throws` and `promise_rejects` is closer, can you update this thread?

stephenmcgruer:
Unless I'm counting wrong, the number of `promise_rejects` has fallen substantially in Chromium from the 300 reported above.

["assert_throws("](https://cs.chromium.org/search/?q=%22assert_throws(%22+file:third_party/blink/web_tests+-file:external/wpt&sq=package:chromium&type=cs ) - ~300

["promise_rejects("](https://cs.chromium.org/search/?q=%22promise_rejects(%22+file:third_party/blink/web_tests+-file:external/wpt&sq=package:chromium&type=cs ) - ~70

bzbarsky:
@foolip will do!  I haven't even looked at `promise_rejects` yet.  ;)

bzbarsky:
Fwiw, it would also help if people stopped adding new `assert_throws`.  Just in the last few days, since I wrote some patches to remove a bunch, three changesets have added more...

youennf:
> @youennf can you check how this would impact WebKit?

We have some WebKit internal tests using assert_throws/promise_rejects.
Something like 50 files at least.
I guess the plan is to keep assert_throws/promise_rejects support in testharness.js for a while but make the linter forbid its use?

foolip:
Adding a lint for it would cause people to notice that there are new forms that they should use instead. But there are still over 1000 files matching `assert_throws[^_]` so maybe it would be a good idea to get that down to a smaller number first.

zcorpan:
What can be done sooner is to update the documentation. https://github.com/web-platform-tests/wpt/issues/21288

jgraham:
I'm not sure what the specific disadvantage of having a lint with a lot of existing cases is; maybe some performance penalty. In any case I created a PR: https://github.com/web-platform-tests/wpt/pull/21294

stephenmcgruer:
I agree that it's reasonable to land a lint step, but looking at the PR I think we should have some documentation to link people to before landing it. `assert_throws_js` and `assert_throws_dom` are tricky to understand at face value, and currently there's nothing in https://web-platform-tests.org/ to explain to me (as a test writer) which I should be using.

bzbarsky:
Fwiw, I have patches to get the number of `assert_throws` in the test suite down to ~200.  Hoping to get those up for review today or so.

jgraham:
I added some basic docs changes to my PR.

bzbarsky:
I created https://github.com/web-platform-tests/wpt/pull/21350 for the main mass-replacements of `assert_throws`.

stephenmcgruer:
With bzbarsky's recent changes, we're down to 121 counts of `assert_throws(` in WPT.

I have been fielding some light push-back from folks as these changes filter through. Some folks are finding them confusing in terms of which function to use, what they actually mean, etc. I don't think any action is needed yet, but this is on my mind as I look at [`assert_object_equals`](https://github.com/web-platform-tests/wpt/issues/2033) too.

bzbarsky:
> With bzbarsky's recent changes, we're down to 121 counts of assert_throws( in WPT.

Hmm.  It's 51 in my local tree on master, as far as I can tell. And 41 once https://github.com/web-platform-tests/wpt/pull/21388 merges, I think.

> Some folks are finding them confusing in terms of which function to use, what they actually mean, etc.

Do you have a sense about whether the documentation for the functions in testharness.js is unclear or whether they had not seen those docs and only had the function names to go on?

stephenmcgruer:
> Hmm. It's 51 in my local tree on master, as far as I can tell. And 41 once web-platform-tests/wpt#21388 merges, I think.

Ah, I concur. I had included `.git` by mistake in my naive grep. 51 once that is excluded.

> Do you have a sense about whether the documentation for the functions in testharness.js is unclear or whether they had not seen those docs and only had the function names to go on?

So I would guess nobody reads the source. We hope they might read https://web-platform-tests.org/, but that doesn't yet have documentation. That said, no, I'm afraid I don't have a good sense of whether it would make sense to them if they had read the documentation or not (sorry, not very useful :()

bzbarsky:
I posted https://github.com/web-platform-tests/wpt/pull/21582 to remove a bunch of the `promise_rejects` calls with `promise_rejects_js`.  Once that lands, I have similar changes for `promise_rejects_dom`.

Still working on mopping up the remaining `assert_throws` bits too.

bzbarsky:
OK, at this point I have all of `assert_throws` and `promise_rejects` removed locally.  Just waiting on various open PRs to merge so I can put up the final cleanup PRs...

bzbarsky:
This is now completely done.

stephenmcgruer:
FYI as of https://crbug.com/1051932#c22 this is now (finally!) done in Chromium downstream too.

PR #33, opened 2019-10-09
Add RFC for `crash` test type

jgraham:
Add a test type that just ensures that a page can be loaded without error.

### DIFF
--- a/rfcs/load_test.md
+++ b/rfcs/load_test.md
@@ -1,0 +1,46 @@
+# RFC 33: Crash Tests
+
+## Summary
+
+Add a test type in which the test document is simply loaded. The test
+passes if this occurs without encountering a fatal error e.g. a crash.
+
+## Details
+
+Load Tests are static tests typically used for cases where browsers
+are known to have a bug such as a crash or a leak. They differ from
+other tests such as reftests and testharness tests in that there is no
+pass condition other than the page loading successfully. In vendor
+infrastructure these tests may also be run with additional
+instrumentation e.g. address sanitizing, making the pass condition
+more stringent. The lack of an explicit pass condition makes it easier
+to write these tests than tests where a detailed understanding of the
+correct behaviour is required, making them suitable for autogeneration
+e.g. as the output of a fuzzing tool.
+
+The infrastructure changes are as follows:
+
+* A new test type `crash` recognised by the manifest and assigned to
+  any test file with a name containing `-crashtest` immediately before
+  the extension or in a subdirectory called `crashtests`.
+
+* A new executor type with the following behaviour:
+  - Load the test URL
+  - If there is no `class=wait` on the root element, mark the test as
+    passed when the load is complete, all fonts have finished loading,
+    and the rendering is stable.
+ - If `class=wait` appears on the root element, mark the test as
+   complete once the previous conditions are met and that class has been
+   removed.
+
+## Risks
+
+Adding this test type may encourage test authors to write crashtests
+for cases where a more useful cross-browser test would have a real
+pass condition.
+
+Crashes and other issues caught by these tests may not be sufficiently
+common between browsers to make sharing these tests a good use of
+resources. However there's data showing cases where a tests that crash
+one browser also crash others. Therefore this concern seems less
+serious than it may at first appear.
--- a/rfcs/load_test.md
+++ b/rfcs/load_test.md
@@ -21,7 +21,7 @@ e.g. as the output of a fuzzing tool.
 The infrastructure changes are as follows:
 
 * A new test type `crash` recognised by the manifest and assigned to
-  any test file with a name containing `-crashtest` immediately before
+  any test file with a name containing `-crash` immediately before
   the extension or in a subdirectory called `crashtests`.
 
 * A new executor type with the following behaviour:


jgraham:
@emilio I think you were interested in this. If you have feedback it would be great, particularly if you have evidence that sharing load-type tests (e.g. gecko crashtests) between engines is effective at finding issues.

rfcs/load_test.md
```diff
@@ -0,0 +1,45 @@
...
+* A new test type `load` recognised by the manifest and assigned to
+  any test file with a name containing `-load` or `-loadtest`
+  immediately before the extension (the former will require renaming
+  some existing tests) or in a subdirectory called `loadtests` (the
+  exact set of options to choose here is tdb based on feedback).
+
+* A new executor type with the following behaviour:
+  - Load the test URL
+  - If there is no `class=wait` on the root element, mark the test as
+    passed when the load is complete (in the `load` event handler)
```
zcorpan: I believe reftests have logic to also wait for fonts and such to load. Should these use the same logic as reftests?

jgraham: Yes, that makes sense I think. I'll update once there's more feedback.

rfcs/load_test.md
```diff
@@ -0,0 +1,45 @@
...
+instrumentation e.g. address sanitizing, making the pass condition
+more stringent. The lack of an explicit pass condition makes it easier
+to write these tests than tests where a detailed understanding of the
+correct behaviour is required, making them suitable for autogeneration
+e.g. as the output of a fuzzing tool.
+
+The infrastructure changes are as follows:
+
+* A new test type `load` recognised by the manifest and assigned to
+  any test file with a name containing `-load` or `-loadtest`
```
zcorpan: `-load` is too generic and is already used by existing tests e.g.
```
./idle-detection/idle-detection-allowed-by-feature-policy-attribute-redirect-on-load.https.sub.html
./css/css-paint-api/valid-image-after-load.https.html
```
`-loadtest` suffix or `loadtests` directory seems OK.

jugglinmike: Since "load" can be taken to mean "encumber," people may expect these are stress tests. @zcorpan mentioned that "crash" is already conventional, so maybe that represents a preference in terminology. Would there be a problem with updating those tests?

jugglinmike:
> Crashes and other issues caught by these tests may not be sufficiently common between browsers to make sharing these tests a good use of resources.

This was my concern while reading the proposal, so I'm glad you included it! Personally, I'm pessimistic that tests like these do provide much value between implementations. I'm no implementer, though, so I'm very curious about what @emilio has to say on the subject.

emilio:
I have actually some data on this, need to write it down, but this are the results thus far (I haven't filed all bugs yet) from running Gecko's crashtest suite in WebKit and Chrome, and Chrome's crashtests in Gecko:

 * https://bugs.chromium.org/p/chromium/issues/detail?id=1009830
 * https://bugs.chromium.org/p/chromium/issues/detail?id=1009827
 * https://bugs.chromium.org/p/chromium/issues/detail?id=1009825
 * https://bugs.chromium.org/p/chromium/issues/detail?id=1009821
 * https://bugs.chromium.org/p/chromium/issues/detail?id=1009817
 * https://bugs.chromium.org/p/chromium/issues/detail?id=1009806
 * https://bugzilla.mozilla.org/show_bug.cgi?id=1585226
 * https://bugzilla.mozilla.org/show_bug.cgi?id=1585303
 * https://bugs.webkit.org/show_bug.cgi?id=202803
 * https://bugs.webkit.org/show_bug.cgi?id=202804
 * https://bugs.webkit.org/show_bug.cgi?id=202805
 * https://bugs.webkit.org/show_bug.cgi?id=202807
 * https://bugs.webkit.org/show_bug.cgi?id=202809
 * https://bugs.webkit.org/show_bug.cgi?id=202811
 * https://bugs.webkit.org/show_bug.cgi?id=202812

And as I said I'm still not done running tests and filing bugs. Pending stuff:

 * Running Chromium's crashtests on WebKit.
 * Running WebKit's crashtests in Gecko.
 * Running WebKit's crashtests in Chromium.

So yeah, I do think they're useful :)

emilio:
And this was a _very_ simple test runner, there's tons of test-cases that I haven't ran or what not.

gsnedders:
I'd rather call them crash tests, given that's what they're called everywhere else.

jgraham:
The reason for not calingthem crashtests was that crashing is only one possible failure mode; they could also fail with an assert or sanitizer failure or similar in vendor infrastructure. But I'm sympathetic to the idea that "load" can be confusing and there is precedent here.

gsnedders:
I think that's an overly pedantic viewpoint to take when there's prior art in all vendors here already, even if the name is slightly confusing. (Just make all assertions and saniziter failures crashes, pff!)

rfcs/load_test.md
```diff
@@ -0,0 +1,46 @@
+# RFC 33: Crash Tests
+
+## Summary
+
+Add a test type in which the test document is simply loaded. The test
+passes if this occurs without encountering a fatal error e.g. a crash.
```
Ms2ger: Does it pass if it encounters a non-fatal error, such as a leak?

rfcs/load_test.md
```diff
@@ -0,0 +1,46 @@
...
+* A new test type `crash` recognised by the manifest and assigned to
+  any test file with a name containing `-crashtest` immediately before
+  the extension or in a subdirectory called `crashtests`.
+
+* A new executor type with the following behaviour:
+  - Load the test URL
+  - If there is no `class=wait` on the root element, mark the test as
+    passed when the load is complete, all fonts have finished loading,
+    and the rendering is stable.
+ - If `class=wait` appears on the root element, mark the test as
```
Ms2ger: Fix indentation

rfcs/load_test.md
```diff
@@ -0,0 +1,46 @@
...
+more stringent. The lack of an explicit pass condition makes it easier
+to write these tests than tests where a detailed understanding of the
+correct behaviour is required, making them suitable for autogeneration
+e.g. as the output of a fuzzing tool.
+
+The infrastructure changes are as follows:
+
+* A new test type `crash` recognised by the manifest and assigned to
+  any test file with a name containing `-crashtest` immediately before
+  the extension or in a subdirectory called `crashtests`.
```
gsnedders: I think calling this `crashtest` goes against the pattern of `manual` and `visual` here.

The prior art here is `-crash` being used in as a filename suffix in WebKit, and `crashtests` as a directory (though combined with a `crashtests.list` manifest, rather than simply considering *everything* in the directory a test).

rfcs/load_test.md
```diff
@@ -0,0 +1,46 @@
...
+
+* A new executor type with the following behaviour:
+  - Load the test URL
+  - If there is no `class=wait` on the root element, mark the test as
```
gsnedders: Will we want to generalize `class=wait` to reftests in future?

rfcs/load_test.md
jgraham: `-crash` turns out to already be used a lot, possibly for things that were converted from crashtests to other types.

Note that `foo/crashtests/support/` still allows including external resources in tests without requiring an explicit manifest.

jgraham:
https://github.com/web-platform-tests/wpt/pull/20017 is a draft PR for this RFC. It's not done yet (hence draft), but what's there seems to work.

rfcs/load_test.md
foolip: How much is it used? If there are less than 100 then fixing/renaming them would free up this good name for use.

jgraham:
```
jgraham@flitwick:~/develop/web-platform-tests$ find . -name '*-crash.*'
./2dcontext/wide-gamut-canvas/imagedata-no-color-settings-crash.html
./payment-request/payment-request-constructor-crash.https.html
./dom/nodes/remove-and-adopt-crash.html
./html/semantics/text-level-semantics/the-ruby-element/rt-without-ruby-crash.html
./html/semantics/interactive-elements/the-details-element/display-table-with-rt-crash.html
./html/semantics/interactive-elements/the-summary-element/display-table-with-rt-crash.html
./html/semantics/forms/the-input-element/time-datalist-crash.html
./html/semantics/forms/the-input-element/type-change-file-to-text-crash.html
./html/semantics/embedded-content/the-object-element/block-object-with-ruby-crash.html
./editing/run/first-letter-crossing-engine-boundary-crash.html
./svg/svg-in-svg/svg-in-svg-circular-filter-reference-crash.html
./svg/extensibility/foreignObject/foreign-object-under-defs-crash.html
./svg/extensibility/foreignObject/foreign-object-circular-filter-reference-crash.html
./svg/extensibility/foreignObject/foreign-object-under-clip-path-crash.html
./svg/pservers/pattern-with-invalid-base-cloned-crash.html
./css/css-pseudo/first-letter-of-html-root-crash.html
./css/css-pseudo/before-in-display-none-crash.html
./css/css-pseudo/first-letter-crash.html
./css/css-pseudo/first-line-first-letter-insert-crash.html
./css/css-animations/keyframes-remove-documentElement-crash.html
./css/css-position/position-absolute-in-inline-crash.html
./css/css-images/infinite-radial-gradient-crash.html
./css/css-images/gradient-crash.html
./css/css-tables/visibility-collapse-rowspan-crash.html
./css/css-tables/visibility-collapse-colspan-crash.html
./css/css-typed-om/set-var-reference-crash.html
./css/css-writing-modes/bidi-inline-fragment-crash.html
./css/css-overflow/shrink-to-fit-auto-overflow-relayout-crash.html
./css/css-overflow/outline-with-opacity-crash.html
./css/css-shapes/shape-outside/supported-shapes/inset/shape-outside-inset-crash.html
./css/css-shapes/shape-outside/supported-shapes/polygon/shape-outside-polygon-crash.html
./css/CSS2/floats-clear/adjoining-float-new-fc-crash.html
./css/CSS2/floats/floats-saturated-position-crash.html
./css/CSS2/linebox/video-needs-layout-crash.html
./css/css-multicol/with-custom-layout-on-same-element-crash.https.html
./css/css-multicol/extremely-tall-multicol-with-extremely-tall-child-crash.html
./css/css-multicol/abspos-in-multicol-with-spanner-crash.html
./css/css-break/break-before-with-no-fragmentation-crash.html
./css/css-text/overflow-wrap/overflow-wrap-break-word-long-crash.html
./css/css-text/overflow-wrap/overflow-wrap-break-word-white-space-crash.html
./css/css-text/white-space/nowrap-wbr-and-space-crash.html
./css/css-text/white-space/pre-line-br-with-whitespace-child-crash.html
./css/css-text/white-space/pre-with-whitespace-crash.html
./css/css-text/text-indent/text-indent-long-line-crash.html
./quirks/table-replaced-descendant-percentage-height-crash.html
./portals/portals-no-frame-crash.html
```

rfcs/load_test.md
jgraham: That seems like a possibility, but I don't think it affects this PR.

rfcs/load_test.md
jgraham: I think that's a browser-specific consideration since leak checking is a browser-specific feature at this point (for gecko at least we can only detect leaks per browser invocation, not per test, so they have to be handled separately).

foolip:
My preference is for `-crash`.

rfcs/load_test.md
gsnedders: So:

We currently have 46 `'*-crash.*'` files. 42 of these include testharness.js, the other 4 are reftests.

Roughly half of the testharness.js tests are already `test(()=>{})` tests. Only 4 tests actually call any assert function.

As such, that's only eight tests that need renamed.

jgraham:
OK changed to `-crash` and `crashtests`. So I don't think there are more remaining issues and will merge.

Hexcles:
Doing some archeology here, I wonder when we changed the wait class from `wait` to `test-wait`? @jgraham

PR #34, opened 2019-10-29
Add TestRendered event

jgraham:


rfcs/reftest_ready.md
```diff
@@ -0,0 +1,15 @@
+# RFC 34: `reftests`: Fire event when painting is done
+
+## Summary
+
+Fire a `ReftestReady` event when layout, fonts and paiting are complete. This coresponds to the point at which the screenshot will be taken in the absence of `reftest-wait`.
+
+## Details
+
+Reftests which want to check for behaviour in the face of dynamic changes need to know when those changes can be made with a complete initial state. This corresponds to the time that a screenshot would be taken in the case that the `reftest-wait` attribute is not present. Firing an event at this time allows the test to schedule dynamic changes and then remove the `reftest-wait` attribute to triger a screenshot without concern that the changes are batched into the intial render. For want of a better name the event fired will be called `ReftestReady`. This corresponds to the gecko `MozReftestInvalidate event. Providing this event is expected to unblock upstreaming some more Gecko reftests.
```
zcorpan:
```suggestion
Reftests which want to check for behaviour in the face of dynamic changes need to know when those changes can be made with a complete initial state. This corresponds to the time that a screenshot would be taken in the case that the `reftest-wait` attribute is not present. Firing an event at this time allows the test to schedule dynamic changes and then remove the `reftest-wait` attribute to triger a screenshot without concern that the changes are batched into the intial render. For want of a better name the event fired will be called `ReftestReady`. This corresponds to the gecko `MozReftestInvalidate` event. Providing this event is expected to unblock upstreaming some more Gecko reftests.
```

Ms2ger:
Is this something that can be implemented in a browser-agnostic way, or do we need browser-internal support? If the latter, would it be hard to implement in non-Gecko browsers?

jgraham:
I think it's sort of a moot point whether we also fire the event in the case where there isn't a `reftest-wait`; although you could do something in response to the event before the screenshot was taken there would be no guarantee that the changes would be reflected in the rendering before the screenshot, so any test doing that would be buggy. But sure, in practice I think we won't fire this except when `reftest-wait` is present to stop people doing the buggy thing. I'll update the PR.

> Is there a need for an event like this for load tests?

Good question. It looks like there are some users of this in m-c: https://searchfox.org/mozilla-central/search?q=MozReftestInvalidate&path=crashtest

jgraham:
I updated the proposal to rename the event `WptTestRendered` which seems like a better name if we're also going to use it for load tests. I also added the detailed steps and prepared a [patch](https://treeherder.mozilla.org/#/jobs?repo=try&revision=f0b4dd92a06595bd9091291ae232afee98f3531b) which implements this for webdriver and marionette (that uses an older iteration of the name without the `Wpt` prefix).

foolip:
Given the capitalization name conflicts with the web platform are virtually impossible, I'd like it better without the `Wpt` prefix :)

jgraham:
> Given the capitalization name conflicts with the web platform are virtually impossible, I'd like it better without the `Wpt` prefix :)

OK, changed.

rfcs/reftest_ready.md
```diff
@@ -0,0 +1,39 @@
+# RFC 34: `reftests`: Fire event when painting is done
+
+## Summary
+
+Fire a `TestRendered` event for reftests when a `reftest-wait` class is present on the root and layout, fonts and painting are complete. This coresponds to the point at which the screenshot would be taken in the absence of `reftest-wait`.
```
foolip: Typo: coresponds

rfcs/reftest_ready.md
```diff
@@ -0,0 +1,39 @@
...
+The detailed steps for running a reftest with this addition are as follows:
+
+* Wait for the document to complete loading, all fonts to load, and for pending paints to be flushed.
+* If the document root element does not contain a class named `reftest-wait`, screenshot the viewport and abort these steps
+* Fire an event at the document root with the name `TestRendered` and the bubbles attribute set to true.
+* Wait for the `reftest-wait` class to be removed from the document root
+* Wait for pending paints to flush
+* Screenshot the viewport
+
+This addition corresponds to the gecko `MozReftestInvalidate` event. Providing this event is expected to unblock upstreaming some more Gecko reftests.
```
foolip: Yay!

jgraham:
I don't think this has substantive disagreement, so unless something comes up in the next day and a half I'm going to merge the issue and make a PR implementing the feature.

foolip:
I think this is ready to go, I'll go ahead and merge.

foolip:
No, actually, the file will and PR will need to be renamed, so I'll leave it you, @jgraham!

PR #37, opened 2019-12-10
RFC 37: Split META.yml suggested_reviewers into suggested_reviewers and notify

zcorpan:
See the [rendered version of this RFC](https://github.com/web-platform-tests/rfcs/blob/zcorpan/meta-yml-notify/rfcs/meta_yml_notify.md).

### DIFF
--- a/rfcs/meta_yml_notify.md
+++ b/rfcs/meta_yml_notify.md
@@ -1,0 +1,50 @@
+# RFC 37: Split META.yml `suggested_reviewers` into `suggested_reviewers` and `notify`
+
+## Summary
+
+Change `META.yml` files in wpt to have `notify` in addition to `suggested_reviewers`.
+@wpt-pr-bot mentions everyone in `notify` and then requests review from one person from `suggested_reviewers`.
+
+## Details
+
+At TPAC 2019 we set out some [priorities for 2020](https://docs.google.com/document/d/1gie7LFb6cAUfabY86MYuWM7m7ux_FaKhDkLdpz0zWkg/edit),
+which include increasing PR review velocity and reduce the number of stale PRs.
+
+wpt has META.yml files with `suggested_reviewers`. @wpt-pr-bot will request review from everyone listed in `suggested_reviewers` for PRs that change that directory, and also assign one of them in a round-robin fashion.
+
+See https://web-platform-tests.org/reviewing-tests/index.html#notifications
+
+The documentation only says that `suggested_reviewers` is for getting notifications for PRs.
+
+This has a few problems:
+
+* There is a disconnect between what contributors would reasonably expect and what often actually happens. When people are requested review and someone is assigned, it communicates that the person assigned will help the contributor to get it landed.
+* The current system doesn't support varying preferences from reviewers. Some people want to see all PRs for a given folder (which is supported by the current system), while others want to help with reviews but not be notified for everything, and some may want to be notified for everything but can't commit to helping with reviews.
+
+To address these problems, have two separate keys in `META.yml`:
+
+* `suggested_reviewers`: @wpt-pr-bot will only request review from one of these, similar to how assignee currently works.
+* `notify`: @wpt-pr-bot will add a comment (before requesting reviews) notifying everyone listed.
+
+Either @wpt-pr-bot can continue to also assign the selected reviewer, or it can only request review and not assign anyone. (This should be decided before merging this RFC.)
+
+This setup allows the following combinations:
+
+* People who want to help with review and want to be notified for everything add themselves to both lists.
+* People who want to help with review but not be notified for everything add themselves to `suggested_reviewers`.
+* People who don't want to help with review but still be notified for everything add themselves to `notify`.
+
+## Risks
+
+Having bots add comments in wpt has been a source of contention in the past. This RFC reintroduces a comment.
+
+The known problems are:
+
+* Additional notifications. When someone is requested review, and a bot later adds a comment, it can cause a new notification.
+* Clutter the thread with bot comments.
+
+These are partially mitigated by having the bot add the comment before (or at the same time as) requesting review. The bot could further keep the comment short, and edit its first comment instead of adding more comments if the PR is updated.
+
+## Alternatives considered
+
+* Encourage reviewers to use email filters. Not everyone uses email for GitHub in the first place, and most likely not everybody is fond of setting up email filters for wpt.
--- a/rfcs/meta_yml_notify.md
+++ b/rfcs/meta_yml_notify.md
@@ -23,10 +23,10 @@ This has a few problems:
 
 To address these problems, have two separate keys in `META.yml`:
 
-* `suggested_reviewers`: @wpt-pr-bot will only request review from one of these, similar to how assignee currently works.
-* `notify`: @wpt-pr-bot will add a comment (before requesting reviews) notifying everyone listed.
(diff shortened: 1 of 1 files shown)


rfcs/meta_yml_notify.md
```diff
@@ -0,0 +1,50 @@
...
+
+Change `META.yml` files in wpt to have `notify` in addition to `suggested_reviewers`.
+@wpt-pr-bot mentions everyone in `notify` and then requests review from one person from `suggested_reviewers`.
+
+## Details
+
+At TPAC 2019 we set out some [priorities for 2020](https://docs.google.com/document/d/1gie7LFb6cAUfabY86MYuWM7m7ux_FaKhDkLdpz0zWkg/edit),
+which include increasing PR review velocity and reduce the number of stale PRs.
+
+wpt has META.yml files with `suggested_reviewers`. @wpt-pr-bot will request review from everyone listed in `suggested_reviewers` for PRs that change that directory, and also assign one of them in a round-robin fashion.
```
foolip: Is implementing the round robin review part of this RFC?

rfcs/meta_yml_notify.md
```diff
@@ -0,0 +1,50 @@
...
+
+* There is a disconnect between what contributors would reasonably expect and what often actually happens. When people are requested review and someone is assigned, it communicates that the person assigned will help the contributor to get it landed.
+* The current system doesn't support varying preferences from reviewers. Some people want to see all PRs for a given folder (which is supported by the current system), while others want to help with reviews but not be notified for everything, and some may want to be notified for everything but can't commit to helping with reviews.
+
+To address these problems, have two separate keys in `META.yml`:
+
+* `suggested_reviewers`: @wpt-pr-bot will only request review from one of these, similar to how assignee currently works.
+* `notify`: @wpt-pr-bot will add a comment (before requesting reviews) notifying everyone listed.
+
+Either @wpt-pr-bot can continue to also assign the selected reviewer, or it can only request review and not assign anyone. (This should be decided before merging this RFC.)
```
foolip: I think it'd be better to stop using the assignee field, even though that means I'll have to update mail filters again.

rfcs/meta_yml_notify.md
```diff
@@ -0,0 +1,50 @@
...
+
+This setup allows the following combinations:
+
+* People who want to help with review and want to be notified for everything add themselves to both lists.
+* People who want to help with review but not be notified for everything add themselves to `suggested_reviewers`.
+* People who don't want to help with review but still be notified for everything add themselves to `notify`.
+
+## Risks
+
+Having bots add comments in wpt has been a source of contention in the past. This RFC reintroduces a comment.
```
jgraham: Yeah, I'm worried about this. In general I think if we were going to have a comment from bots, I'd expect much more value in providing a concise summary of test failures, surfacing warnings, or linking to results, than in reverting to a system where we `@`-notify people in a comment in the hope that tweaking notifications makes people spend more time reviewing PRs.

foolip: The purpose of those comments wouldn't mainly be to get review, but rather:
- move people from `suggested_reviewers` to `notify` if they have no intention of reviewing
- put people in both `suggested_reviewers` and `notify` if they do want to review but also don't want to miss anything. I think @annevk is likely in this category.

foolip:
@zcorpan by happenstance I noticed that the GitHub API has a `requested_teams` field on PRs, and that indeed it's possible to request review from a team. Can you check what this does in terms of notifications and email filtering? I don't think that it's a solution to any of our problems, but if it does something different to requesting from everyone separately it might be worth considering.

rfcs/meta_yml_notify.md
jgraham: Sure, but the main concern still applies. If we have some acceptable non-zero number of bot comments we can add without annoying people too much, why is this the most useful thing to spend that budget on? We have a lot of information that's buried several links deep in the checks API that we could instead surface in a comment, for example. To me that seems likely to be higher impact than tweaking the notification system to allow people to be notified without being flagged for review.

zcorpan: I don't think having mentions instead of review requests is more important than surfacing other information hidden in CI checks, but I think the problem this RFC is trying to address is worth doing something about.

Personally, I'd love to have a bot comment that surfaces all useful information we want people to see (such as [wptpr.live links](https://github.com/web-platform-tests/wpt/issues/19607)).

annevk: I don't want bot comments. A bot modifying OP as the PR bot for standards does is fine, but adding comments is noisy and annoying.

Does adding usernames in an edit result in notifications for those added?

zcorpan: It does, yes. (Tested in https://github.com/web-platform-tests/wpt/pull/20660 )

Editing OP would be ok for me.

zcorpan:
(I haven't yet looked into `requested_teams`. I've made a todo to do so in early January.)

Having `notify` for "I'm interested in everything" could also allow for different delegation of 
`suggested_reviewers`.

For example, myself and @foolip are listed in `html/META.yml`, but for `html/obsolete/META.yml` only @foolip is. Right now the latter makes no difference; we're both notified for PRs in `obsolete`. An alternative would be for @wpt-pr-bot to only consider the nearest META.yml for `suggested_reviewers`, and all ancestor directories' META.yml for `notify`.

I don't know if this would be an improvement or a regression, though. Thoughts?

gsnedders:
Pretty sure that would be a regression. I'm not opposed to some way to non-recursively add yourself as a reviewer/notify or a way to exclude certain children.

rfcs/meta_yml_notify.md
zcorpan: I've changed the RFC to say to edit OP.

rfcs/meta_yml_notify.md
zcorpan: Done.

rfcs/meta_yml_notify.md
zcorpan: Yes, clarified below.

zcorpan:
I've updated this to address comments. I still haven't looked into `requested_teams`.

stephenmcgruer:
@zcorpan At this point, are you still interested in pursuing this RFC?

If not, we should determine whether there is appetite from another community member to champion this. It seems like there is enough interest that we shouldn't just drop it.

zcorpan:
Unfortunately, I don't have bandwidth to work on this at the moment (or the next few months probably). I would be happy if someone else can take this. :)

stephenmcgruer:
Thanks @zcorpan ; it can be hard to take the step back and acknowledge when one is overloaded, but it's also important so I appreciate you doing so :)

I have reviewed this thread and the RFC itself. Outside of the actual implementation work, it appears there is not that much _currently_ to do here. I came up with the following list:

- Determine what `requested_teams` does and if would be useful.
- (Optionally) Consider other extensions that would be useful.

For the latter, I took a look through Chromium's OWNER files for inspiration. We have:

- `noparent` (tells an OWNERS file not to recurse _upwards_),
- `per_file` (accepts a glob-pattern for files a particular person cares about),
- `file` (which appends another OWNERS file onto this one),
- and a `WATCHLIST` concept which sort of separates our 'notify' and 'reviewer' sets into different files.

jgraham:
It seems like this is not making progress currently and there's no one who is able to champion it. I'm going to close the issue for now, but we can reopen it if we identify a way forward.

PR #38, opened 2019-12-11
Enable HTTP/2 server

jgraham:


rfcs/http2.md
```diff
@@ -0,0 +1,19 @@
...
+
+By default this server is configured to run on port 9000.
+
+In addition, actually using the server for testing is easier if one can directly load top-level tests on the server. For the https server we do this with a filename flag `.https.`. By analogy, the HTTP/2 server will work the same way with a `.h2.` flag.
+
+## Risks
+
+Running an extra server is always some risk since it may cause instability.
+
+Vendor CI systems may not allow the use of port `9000` for some reason.
```
Hexcles: We can always customize it, like all the other ports, right?

jgraham: Yes. I was just mentioning it so people running those systems are aware that something is changing in case there are config changes required. But I guess it's not a risk of accepting the change so much as a risk that the change requires downstream work.

jgraham: Added this based on the comments in the source; it's a little unclear if 2.7.10+ works or if 2.7.15 is required, so I added something conservative.

jgraham:
This has now been open for a week without any pushback, but I'd like to see some affirmitive approval before merging it.

PR #39, opened 2019-12-31
Consider adding a `.www` file name flag.

mikewest:
It would be simpler to write tests for features that rely on interaction between subdomains in a site
if the top-level document could be loaded from a domain other than the `web-platform.test` apex.

Addresses web-platform-tests/wpt#20932

### DIFF
--- a/rfcs/top_level_subdomain.md
+++ b/rfcs/top_level_subdomain.md
@@ -1,0 +1,16 @@
+# RFC XX: Loading top-level tests on a subdomain.
+
+## Summary
+
+Add a `.www` [file name flag](https://web-platform-tests.org/writing-tests/file-names.html) that causes a given test to be loaded from the subdomain `www.web-platform.test` rather than the apex domain `web-platform.test`.
+
+## Details
+
+Tests are typically loaded from `http://web-platform.test/`. For some features that verify behavior within a site, it would be helpful to load the top-level document from a subdomain (e.g. `http://www.web-platform.test/`). Tests for cookies' `Domain` attribute are a good example: it's possible to write tests that verify that attribute's effects on requests from `www.web-platform.test` to `web-platform.test`, but requires creating a proxy document in a new window or frame, running tests in that context, and `postMessage()`ing results back to the main document. This adds a good deal of complexity that could be avoided if we opened the top-level document on the subdomain to begin with.
+
+We distinguish files that ought to be loaded from `https://web-platform.test/` with an `.https` [file name flag](https://web-platform-tests.org/writing-tests/file-names.html).  We could support loading tests from `www.web-platform.test` by adding a `.www` flag.
+
+## Risks
+
+*   At some point, we might end up with a more-than-reasonable number of flags to consider when writing a test: `amazing-cookie-test.https.www.sub.tentative.any.js` might be a bit much...)
+*   Naming is hard. `.subdomain` is appealing, but confusable with `.sub`.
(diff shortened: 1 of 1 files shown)


jgraham:
Generally this seems reasonable.

I wonder if people are going to expect other subdomains to work as well. I'm hoping that there isn't a use case for more than one sub domain for the top level test (although I suspect there will be uses for e.g. `www.www.web-platform.test`) in which case the RFC should propose adding a lint to ensure that people don't use those subdomains in the test name.

rfcs/top_level_subdomain.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Details
+
+Tests are typically loaded from `http://web-platform.test/`. For some features that verify behavior within a site, it would be helpful to load the top-level document from a subdomain (e.g. `http://www.web-platform.test/`). Tests for cookies' `Domain` attribute are a good example: it's possible to write tests that verify that attribute's effects on requests from `www.web-platform.test` to `web-platform.test`, but requires creating a proxy document in a new window or frame, running tests in that context, and `postMessage()`ing results back to the main document. This adds a good deal of complexity that could be avoided if we opened the top-level document on the subdomain to begin with.
+
+We distinguish files that ought to be loaded from `https://web-platform.test/` with an `.https` [file name flag](https://web-platform-tests.org/writing-tests/file-names.html).  We could support loading tests from `www.web-platform.test` by adding a `.www` flag.
+
+## Risks
+
+*   At some point, we might end up with a more-than-reasonable number of flags to consider when writing a test: `amazing-cookie-test.https.www.sub.tentative.any.js` might be a bit much...)
```
foolip: A bit much, yeah :)

However, it seems like this isn't solved well by not adding flags, if it does get unreasonable then we'll have to find a way out of the mess and the number of flags then won't really make a difference to the solution.

mikewest:
> Do you have a draft implementation to get some confidence there's nothing unwieldy about this in practice?

No. I was imagining it would just follow along with whatever we're doing to generate the URL for `.https`, but I couldn't quickly figure out how to determine what the subdomain that would be generated for `{{domains[www]}}` would be at the point that we need it. I'm just not familiar enough with WPT internals. :(

Hexcles:
cc @npm1 (I forgot if I told you about this: it'd satisfy your use case of specifying which origin a test should run in)

stephenmcgruer:
This RFC has two approvals, and no clear outstanding questions aside from implementation complexity (where @mikewest deferred due to not knowing enough about WPT internals).

I'd like to start the 1-week clock here, but am concerned that we may end up with an RFC that is accepted but the actual implementation work never happens (if this doesn't have someone willing to implement it). Thoughts, @web-platform-tests/wpt-core-team ?

sideshowbarker:
I agree that we have a problem if nobody is committed to working on implementing this. But I don’t have any novel suggestions about how to solve that problem.

That said, I would think it could be helpful to go ahead with getting the RFC accepted first. At that point it’s clear at least that nobody has objected to the identified need. And that would at least not be harmful to the goal of getting somebody to work on actually implementing it.

mikewest:
If this is something y'all can accept, I can try to hack an implementation together. I suspect I will need help. :)

stephenmcgruer:
> If this is something y'all can accept, I can try to hack an implementation together. I suspect I will need help. :)

Sorry, I dropped the ball here. I think this is definitely something we can accept; the RFC has approvals and no concerns. I think if you can produce a prototype PR we will happily merge this RFC.

miketaylr:
> I think if you can produce a prototype PR we will happily merge this RFC.

FYI: I have a prototype PR in https://github.com/web-platform-tests/wpt/pull/26263

stephenmcgruer:
Landing since (wayyyy) more than 7 days have passed, and the prototype looks pretty good (can work out any minor problems in review).

Issue #20, opened 2019-03-20
Manifest vNEXT (trie)


gsnedders:
Posting this as an issue, as it's really just somewhere to get a bunch of opinions on vague ideas rather than anything concrete:

We have a number of things in the manifest currently which really exist to deal with older generated manifests (which contain a number of bugs), see especially comments containing `MANIFEST6` in [manifest.item](https://github.com/web-platform-tests/wpt/blob/702c8f03a81135e52341a3a288d055f086f90b6f/tools/manifest/item.py).

I think it's time we start considering what changes we want to make if we're going to bump the version, given we can certainly go further than just forcing regeneration to get rid of any old bugs.

There are some reasons why moving away from JSON seems tempting, but I don't know how justified they are (on my home computer, parsing/serializing the JSON with `ujson` takes ~300ms, and ~500ms with `json`). Older checkouts of wpt would fail horribly if they couldn't read the `MANIFEST.json` (anything prior to https://github.com/web-platform-tests/wpt/pull/10323 landing on 5 Apr 2018), but presumably we wouldn't use the `.json` extension if we weren't using JSON. Some people have spoken about protobuf or msgpack, and both are somewhat reasonable suggestions if we want to move away from JSON.

I would be pretty tempted to move tests to being stored in a path trie (so instead of `tests["reftest"]["css/CSS2/bidi-005.xht"]` we'd have `tests["reftest"]["css"]["CSS2"]["bidi-005.xht"]`), so that we can entirely skip directories (e.g., for `./wpt run --include html/ firefox` we don't have to iterate through the entire dict of tests to find only those included). If we moved away from JSON, we might be able to do some sort of lazy parsing or skip over dicts we don't care about and avoid the not insubstantial cost of allocating Python data structures for them.

It might also be nice to use some format whereby we can simply append to the file for a small number of changes, rather than having to re-serialize the entire file, but that might not be worth it if we can decrease the cost of serialising.

jgraham:
I believe there are concerns about the format not being text-based; both Chrome and Servo check the manifest in to their repositories, so having something binary is probably bad.

Hexcles:
> If we moved away from JSON, we might be able to do some sort of lazy parsing or skip over dicts we don't care about and avoid the not insubstantial cost of allocating Python data structures for them.

[...]

PR #40, opened 2020-01-21
Manifest path trie RFC

gsnedders:
Rendered: https://github.com/gsnedders/rfcs/blob/manifest-path-trie/rfcs/manifest-path-trie.md.

See also #20 and https://github.com/web-platform-tests/wpt/pull/16537.

### DIFF
--- a/rfcs/manifest-path-trie.md
+++ b/rfcs/manifest-path-trie.md
@@ -1,4 +1,4 @@
-# RFC XX: New Manifest Format (path trie)
+# RFC 40: New Manifest Format (path trie)
 
 ## Summary
 
@@ -10,9 +10,9 @@ At the moment, the manifest is formed of three components: the url_base (a strin
 
 The proposal here is two-fold:
 
-Firstly, replace the items objects with a series of nested objects forming a trie of path segments. This makes it much quicker to iterate through only specific directories, which is often enough done in a test-debug cycle.
+Firstly, replace the items objects with a series of nested objects forming a trie of path segments. This allows us to make it much quicker to iterate through only specific directories, which is often enough done in a test-debug cycle, as part of redoing the implementation of include/exclude (this is future work for this RFC).
 
-Secondly, we get rid of path_hash: we migrate the hashes into the item object, thereby getting rid of the duplication of all the paths. This almost halves the size of the JSON created, which when the small-update case was previously dominated by parsing/serializing the JSON is significant.
+Secondly, we get rid of path_hash: we migrate the hashes into the item object, thereby getting rid of the duplication of all the paths. This almost halves the size of the JSON created, which when the small-update case was previously dominated by parsing/serializing the JSON is significant. This provides a ~10% speedup running `wpt run` after modifying a single test.
 
 Additionally, we stop storing the URL if it is identical to the path; this again provides size savings but is not a critical part of this proposal.
 
--- a/rfcs/manifest-path-trie.md
+++ b/rfcs/manifest-path-trie.md
@@ -1,0 +1,21 @@
+# RFC XX: New Manifest Format (path trie)
+
+## Summary
+
+We propose restructuring the manifest to be a trie of path segments, rather than the status-quo of being an object with (full) paths as keys.
+
+## Detail
+
+At the moment, the manifest is formed of three components: the url_base (a string), the items (an object of type to object pairs, then an object giving each path and the items contained therein), and the path_hash (storing all the paths—again—and the type & file hash).
+
+The proposal here is two-fold:
+
+Firstly, replace the items objects with a series of nested objects forming a trie of path segments. This makes it much quicker to iterate through only specific directories, which is often enough done in a test-debug cycle.
+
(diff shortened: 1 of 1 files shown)


jgraham:
If we have performance numbers for this adding them to the RFC would help make the case for accepting this change.

The details section could mention more clearly the benefits other than file size e.g. the ability to support partial updates.

rfcs/manifest-path-trie.md
```diff
@@ -0,0 +1,21 @@
...
+
+We propose restructuring the manifest to be a trie of path segments, rather than the status-quo of being an object with (full) paths as keys.
+
+## Detail
+
+At the moment, the manifest is formed of three components: the url_base (a string), the items (an object of type to object pairs, then an object giving each path and the items contained therein), and the path_hash (storing all the paths—again—and the type & file hash).
+
+The proposal here is two-fold:
+
+Firstly, replace the items objects with a series of nested objects forming a trie of path segments. This makes it much quicker to iterate through only specific directories, which is often enough done in a test-debug cycle.
```
foolip: Grammar around “done”

gsnedders:
> If we have performance numbers for this adding them to the RFC would help make the case for accepting this change.

This RFC itself makes little difference to performance; the big gain is it makes include/exclude much easier to reimplement without iterating through all tests. The only significant change here is it reduces serialization cost, which is primarily apparent when a small number of tests are modified.

> The details section could mention more clearly the benefits other than file size e.g. the ability to support partial updates.

This isn't actually any different than before, and I don't think it's changed the implementation complexity of that (reftest-simplification did that).

gsnedders:
> In the case of no substantive disagreement the RFC is considered accepted after 1 week.

tabatkins:
So yay, I wasn't aware of this change, which caused all of Bikeshed's WPT data to become incorrect and so the entire `<wpt>` feature to be broken.

I can monitor this repo if necessary, but I wasn't even aware that it *existed* - it's not, to my knowledge, linked from the README in the main /wpt repo.

Somewhat annoyed that Bikeshed, a known major consumer of the manifest data, wasn't notified of a major change to the manifest format.

Hexcles:
@tabatkins we're terribly sorry. This really should not have happened. I'm not sure why Bikeshed didn't cross anyone's mind... Especially when I filed https://github.com/web-platform-tests/wpt.fyi/issues/1777 (I even said "all features that rely on /api/manifest are off by default", sigh...)

We are taking steps to prevent this from happening. In addition to the making the RFC repo more prominent from the main WPT repo and including the version number in the response headers of `wpt.fyi/api/manifest`, I have another idea: https://github.com/web-platform-tests/rfcs/issues/43

jgraham:
So first of all, apologies for breaking bikeshed; that was obviously unintentional. I had no idea that it was using the manifest, and to be honest didn't even remember that wpt.fyi was providing a public API here.

So clearly one thing that would have improved the situation would be having more consumers pay attention to the RFC repo. But I honestly doubt linking it in the README — though certainly a good idea — is going to make much difference. The RFC process was for example announced on the mailing list, and rather extensively discussed on IRC. It seems unlikely someone who missed all of that would notice a link buried in an already extensive README.

Even if you knew about the RFC repo, it technically doesn't cover wpt.fyi endpoints, and so it would be reasonable to assume that as a consumer of those endpoints changes in the RFC repo don't apply to you.

So in addition to the question about how to make people more aware of RFCs, I think we need to have a clear understanding of what stability guarantees are supposed to be offered by wpt.fyi endpoints, and ensure that there's some mechanism to get notified of changes there. It seems possible that the expectations around stability of the manifest format (largely: that it isn't) are incompatible with the kind of guarantees that wpt.fyi might want to make in general, so perhaps we should only make the raw manifest available on a URL that's explicitly marked unstable, and make other data from the manifest available in a stable format.

tabatkins:
Apologies accepted, it was a minor hiccup overall and quickly fixed.

> The RFC process was for example announced on the mailing list, 

I just did a search thru my email for "wpt rfc" and didn't see any relevant hits. When/where did this discussion happen?

> and rather extensively discussed on IRC. 

w3.org#testing? That channel is *extremely* noisy with BitBot announcements; if I catch any discussion *at all* I consider myself lucky. Right now, my 30' vertical-oriented screen shows messages in that room up to 2am this morning (~11 hours ago)

> It seems unlikely someone who missed all of that would notice a link buried in an already extensive README.

Strong disagree. There's a nice list of links already in the README with explanations of why you might want to visit each one; having one linking to /rfcs/ and explaining that breaking changes get announced/discussed there would have been caught by me, at least.

tabatkins:
(That said, I wouldn't mind a more dedicated mailing list for this kind of thing.)

foolip:
The list where this was discussed is public-test-infra: https://lists.w3.org/Archives/Public/public-test-infra/2019JanMar/0002.html

However, I'm not surprised that not everyone who could be affected by changes to WPT are on that list. Anyway, looks like things got sorted pretty quickly here, and https://github.com/web-platform-tests/wpt/pull/21682 updated the README.

tabatkins:
Yup, I wasn't subscribed to that; I am now.

And yeah, the README change is good; I consider my issue dealt with. Thanks, all!

PR #42, opened 2020-02-05
RFC 42: QUIC server in WPT

Hexcles:
[Rendered](https://github.com/web-platform-tests/rfcs/blob/quic/rfcs/quic.md)

This RFC addresses https://github.com/web-platform-tests/wpt/issues/19114

cc @yutakahirano

### DIFF
--- a/rfcs/quic.md
+++ b/rfcs/quic.md
@@ -19,7 +19,7 @@ See the [original issue][original-issue] for more background.
 
 ### Requirements
 
-1.  The QUIC server requires decent cyptography support; pure Python isn't
+1.  The QUIC server requires decent cryptography support; pure Python isn't
     sufficient.
 2.  The server needs to be scriptable, and the custom handlers should live in
     the wpt repo next to test files, so that they can be updated together.
@@ -46,10 +46,10 @@ For **test writers**: the QUIC server will support Python custom handlers
 similar to `wptserve`. The APIs will likely be different. To more easily
 distinguish the two, we add a filename flag to QUIC custom handlers
 (`.quic.py`). These handlers will be ignored by `wptserve`. Tests that require
-the QUIC server need to be marked with a meta tag (tentaively `<meta name="quic"
-content="true">` and `// META: quic=true`). This information will be extracted
-by `wpt manifest` into `MANIFEST.json`, so that `wptrunner` can skip the tests
-that require the QUIC server easily.
+the QUIC server need to be marked with a meta tag (tentatively `<meta
+name="quic" content="true">` and `// META: quic=true`). This information will be
+extracted by `wpt manifest` into `MANIFEST.json`, so that `wptrunner` can skip
+the tests that require the QUIC server easily.
 
 For **test runners**: the goal is to not introduce any observable changes to
 infrastructure maintainers not yet concerned with QUIC or WebTransport. All
@@ -95,7 +95,7 @@ startup. The server listens on port 9001 by default, which may affect services
 that start afterwards if they happen to use the same port.
 
 The change required in `wpt` is limited, and the QUIC server itself is
-standalone, so the maintenance burder will be low. The Ecosystem Infra team at
+standalone, so the maintenance burden will be low. The Ecosystem Infra team at
 Google will be responsible for maintaining this integration point.
 
 A new meta tag means a new field in the manifest JSON (only needed for tests
--- a/rfcs/quic.md
+++ b/rfcs/quic.md
@@ -2,16 +2,17 @@
 
 ## Summary
 
-Add an **optional** [QUIC](https://quicwg.org/) server to wpt, implemented with
-[aioquic](https://github.com/aiortc/aioquic) in Python 3 and listening on port
-`9001` by default. This server is used to test web APIs that rely on the
-QUIC protocol (e.g. [WebTransport](https://wicg.github.io/web-transport/)). The
-server is optional but tries to start by default; `wpt` front-end will skip the
-tests that rely on QUIC if the QUIC server cannot be started (e.g. due to
-missing dependencies).
+Add an *optional*, *off-by-default* [QUIC](https://quicwg.org/) server to wpt,
+implemented with [aioquic](https://github.com/aiortc/aioquic) in Python 3. This
+server is used to test web APIs that rely on the QUIC protocol (e.g.
+[WebTransport](https://wicg.github.io/web-transport/)). If the server is not
+enabled, `wptrunner` will skip the tests that rely on QUIC.
 
 QUIC custom handlers use a new filename flag `.quic` (`.quic.py`). And a meta
 tag is included by tests that require the QUIC server.
+
+This RFC focuses on the high-level architecture and dependency management to
+enable prototyping the QUIC server. The server APIs are not yet finalized.
 
 See the [original issue][original-issue] for more background.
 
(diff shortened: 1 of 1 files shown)


jgraham:
@annevk Assuming I look at the implmentation details, is there someone from Mozilla who should give feedback on whether this meets any requirements from our side?

rfcs/quic.md
```diff
@@ -0,0 +1,104 @@
+# RFC 42: Add an optional QUIC server to wpt
+
+## Summary
+
+Add an **optional** [QUIC](https://quicwg.org/) server to wpt, implemented with
+[aioquic](https://github.com/aiortc/aioquic) in Python 3 and listening on port
+`9001` by default. This server is used to test web APIs that rely on the
+QUIC protocol (e.g. [WebTransport](https://wicg.github.io/web-transport/)). The
+server is optional but tries to start by default; `wpt` front-end will skip the
```
jgraham: Can we make this opt-in initially, like we did for http2? In fact I think I'd prefer "opt in and fails if you opt in and it can't start" rather than ending up in a situation where you only realise the server failed to start when investigating why tests aren't running.

rfcs/quic.md
```diff
@@ -0,0 +1,104 @@
...
+Its entrypoint will have a shebang that explicitly requires `python3`. An
+implication is that the QUIC server will be relatively self contained, without
+dependencies on other modules in wpt that are currently Python 2-only.
+
+Code for the QUIC server itself will live in `tools/quic`.
+
+### User interface
+
+For **test writers**: the QUIC server will support Python custom handlers
+similar to `wptserve`. The APIs will likely be different. To more easily
```
jgraham: I think we should have details of what the actual API will be as part of this or some other RFC.

rfcs/quic.md
```diff
@@ -0,0 +1,104 @@
...
+distinguish the two, we add a filename flag to QUIC custom handlers
+(`.quic.py`). These handlers will be ignored by `wptserve`. Tests that require
+the QUIC server need to be marked with a meta tag (tentatively `<meta
```
jgraham: This isn't a filename flag because it isn't about loading the tests themselves over quic, but about being able to filter out quic-requiring tests where the server isn't available, correct? So we might hope to remove this in the future. Is there some way to enforce this; it seems like it's easy for a test not to add the meta element and still work correctly in configurations where quic is enabled?

rfcs/quic.md
```diff
@@ -0,0 +1,104 @@
...
+name="quic" content="true">` and `// META: quic=true`). This information will be
+extracted by `wpt manifest` into `MANIFEST.json`, so that `wptrunner` can skip
+the tests that require the QUIC server easily.
+
+For **test runners**: the goal is to not introduce any observable changes to
+infrastructure maintainers not yet concerned with QUIC or WebTransport. All
+existing tests will continue to run without any infra changes, and all new
+QUIC-dependent tests will be skipped unless all dependencies are met (or can
```
jgraham: It seems hard to make this true if the only mechanism is adding metadata that doesn't affect the test in the case QUIC is enabled.

rfcs/quic.md
```diff
@@ -0,0 +1,104 @@
...
+also be skipped explicitly).
+
+### Notable dependencies
+
+* [sys] Python 3 (>=3.6)
+* [py] aioquic
```
jgraham: Are we happy with the mainteinance story of this library? How did you decide that it has the features we need to give the low-level protocol access that's required for writing tests?

rfcs/quic.md
```diff
@@ -0,0 +1,104 @@
...
+    * [py] cryptography
+        * [sys] openssl (>=1.0.1, already required by [HTTP/2](http2.md))
+    * [py] pylsqpack
+
+All of the Python dependencies above have native extensions, which requires
```
jgraham: So it seems like updating the version here is going to be an infra-breaking change for vendors because work will be required on the vendor side to update to the new version (e.g. updating the deps file or uploading the wheel to the internal pypi mirror or whatever each vendor does). That seems like a pretty big concern to me. If we did vendor the source, we could have logic like "check if the correct version is available, or pip install, or build from source". Maybe there are other possibilities to address this?

martinthomson:
I think that this is a reasonable plan, but I can't comment on the specifics until I have more information about webtransport. In that respect, this is probably well premature. 

If you regard the current design being proposed by Google as a fait accompli, this is probably enough, but the extra effort required to build something usable on top of a generic QUIC stack could turn out to be such that this approach becomes unwieldy.

rfcs/quic.md
Hexcles: Sure we can do that. I'll change the RFC later.

rfcs/quic.md
Hexcles: > This isn't a filename flag because it isn't about loading the tests themselves over quic, but about being able to filter out quic-requiring tests where the server isn't available, correct?

Correct. I thought about using `.quic.html`, similar to `.h2`, but realized that they are in fact different. The test themselves should not be loaded via QUIC.

> So we might hope to remove this in the future.

Probably, when all major vendors have WebTransport implementations and we run these tests by default.

> Is there some way to enforce this; it seems like it's easy for a test not to add the meta element and still work correctly in configurations where quic is enabled?

Great point! I don't have good ideas right now. Any thoughts?

rfcs/quic.md
Hexcles: @yutakahirano knows more details here.

rfcs/quic.md
Hexcles: I'm thinking about this from two perspectives:

* For average human users, it's most likely they are using a platform supported by PyPI wheels, and the dependencies will be installed into `_venv` using `pip` automatically. Vendoring the source isn't very useful for them.
* For browser infras, AFAIK both Firefox and Chromium discourage or even disallow building a native Python extension at run time; these dependencies will have to come from a PyPI-like repo. If that's the case, vendoring the source doesn't seem helpful, either.

I agree that this will incur a maintenance burden on the vendors if they want to run QUIC tests. (I'll acknowledge this in the risks section.) Meanwhile, I don't see vendoring the source code would help the browser vendors if my understanding above is correct -- someone would still need to build and distribute the new version.

Hexcles:
Thanks for the constructive and insightful feedback!

I want to highlight one point: this has NOT been implemented yet, and the design of the QUIC server itself (internal architecture, handler API, etc.) is intentionally vague because I don't have the details yet and we will probably iterate on the design once we start prototyping -- @yutakahirano and the team will work on this.

The key part of this RFC is the integration between the QUIC server and the rest of WPT infra, which should not need to change regardless of the detailed design of the QUIC server itself. This RFC paves the way for prototyping and eventually productionizing the QUIC server within WPT with minimum overhead. We can send out another RFC when we are happy with a detailed server design or implementation.

rfcs/quic.md
jgraham: I don't think our test environment forbids building a native extension, although it also seems likely that the required libraries aren't available. What would definitely be possible (although not tivial) is adding the build step to the Firefox build and distributing the resulting package to the test machines; we already have various test-only binaries that need to be produced.

What worries me about the proposed setup is that updating the library version is a sync-blocking change for Gecko since we're a) using the tools/ directory directly and b) don't allow PyPI access from test machines for robustness reasons.

Hexcles: That's true for Chromium, too. I imagine that in the future when we need to update these native dependencies, we will need to create a PR that updates the `requirements.txt` for venv, and ask each vendor to prepare the packages in their infras, before merging the PR. Native dependencies are difficult to deal with in general; I'm intentionally making some tradeoff here between ongoing maintenance costs and complexity.

Do you have a more concrete alternative proposal? If you think vendoring the source into the codebase is easier for Firefox infra, then I'd be happy to add that as a fallback to wheels, too. I just didn't want to vendor them if it wouldn't be used by anyone.

Hexcles:
@jgraham I've updated the RFC:
1. Switch to off-by-default
2. Clarify the scope of the RFC (i.e. not including server APIs)
3. Clarify dependency management and list it as a "risk"

Could you take another look? There are a few open questions now:
* Deferring the detailed server API design
* Unable to strictly enforce that a test without the QUIC tag does not need QUIC
* Dependency management

Do you see any of these as blocking? These are some genuinely good questions to which I don't have perfect answers to. The currently proposed design makes some trade-offs in these areas.

Hexcles:
ping @jgraham

rfcs/quic.md
jgraham: I think that adding the source would *in principle* be better for Mozilla, since we could in theory add it to the build system. In practice I don't know that we'd actually get around to doing that :/ It probably depends on how hard it is vs doing manual updates. But I think I'd like us to do it anyway.

In any case I think the process probably demands an RFC for version updates, since it's a change that affects multiple vendors. That's pretty heavyweight, but this really is the kind of change that can break integrations.

Hexcles:
@jgraham I think I've addressed all the remaining comments. Please take another look at the latest commit. Thanks!

Hexcles:
ping @jgraham

rfcs/quic.md
```diff
@@ -0,0 +1,115 @@
...
+users to use a platform supported by the prebuilt wheels from PyPI or build and
+distribute the wheels themselves (e.g. in Chromium through
+[cipd](https://chromium.googlesource.com/chromium/src/+/master/docs/cipd.md)).
+
+The source code of all dependencies will also be vendored into
+`tools/third_party` to allow building from source if needed, even though we
```
jgraham: Remove everything after the comma (there's a typo, but also I don't think there's any sense in talking about whether there's a plan for that in the wpt infra; this is for vendors).

rfcs/quic.md
```diff
@@ -0,0 +1,115 @@
...
+there is no plan for that yet.
```
Hexcles:
```suggestion
`tools/third_party` to allow building from source if needed.
```

rfcs/quic.md
Hexcles: Done.

Hexcles:
Heads up: if there's no more comment I'll be merging this in 3 days (which will be a full week after the last comment).

PR #44, opened 2020-02-11
RFC 44: testing origin policy

domenic:
[Rendered](https://github.com/domenic/rfcs/blob/patch-1/rfcs/origin_policy.md)

### DIFF
--- a/rfcs/origin_policy.md
+++ b/rfcs/origin_policy.md
@@ -1,0 +1,26 @@
+# RFC #44: Testing origin policy
+
+## Summary
+
+[Origin policy](https://wicg.github.io/origin-policy/) is a proposed web platform feature that applies origin-wide settings. Writing web platform tests for origin policy is complicated, because:
+
+- The origin policy is a per-origin resource, located at `/.well-known/origin-policy`. To test different origin policy manifests, a test needs different responses to be served from that resource.
+- Origin policy settings apply origin-wide, which can cause interference with other tests running on that origin. For example, a CSP that disables image loading would interfere with any image-loading tests running on that origin.
+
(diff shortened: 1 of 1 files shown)


zcorpan:
This looks good to me, but is there a downside to using a greater number of subdomains from the start (like 99)? I'm thinking that there may be some churn to changing the subdomains, so it's better to do it fewer times and end up with more subdomains than we need, than to change them multiple times.

domenic:
Yeah, I think you're right; that'd probably be a better place to start. I will update the RFC and pull request.

jgraham:
Is there a requirement that each test uses a different origin? Do we have some way to enforce/lint for that?

domenic:
No, it's not. As long as the tests are careful to test non-overlapping features, they can share. You can see some examples of that in https://github.com/web-platform-tests/wpt/pull/21739 (which was for some reason closed....)

jgraham:
Hmm, having to be careful about testing non-overlapping features makes me nervous about conflicts between apparently unrelated tests (e.g. if two authors write tests at the same time for the same feature). However given that we can't make a wildcard certificate I don't currently have any good ideas for improvements here.

domenic:
OK, cool! It looks like it's been 7 days; would we be able to merge this---and more importantly, https://github.com/web-platform-tests/wpt/pull/21705 --- now?

gsnedders:
Note we previously had problems on Windows where IIRC we end up with all the hosts in an environment variable and this might take us over the maximum length of a variable.

domenic:
@gsnedders yes, that was dealt with in https://github.com/web-platform-tests/wpt/pull/21705#issuecomment-587578635

PR #48, opened 2020-03-26
RFC #48: Split 'assert_precondition' into optional and non-optional variants

stephenmcgruer:
See https://github.com/web-platform-tests/wpt/issues/21971 for background information

### DIFF
--- a/rfcs/assert_precondition_rename.md
+++ b/rfcs/assert_precondition_rename.md
@@ -5,8 +5,8 @@
 Split the `assert_precondition` function into `assert_implements_optional` and
 `assert_implements`. The `assert_implements_optional` function will behave as
 `assert_precondition` does today; a failed assert will record a
-'PRECONDITION\_FAILED' status. The `assert_implements` function will be
-synactical sugar for `assert_true`.
+'PRECONDITION\_FAILED' status. The `assert_implements` function behaves
+similarly, but records a `FAIL` status if the assert fails.
 
 Reuse the 'PRECONDITION\_FAILED' status rather than renaming it due to the
 technical difficulties of getting mozlog updated.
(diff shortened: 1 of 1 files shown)


rfcs/assert_precondition_rename.md
```diff
@@ -0,0 +1,58 @@
...
+
+The difference matters here due to the 'PRECONDITION\_FAILED' status, which has
+been interpreted by some users as a 'non-failing' status - which only makes
+sense if the failure is for truly OPTIONAL behavior.
+
+### Proposal
+
+Rename the `assert_precondition` function to `assert_implements_optional`, with
+the same 'PRECONDITION\_FAILED' status produced, and add a new function`
+assert_implements` that is essentially syntactical sugar for `assert_true`.
```
zcorpan:
```suggestion
the same 'PRECONDITION\_FAILED' status produced, and add a new function
`assert_implements` that is essentially syntactical sugar for `assert_true`.
```

rfcs/assert_precondition_rename.md
```diff
@@ -0,0 +1,58 @@
...
+That is:
+
+> __assert_implements(condition, description)__
+>
+> Concludes the test with a `FAIL` status if `condition` is not truthy.
```
zcorpan: Truthy isn't what `assert_true` does. `assert_true` does a strict check for `true`. I think checking truthy instead of strict true is better ergonomics for this, but we shouldn't say that the function is syntactical sugar for `assert_true` if it isn't.

stephenmcgruer: Whoops, great catch! I agree and fixed.

Hexcles:
Nit: I think we want to match the RFC number with the PR number, i.e. 48 for this one.

stephenmcgruer:
> Nit: I think we want to match the RFC number with the PR number, i.e. 48 for this one.

Oh is *that* how that works? (I could not figure out how the RFC numbers were chosen other than incrementing.) We should document that. Which will require a PR. Which will cause RFCs to be skipped. Hrm.

jgraham:
Skipping numbers isn't a problem, we won't run out ;) The point is not to know the order of RFCs being submitted, or have a simple mechanism to count RFCs, but to have a simple relationship between the RFC document and the associated discussion.

Hexcles:
PR numbers are not consecutive anyways because of issues.

stephenmcgruer:
Both very good points :). I'll send out a PR to document this, and update the text here.

rfcs/assert_precondition_rename.md
```diff
@@ -0,0 +1,59 @@
+# RFC 43: Split 'assert\_precondition' into optional and non-optional variants
```
stephenmcgruer:
```suggestion
# RFC 48: Split 'assert\_precondition' into optional and non-optional variants
```

stephenmcgruer:
>  I'll send out a PR to document this

SHS0t0ATMRs

Sigh. I'll be over here, in the corner with people who don't read the README properly -_-

stephenmcgruer:
With two approvals and no dissenting comments in the last 7 days, I am merging this per the RFC process.

PR #49, opened 2020-04-03
RFC 49: Python 3 migration for wptserve

Hexcles:
[Rendered](https://github.com/web-platform-tests/rfcs/blob/ee5eb6335c45da2357d6fc39fc8aab89ec6c6545/rfcs/wptserve_py3.md)

### DIFF
--- a/rfcs/wptserve_py3.md
+++ b/rfcs/wptserve_py3.md
@@ -33,7 +33,7 @@ Roughly in the order of importance:
 
 ## Root problem (or; why is this hard?)
 
-In both Python 2 and 3, unprefixed string literals have the `str` type. However, the meaning of the `str` type has changed: in Python 2 it is a raw byte sequence, whereas in Python 3 it is a Unicode text string. The rationale behind the change was to move to a Unicode-by-default world: arguably most people who used `str` in Python 2 were probably dealing with text and should use Unicode, so this decision minimized the impact on code handling human-readable text. However in WPT we actually intended to use byte sequences (see [background](#background)).
+In both Python 2 and 3, unprefixed string literals have the `str` type. However, the meaning of the `str` type has changed: in Python 2 it is a raw byte sequence, whereas in Python 3 it is a Unicode text string. The rationale behind the change was to move to a Unicode-by-default world: arguably most people who used `str` in Python 2 were probably dealing with text and should use Unicode, so this decision minimized the impact on code handling human-readable text. However in wptserve we often intended to use byte sequences (see [background](#background)).
 
 This pain is further compounded by the choice of types in some parts of the Python 3 standard library. The low-level [`http`](https://docs.python.org/3/library/http.html) library takes and returns the `str` type (i.e. Unicode string) almost everywhere except the body (see the [type](https://github.com/python/typeshed/blob/master/stdlib/3/http/client.pyi) [annotations](https://github.com/python/typeshed/blob/master/stdlib/3/http/server.pyi)). This is convenient for existing code from Python 2 but semantically different and makes it difficult to send raw bytes on the wire. On the upside, the standard library carefully and intentionally uses an 8-bit isomorphic encoding (latin-1/ISO-8859-1) when appropriate (e.g. [headers](https://github.com/python/cpython/blob/8c3ab189ae552401581ecf0b260a96d80dcdae28/Lib/http/client.py#L220)) to allow arbitrary byte sequences:
 
@@ -52,15 +52,17 @@ The preferred approach of the community is to keep a consistent and semantically
 wptserve changes:
 
 *   Introduce a pair of helper functions `isomorphic_{encode,decode}` that use ISO-8859-1 under the hood and only encode/decode if necessary; both of them can accept either binary or text strings, but always return binary/text strings respectively regardless of the Python version.
-*   Most public APIs for custom handlers (including methods and properties of `Request` and `Response`) are binary, with the notable exception of the response body which can be text or binary strings.
+*   Most public APIs for custom handlers (including methods and properties of `Request` and `Response`) only accept and return binary, with the notable exception of the response body which can be text or binary strings (if a text string is given, it will be encoded with `Response.encoding` automatically).
 
-Guidelines for converting custom handlers:
+[Python file handlers](https://web-platform-tests.org/writing-tests/python-handlers/index.html?highlight=file%20handler):
 
-*   Make sure all strings are either always text or always bytes; all string literals in handlers should be prefixed with `b""` or `u""`. This will be enforced through a [lint](https://github.com/web-platform-tests/wpt/pull/23347). Existing code needs to be converted and reviewed manually ([example](https://github.com/web-platform-tests/wpt/pull/23363)). Note that we could add a hack to setters to accept both `str` and `bytes`, but on the getter side there is little we could do -- we don’t know whether the returned value will be used together with `bytes` or `str`, and things like `dict.get()` will be particularly tricky.
+*   All strings will be either always text or always bytes; all string literals in handlers should be prefixed with `b""` or `u""`.
+    * This will be enforced through a [lint](https://github.com/web-platform-tests/wpt/pull/23347).
+    * Existing code needs to be converted and reviewed manually ([example](https://github.com/web-platform-tests/wpt/pull/23363)).
 *   **Headers** on both requests and responses are always *binary strings* for both keys and values. This includes:
     * **HTTP Basic Auth**: `Request.auth.{username,password}` are binary strings.
     * **Cookies**: Names and values in cookies are always binary strings, including `Request.cookies` and `Response.{set,unset,delete}_cookie`.
-*   **Request URL/form parameters** are all *binary strings* for both keys and values, accessible via `Request.GET` or `Request.POST`.
+*   **Request URL/form parameters** are always *binary strings* for both keys and values, accessible via `Request.GET` or `Request.POST`.
 *   **Response body** can be written via `Response.writer.*` or via the [return values](https://web-platform-tests.org/tools/wptserve/docs/handlers.html#python-handlers), which can be either text or binary strings, but the two types should never be mixed and string literals must be prefixed.
     * As an exception, `Response.set_error` takes only text message strings.
 
--- a/rfcs/wptserve_py3.md
+++ b/rfcs/wptserve_py3.md
@@ -64,7 +64,7 @@ wptserve changes:
     * **Cookies**: Names and values in cookies are always binary strings, including `Request.cookies` and `Response.{set,unset,delete}_cookie`.
 *   **Request URL/form parameters** are always *binary strings* for both keys and values, accessible via `Request.GET` or `Request.POST`.
 *   **Response body** can be written via `Response.writer.*` or via the [return values](https://web-platform-tests.org/tools/wptserve/docs/handlers.html#python-handlers), which can be either text or binary strings, but the two types should never be mixed and string literals must be prefixed.
-    * As an exception, `Response.set_error` takes only text message strings.
+    * As an exception, `Response.set_error` takes only text message strings, as this message is used in the JSON response.
 
 #### Outcomes and risks
 
--- a/rfcs/wptserve_py3.md
+++ b/rfcs/wptserve_py3.md
@@ -3,7 +3,7 @@
 
 ## Summary
 
-This is a proposal to always use the native string type `str` in wptserve APIs in both Python 2 and 3 for a smoother and more ergonomic migration towards Python 2 & 3 compatibility.
+This is a proposal to always use string types with consistent semantics (mostly `six.binary_type` with `six.text_type` in rare cases) and never the native `str` type in wptserve APIs in both Python 2 and 3.
 
 
 ## Background
@@ -24,9 +24,11 @@ This is a rough count of all Python files excluding some directories, which stil
 
 Roughly in the order of importance:
 
-1.  Everything, including wptserve and custom handlers, must continue to work in both Python 2 and 3 consistently. (This was discussed and agreed upon at TPAC 2019.)
+1.  Everything, including wptserve and custom handlers, must continue to work in both Python 2 and 3. (This was discussed and agreed upon at TPAC 2019.)
 1.  Custom handlers must continue to be able to read and write raw bytes from requests and to responses, including headers and body. wptserve must not change the byte sequences in any way; no errors should be raised if the byte sequences do not have a valid encoding.
-1.  _Optional_: minimize code change to custom handlers, and consequently minimize the burden on test authors in the future by imposing fewer new requirements. (Whether to strive for this goal will largely affect which option we choose.)
+1.  _Optional/competing_: (Which goal to prioritize largely determines which option we choose.)
+    a. Keep type semantics consistent across Python 2 and 3 (a string is either always binary or always text).
+    a. Minimize code change to custom handlers, and consequently minimize the burden on test authors in the future by imposing fewer new requirements.
 
 
 ## Root problem (or; why is this hard?)
@@ -41,15 +43,41 @@ This pain is further compounded by the choice of types in some parts of the Pyth
 
 ## Options
 
+### [Preferred] Keep the semantics: always use byte sequences
 
-### [Recommended] Keep the ergonomics: always use str
+The preferred approach of the community is to keep a consistent and semantically correct encoding model. That is to **always use byte sequences: `str` in Python 2 and `bytes` in Python 3** (with some rare exceptions below).
+
+#### Detailed changes
+
+wptserve changes:
+
+*   Introduce a pair of helper functions `isomorphic_{encode,decode}` that use ISO-8859-1 under the hood and only encode/decode if necessary; both of them can accept either binary or text strings, but always return binary/text strings respectively regardless of the Python version.
+*   Most public APIs for custom handlers (including methods and properties of `Request` and `Response`) are binary, with the notable exception of the response body which can be text or binary strings.
+
+Guidelines for converting custom handlers:
+
+*   Make sure all strings are either always text or always bytes; all string literals in handlers should be prefixed with `b""` or `u""`. This will be enforced through a [lint](https://github.com/web-platform-tests/wpt/pull/23347). Existing code needs to be converted and reviewed manually ([example](https://github.com/web-platform-tests/wpt/pull/23363)). Note that we could add a hack to setters to accept both `str` and `bytes`, but on the getter side there is little we could do -- we don’t know whether the returned value will be used together with `bytes` or `str`, and things like `dict.get()` will be particularly tricky.
+*   **Headers** on both requests and responses are always *binary strings* for both keys and values. This includes:
+    * **HTTP Basic Auth**: `Request.auth.{username,password}` are binary strings.
+    * **Cookies**: Names and values in cookies are always binary strings, including `Request.cookies` and `Response.{set,unset,delete}_cookie`.
+*   **Request URL/form parameters** are all *binary strings* for both keys and values, accessible via `Request.GET` or `Request.POST`.
+*   **Response body** can be written via `Response.writer.*` or via the [return values](https://web-platform-tests.org/tools/wptserve/docs/handlers.html#python-handlers), which can be either text or binary strings, but the two types should never be mixed and string literals must be prefixed.
+    * As an exception, `Response.set_error` takes only text message strings.
+
+#### Outcomes and risks
+
+*   Almost every line in the existing handlers will need to be modified to add prefixes to string literals. This is a huge undertaking.
+*   We need to educate test authors and reviewers about the aforementioned coding guidelines to make sure string literals are always prefixed correctly.
+*   Printing and logging will also need to be changed in Python 3. Custom handlers need to have a special code path for Python 3 to encode the strings.
+
+
+### [Alternative] Keep the ergonomics: always use str
 
 I would argue that the most ergonomic approach (for test authors) is to **always use unprefixed string literals and the `str` type** in the majority of cases regardless of Python 2 or 3 despite the different semantics for the following reasons:
 
 *   It avoids having to modify hundreds of custom handlers (most likely manually).
 *   It does not impose additional burden on test authors (e.g. remembering to prefix string literals with b even if it is a no-op in Python 2, which will still remain the default for a while). In addition, any additional requirement on types would be impossible to fully enforce in theory since Python is dynamically typed.
 *   This is what the Python community has chosen to do, so it aligns well with the standard library without additional en/decoding that depends on the version of Python. Custom handlers do not need `ensure_str`/`maybe_encode` or have different code paths for Python 2 and 3.
-
 
 #### Detailed changes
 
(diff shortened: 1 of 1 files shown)


Hexcles:
Tagging people who have been involved with Python 3 migration: @annevk @jgraham @stephenmcgruer @svillar @ziransun

tabatkins:
I strongly suspect you're going to run into a lot of pernicious, annoying issues with this. You probably want to instead be stricter about always using bytes; any time I, and many other people with similar problems, have tried to be lazy about bytes vs strings, it's bitten us hard and been extremely annoying and frustrating.

I don't think there's a way to make Python3 emit bytes by default for untagged strings, but you can force Python2 to allow the `b` modifier, same as Py3. Unfortunately, that change causes untagged strings to switch to unicode, same as Py3, so it would imply some migration costs.

Hexcles:
Thanks for the feedback, Tab. Yes I'm aware of the potential issues (see the [risks](https://github.com/web-platform-tests/rfcs/blob/6c4e1223a8af2ba71905b61d8905cf072eb6ecce/rfcs/wptserve_py3.md#outcomes-and-risks)) and am trying to make conscious and careful trade-offs here. For the core use case (reading and serving bytes as is on the wire), my investigation suggests that most custom handlers would work without modifications. wptserve is different from bikeshed as it doesn't interact with "text" much, so we could get away with treating arbitrary data as latin-1-encoded strings, even though the semantics are wrong (printing, counting the length, etc. are broken). Yet we can't be sure until we try.

I was in favour of using bytes everywhere as I believed (and still believe) that'd be the "correct" way, e.g. in https://github.com/web-platform-tests/wpt/pull/13246 , but I've become convinced that it'd require a herculean effort to migrate the hundreds of custom handlers and prevent future regression. As you said, there's no way to make unprefixed strings bytes in Py3 and the only way out would be to use the `b` prefix (which also works / is a no-op in Python 2) when needed (you can't just add `b` to every string literal). I don't think it's possible to check the usage programatically using lint or more advanced static analysis, which means 1. humans need to examine and modify the existing code and 2. humans need to review all future code. Some things will be easy to catch (e.g. comparing different types), while others may be tricky (e.g. wrong type in `dict.get()`). If we didn't need to consider Python 2 compatibility, I'd probably still prefer this approach.

On top of the migration/maintenance cost, I also considered how the standard library and a bunch of popular HTTP-related libraries handle headers. All of them chose to use `str` instead of trying to ensure bytes. I feel like I was trying to fight the standard library, wrapping the API calls or even as @jgraham joked forking it.

jgraham:
I need to look more at this on Monday, but my initial reaction is that I'm very concerned with the proposal.  I think using the `str` type is going to cause a bunch of hard-to-understand issues.

My proposal to "fork the stdlib" wasn't a joke, but also wasn't as far-reaching as that implied; it would specifically mean not using the stdlib parts we currently use for reading and writing HTTP requests specifically.

annevk:
If we allow strings these "technically byte" APIs better throw for code points higher than U+00FF. I would favor using bytes though so it's clearer what's going on. (There are a lot of tests where the specifics around this matter.)

Hexcles:
> If we allow strings these "technically byte" APIs better throw for code points higher than U+00FF.

That's a good point. Eventually this would throw when calling `encode('latin-1')`:
> UnicodeEncodeError: 'latin-1' codec can't encode character '\u01ff' in position 0: ordinal not in range(256)
But we could add assertions to make this easier to understand.

> There are a lot of tests where the specifics around this matter.

If you look at these [examples](https://github.com/web-platform-tests/rfcs/blob/6a3f5a1c87e44b2805eafeb1a478635a8454c71a/rfcs/wptserve_py3.md#outcomes-and-risks), would those suffice the testing purpose? My cursory scan of the code base suggests that the second form is the most common. Or does it confuse you cognitively that this is a text string in Python 3?

annevk:
I'm probably not the best person to ask that question as I'm intimately familiar with HTTP and text encodings. ("latin-1" might be confusing to some as on the web it means windows-1252.)

jgraham:
OK, I've read this in more detail now, so I think I understand the tradeoffs being made, although I wouldn't say I yet have a good feeling for whether they're the right ones to make. The suggested approach is the exact opposite to the one we've been taking with all the internal code (where we have tried to be clear about what's Text and what's Bytes and use those types consistently between 2 and 3 where possible). But it's of course possible that the different constraints here wind up giving us a different optimal solution. Having said that I also think that our requirements are very different to normal HTTP libraries, so the prior art there isn't as relevant as it might seem.

I wonder if we can enumerate all the API surface we have that deal with strings and are interfacing with the stdlib. I don't think there are so many places where we *directly* expose the stdlib and in those places it may be possible to wrap-or-replace. So it at least seems conceviable that the concern that we might be able to mediate access and provide a better API.

I also don't think that linting for bytes-vs-text is the only possible way to enforce that we get the correct types; runtime asserts for example would cause tests to error and we could ensure that we get a useful output.

Hexcles:
> I wonder if we can enumerate all the API surface we have that deal with strings and are interfacing with the stdlib.

With the help of typeshed, we should be able to get that list relatively easily.

> I don't think there are so many places where we directly expose the stdlib and in those places it may be possible to wrap-or-replace. So it at least seems conceviable that the concern that we might be able to mediate access and provide a better API.

I agree that cleaning up our API surface wouldn't be hard, but my main concern is still around the custom handlers.

> runtime asserts for example would cause tests to error and we could ensure that we get a useful output.

Is there a way to implement such asserts in Python 2? AFAIK, you can't actually tell the difference between `b"..."` and `"..."` at run time, which means we kinda have to bring up the whole Python 3 CI before being able to enforce such rules.

svillar:
> I was in favour of using bytes everywhere as I believed (and still believe) that'd be the "correct" way, e.g. in [web-platform-tests/wpt#13246](https://github.com/web-platform-tests/wpt/pull/13246) , but I've become convinced that it'd require a herculean effort to migrate the hundreds of custom handlers and prevent future regression. As you said, there's no way to make unprefixed strings bytes in Py3 and the only way out would be to use the `b` prefix (which also works / is a no-op in Python 2) when needed (you can't just add `b` to every string literal). I don't think it's possible to check the usage programatically using lint or more advanced static analysis, which means that humans need to examine and modify the existing code (whereas in the recommended approach, we could fix just the handlers that throw and the tests that produce different results). If we didn't need to consider Python 2 compatibility, I'd probably still prefer this approach.

From my personal experience migrating the code in `tools/`, static analysis won't help in this case as you mentioned. Python is a dynamic language after all.

After reading your proposal I'm wondering whether the "herculean" effort you mention is about complexity or quantity. If we're talking about the former then I'd agree that making some trade-offs and KISS might be the right approach. But if the concern is the amount of required changes, then I think we should seriously consider the alternative of dealing always with bytes. These kind of "mechanical" changes allow great levels of parallelization (many people working on them at the same time) and after some time it'd be relatively easy to foresee the amount of time we'd need for the transition.

jgraham:
> Is there a way to implement such asserts in Python 2? AFAIK, you can't actually tell the difference between b"..." and "..." at run time, which means we kinda have to bring up the whole Python 3 CI before being able to enforce such rules.

We can fullt control the parsing of the code here. So we can, if we want, do things like fail the parse for unprefixed literals, or treat all literals as bytestrings unless otherwise specified (c.f. https://www.mercurial-scm.org/repo/hg/rev/1c22400db72d)

Hexcles:
> I'm probably not the best person to ask that question as I'm intimately familiar with HTTP and text encodings. ("latin-1" might be confusing to some as on the web it means windows-1252.)

@annevk Yeah we can use the more technically correct term `iso-8859-1`.

In addition to the scale, I'm concerned by the upkeeping after the massive manual conversion (which is challenging, but as @svillar pointed out, can be parallelized). I think if we were to do this, we would need to either hook into the loading mechanism to check for regression (see @jgraham 's comment above, but probably not overwrite) or set up a full Python 3 CI to compare (a subset of) results (which requires us to get almost everything else working in Python 3 first). @svillar @ziransun IIRC, we can already run `wpt run` and produce the JSON in Python 3 now? I'd very much prefer running Python 3 as opposed to hooking into the loading mechanism.

stephenmcgruer:
> produce the JSON in Python 3 now

What do you mean by 'produce the JSON'?

Hexcles:
Ahh sorry I meant `wptreport.json`, which would allow us to compare results against Python 2.

jgraham:
Note that we don't have to do the hg thing of actually altering the source, we can do something like tokenize the source using the built-in `tokenize` module to validate that all strings are specified with a prefix. We can also make that opt-in during the transition with a magic comment that we remove once we've converted all the existing callers. That means we can make things fail before tests even reach CI which reduces the number of cases where we have to explain how to interpret CI results.

Hexcles:
I think we (including me) don't have enough concrete evidence (e.g. regarding how much work the bytes conversion in custom handlers would require), and it's clear from the discussions so far that most people prefer to have clear and correct semantics, so I'm proposing that we try the "alternative approach", concretely:

* I will double check the wptserve APIs to make sure they always return `bytes` (note that this is basically the status quo, so I may fix the API which doesn't currently always return bytes without landing a RFC).
* I will propose a conversion guideline for custom handlers.
* We then try converting a bunch of custom handlers. I'm thinking maybe `html/`, in other words try https://github.com/web-platform-tests/wpt/pull/22402 again with the new guideline, if @annevk doesn't mind reviewing, or some other people volunteer.

At that point, we can consider how we'd like to proceed. We'd need both consensus around the approach and commitment to work (writing and reviewing the PRs).

I consider this a relatively low-cost experiment and we can always come back to the other approach. WDYT?

jgraham:
I'm happy with that plan. I think we do need a way to enforce that we get correct semantics going forward; presumably this comes under "conversion guidelines" (I also agree it may be premature to actually implement the enforcement mechansim until such a time as there's a clear decision on which path to take).

stephenmcgruer:
This plan SGTM, with one caveat/nit-pick:

> We then try converting a bunch of custom handlers. I'm thinking maybe html/, in other words try web-platform-tests/wpt#22402 again with the new guideline, if @annevk doesn't mind reviewing, or some other people volunteer.

I am concerned that `html/` is too large for a trial run, since the 'only accept bytes' approach requires us to examine and convert every single file. By my rough count, that's 50 files in `html/`:

```
$ find html -name "*.py" -not -path "*/.virtualenv/*" | wc -l
50
```

To me a trial would be more like: pick 5-10 Python handler files at random, and try those.

Hexcles:
I clearly did not count the files :) I thought the PR included everything in html. I was also thinking that trial should be about the same size as that PR (<10 files).

annevk:
@Hexcles iso-8859-1 means windows-1252 on the web too. See https://encoding.spec.whatwg.org/. In standards we use https://infra.spec.whatwg.org/#isomorphic-decode and https://infra.spec.whatwg.org/#isomorphic-encode.

Happy to help review.

Hexcles:
Thanks, @annevk . I'll use "isomorphic decode/encode"; we don't really care about the character sets.

Sorry for the delay. I've done a cursory check of `wptserve` and found few issues (I'll send some PRs later). Keys and values in headers are always byte sequences (though the getter accepts both text and byte strings, so `headers.get("cookies")` works in both Python versions).

I think we can start experimenting converting a few custom handlers (I suggest we try those in https://github.com/web-platform-tests/wpt/pull/22402 first), @ziransun @stephenmcgruer . I'm proposing the following guidelines:
* **Headers**: always use *binary strings* for both keys are values; prefer adding `b""` prefixes to encoding/decoding the other operands. This also applies to basic auth.
* **Request params**: both GET and POST params should be *text strings* (we already percent decode these values). Note that they are currently native strings and work is needed to make them text in Python 2.
* **Response body**: always use *text strings*. The `Response` class defaults to `utf-8`. If the handler needs other encoding, it probably already has explicitly specified the encoding. This implies that string literals should be prefixed with `u""` (although native strings will also work in most places).

stephenmcgruer:
Thanks @Hexcles ! How about you and I work today to put together a short conversion guideline?

> I suggest we try those in web-platform-tests/wpt#22402 first

To be clear, my reading of this is that you suggest we try the exact 8 files touched by that PR (as opposed to all the handlers in `html/`).

jgraham:
> Response body: always use text strings. The Response class defaults to utf-8. If the handler needs other encoding, it probably already has explicitly specified the encoding.

Unless I'm missing something, that's not quite true; if we're given binary data in the `Response` class we [pass it through](https://searchfox.org/mozilla-central/source/testing/web-platform/tests/tools/wptserve/wptserve/response.py#631-638). That seems like reasonable and sensible behaviour we should keep, but it does mean things could be different between py2 and py3 unless we force people to annotate strings with either `u` or `b`.

Hexcles:
@jgraham I stand corrected. I've updated the guidelines above. Any other concerns?

Hexcles:
Actually, if we make all request params binary, we might as well make cookies binary, too, since they share the same underlying data structure (`MultiDict`) and the same rationale also applies.

Note that this will break a lot of things in Python 3, pushing us backward in terms of test pass rates in Python 3.

stephenmcgruer:
On request parameters; making them bytes-only is going to hurt in terms of work-needing-done, since it will break every use of stash (will need to either make stash take bytes, or decode everything being put in stash), it will break templating (I guess we'll fix that in the template handlers), and there's just a lot of places in file handlers that take parameters and pass them to other code (e.g. string operations).

That said, I think we do need to do it for consistent semantics. Would be interested if the file-handlers experts (well, basically @annevk) agree.

annevk:
For representing the value of the name query parameter in the URL `/?name=%E2%82%AC` either `b"\xE2\x82\xAC"` or `u"\u00E2\u0082\u00AC"` are reasonable, but `u"\u20AC"` (result from UTF-8 decode) would not be. `b"\x80"` would be a bit clearer as in the `u` case the upper bound for a code point would have to be `\u00FF`. Hope that helps.

annevk:
@Hexcles there's a dedicated cookie API for `Set-Cookie`? If so, it definitely would be most useful if that can set any kind of header value as otherwise a lot of cookie-related tests would simply have to use the headers API.

Hexcles:
@annevk Thanks! I don't like `u"\u00E2\u0082\u00AC"` so I'll go with `b"\xE2\x82\xAC"`.

There's https://web-platform-tests.org/tools/wptserve/docs/request.html#wptserve.request.Request.cookies for dealing with cookies.

Hexcles:
I've merged https://github.com/web-platform-tests/wpt/pull/23284 and https://github.com/web-platform-tests/wpt/pull/23327 . Now almost everything in wptserve is binary string, with the notable exception of URLs/paths.

@ziransun 's https://github.com/web-platform-tests/wpt/pull/23363 (not ready for review yet) will give us an idea of how the migration will look like.

stephenmcgruer:
As I understand it, https://github.com/web-platform-tests/wpt/pull/23363/files is now ready for review and should be able to give an idea of what the outcome of this RFC would look like. Please take a look, all.

Hexcles:
> what the outcome of this RFC would look like

Minor correction: what the outcome of the "alternative" approach in the RFC as of now would look like (I haven't updated the RFC yet)

Hexcles:
Hi everyone on this thread, I've updated the RFC to reflect our latest consensus, switching the previously recommended and alternative approaches. I also incorporated the guidelines from the earlier [comment](https://github.com/web-platform-tests/rfcs/pull/49#issuecomment-619733129), which has been tested in this [trail PR](https://github.com/web-platform-tests/wpt/pull/23363), into the RFC.

Please take a look again. Thanks!

rfcs/wptserve_py3.md
```diff
@@ -0,0 +1,121 @@
...
+1.  _Optional/competing_: (Which goal to prioritize largely determines which option we choose.)
+    a. Keep type semantics consistent across Python 2 and 3 (a string is either always binary or always text).
+    a. Minimize code change to custom handlers, and consequently minimize the burden on test authors in the future by imposing fewer new requirements.
+
+
+## Root problem (or; why is this hard?)
+
+In both Python 2 and 3, unprefixed string literals have the `str` type. However, the meaning of the `str` type has changed: in Python 2 it is a raw byte sequence, whereas in Python 3 it is a Unicode text string. The rationale behind the change was to move to a Unicode-by-default world: arguably most people who used `str` in Python 2 were probably dealing with text and should use Unicode, so this decision minimized the impact on code handling human-readable text. However in WPT we actually intended to use byte sequences (see [background](#background)).
```
stephenmcgruer, line 36: Nit suggestion: s/WPT/wptserve and s/actually/often

```suggestion
In both Python 2 and 3, unprefixed string literals have the `str` type. However, the meaning of the `str` type has changed: in Python 2 it is a raw byte sequence, whereas in Python 3 it is a Unicode text string. The rationale behind the change was to move to a Unicode-by-default world: arguably most people who used `str` in Python 2 were probably dealing with text and should use Unicode, so this decision minimized the impact on code handling human-readable text. However in wptserve we often intended to use byte sequences (see [background](#background)).
```

rfcs/wptserve_py3.md
```diff
@@ -0,0 +1,121 @@
...
+
+This pain is further compounded by the choice of types in some parts of the Python 3 standard library. The low-level [`http`](https://docs.python.org/3/library/http.html) library takes and returns the `str` type (i.e. Unicode string) almost everywhere except the body (see the [type](https://github.com/python/typeshed/blob/master/stdlib/3/http/client.pyi) [annotations](https://github.com/python/typeshed/blob/master/stdlib/3/http/server.pyi)). This is convenient for existing code from Python 2 but semantically different and makes it difficult to send raw bytes on the wire. On the upside, the standard library carefully and intentionally uses an 8-bit isomorphic encoding (latin-1/ISO-8859-1) when appropriate (e.g. [headers](https://github.com/python/cpython/blob/8c3ab189ae552401581ecf0b260a96d80dcdae28/Lib/http/client.py#L220)) to allow arbitrary byte sequences:
+
+*   Getting bytes out: “encode” the return value with latin-1
+*   Passing bytes in: “decode” the byte sequence with latin-1
+
+
+## Options
```
stephenmcgruer, line 44: For clarity, I think we want to split these more clearly. This one should be something like `Proposed solution` (top-level, replacing 'Options'), and the other can move down to an Appendix or Alternatives with a note that it was not chosen (and why).

rfcs/wptserve_py3.md
```diff
@@ -0,0 +1,121 @@
...
+### [Preferred] Keep the semantics: always use byte sequences
+
+The preferred approach of the community is to keep a consistent and semantically correct encoding model. That is to **always use byte sequences: `str` in Python 2 and `bytes` in Python 3** (with some rare exceptions below).
+
+#### Detailed changes
+
+wptserve changes:
+
+*   Introduce a pair of helper functions `isomorphic_{encode,decode}` that use ISO-8859-1 under the hood and only encode/decode if necessary; both of them can accept either binary or text strings, but always return binary/text strings respectively regardless of the Python version.
+*   Most public APIs for custom handlers (including methods and properties of `Request` and `Response`) are binary, with the notable exception of the response body which can be text or binary strings.
```
stephenmcgruer, line 55: Do we explain why 'response body' can be text or binary anywhere?

stephenmcgruer, line 55: 'are binary' - should that be something like 'only accept and return binary'?

rfcs/wptserve_py3.md
```diff
@@ -0,0 +1,121 @@
...
+
+Guidelines for converting custom handlers:
```
stephenmcgruer, line 57:
```suggestion
[Python file handlers](https://web-platform-tests.org/writing-tests/python-handlers/index.html?highlight=file%20handler):
```

rfcs/wptserve_py3.md
```diff
@@ -0,0 +1,121 @@
...
+
+*   Make sure all strings are either always text or always bytes; all string literals in handlers should be prefixed with `b""` or `u""`. This will be enforced through a [lint](https://github.com/web-platform-tests/wpt/pull/23347). Existing code needs to be converted and reviewed manually ([example](https://github.com/web-platform-tests/wpt/pull/23363)). Note that we could add a hack to setters to accept both `str` and `bytes`, but on the getter side there is little we could do -- we don’t know whether the returned value will be used together with `bytes` or `str`, and things like `dict.get()` will be particularly tricky.
```
stephenmcgruer, line 59: Suggest: change the first sentence slightly, and removed the ' Note that we could add a hack to setters to accept both `str` and `bytes`, but on the getter side there is little we could do -- we don’t know whether the returned value will be used together with `bytes` or `str`, and things like `dict.get()` will be particularly tricky.' because it doesn't seem relevant - we aren't adding this sort of helper, right?

```suggestion
*   All strings will be either always text or always bytes; all string literals in handlers should be prefixed with `b""` or `u""`.
    * This will be enforced through a [lint](https://github.com/web-platform-tests/wpt/pull/23347).
    * Existing code needs to be converted and reviewed manually ([example](https://github.com/web-platform-tests/wpt/pull/23363)).
```

rfcs/wptserve_py3.md
```diff
@@ -0,0 +1,121 @@
...
+*   **Headers** on both requests and responses are always *binary strings* for both keys and values. This includes:
+    * **HTTP Basic Auth**: `Request.auth.{username,password}` are binary strings.
+    * **Cookies**: Names and values in cookies are always binary strings, including `Request.cookies` and `Response.{set,unset,delete}_cookie`.
+*   **Request URL/form parameters** are all *binary strings* for both keys and values, accessible via `Request.GET` or `Request.POST`.
```
stephenmcgruer, line 63:
```suggestion
*   **Request URL/form parameters** are always *binary strings* for both keys and values, accessible via `Request.GET` or `Request.POST`.
```

rfcs/wptserve_py3.md
```diff
@@ -0,0 +1,121 @@
...
+*   **Response body** can be written via `Response.writer.*` or via the [return values](https://web-platform-tests.org/tools/wptserve/docs/handlers.html#python-handlers), which can be either text or binary strings, but the two types should never be mixed and string literals must be prefixed.
+    * As an exception, `Response.set_error` takes only text message strings.
```
stephenmcgruer, line 65: Can we briefly summarize why? (If not brief, then don't)

rfcs/wptserve_py3.md
Hexcles:
```suggestion
*   Most public APIs for custom handlers (including methods and properties of `Request` and `Response`) only accept and return binary, with the notable exception of the response body which can be text or binary strings (if a text string is given, it will be encoded with `Response.encoding` automatically).
```

rfcs/wptserve_py3.md
Hexcles:
```suggestion
    * As an exception, `Response.set_error` takes only text message strings, as this message is used in the JSON response.
```

foolip:
@Hexcles we discussed a method for comparing the the responses of custom handlers between python 2 and 3. Is a script for that something you expect to have in place when we begin making/reviewing changes?

Hexcles:
@foolip I'll look into that in parallel, but don't want to wait for / block on that idea.

stephenmcgruer:
As far as I know, we are now at 10 days without any objections; per the RFC process I am going to merge this later this afternoon.

PR #50, opened 2020-04-09
Update and rename load_test.md to reflect reality

Hexcles:
* We eventually named it "crashtest".
* The class name implemented was `test-wait`.

I'm not sure whether it was an oversight or something changed later. I think we should fix the RFC (or maybe retroactively add another one, which seems a bit much) to avoid the confusion.


Hexcles:
Even though this is not a formal RFC, I'll still leave it open for a while in case anyone has concerns.

Hexcles:
Alright, I'm merging this.

PR #51, opened 2020-04-20
RFC 51: Rename lint.whitelist to lint.ignore

stephenmcgruer:


zcorpan:
I support this change. But:

* The title says `.ignorelist` but the text says `.ignore`. Please pick one. :)
* It would be good to understand the impact of downstream projects better.

foolip:
It's now `lint.ignore` everywhere, and I think that looks good!

rfcs/rename_lint_whitelist.md
```diff
@@ -0,0 +1,25 @@
...
+[web-platform-tests.org](https://web-platform-tests.org) documentation will
+need updating.
+
+## Risks
+
+Known risks:
+
+1. Downstream consumers or infrastructure may rely on the name of our lint
+   file. The level of risk here is unknown, but not expected to be high - it
+   seems unlikely the WPT lint is a critical file.
```
stephenmcgruer, line 24:
```suggestion
1. Downstream consumers or infrastructure may rely on the name of our lint
   file. Neither Gecko and Chromium appear to rely on it, and so the risk here
   is low.
```

stephenmcgruer:
I have updated the RFC with the note about Gecko and Chromium; hopefully that addresses @zcorpan's concerns.

PR #52, opened 2020-04-20
RFC 52: Remove default and negated global support in *.any.js

annevk:


rfcs/any.js-globals.md
```diff
@@ -0,0 +1,14 @@
+# RFC ..: Remove default and negated global support in *.any.js
```
stephenmcgruer, line 1:
```suggestion
# RFC 52: Remove default and negated global support in *.any.js
```

rfcs/any.js-globals.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Summary
+
+https://web-platform-tests.org/writing-tests/testharness.html#multi-global-tests has a confusing syntax for specifying globals. E.g., `global=window` is equivalent to `global=window,dedicatedworker` and `global=jsshell` is equivalent to `global=window,dedicatedworker,jsshell`.
+
```
stephenmcgruer, line 6: Can you add a paragraph on what the new world will look like? (That is, what this RFC will actually change). E.g.

1. globals must be listed explicitly  for the test be run in them
1. if no global meta is specified then the default will apply (maybe specify exactly will the default will be)
1. some change involving negation? I'm not actually sure what you intend to change there.

rfcs/any.js-globals.md
```diff
@@ -0,0 +1,14 @@
+# RFC ..: Remove default and negated global support in *.any.js
+
+## Summary
+
+https://web-platform-tests.org/writing-tests/testharness.html#multi-global-tests has a confusing syntax for specifying globals. E.g., `global=window` is equivalent to `global=window,dedicatedworker` and `global=jsshell` is equivalent to `global=window,dedicatedworker,jsshell`.
```
stephenmcgruer, line 5:
```suggestion
https://web-platform-tests.org/writing-tests/testharness.html#multi-global-tests has a confusing syntax for specifying globals. The default set of globals is always included implicitly, e.g., `global=window` is equivalent to `global=window,dedicatedworker` and `global=jsshell` is equivalent to `global=window,dedicatedworker,jsshell`.
```

rfcs/any.js-globals.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Details
+
+1. Update all `*.any.js` tests that use `// META: global=` to list all the relevant globals explicitly when possible. (Some necessarily rely on `!default` or `!` (negation).
+2. Atomically:
+   1. Remove support for `default` and `!`.
+   1. Update all remaining `*.any.js` tests that use `!` to list all the relevant globals explicitly.
+   
+https://github.com/web-platform-tests/wpt/issues/23111 tracks this work.
```
stephenmcgruer, line 14: Are there any risks that should be detailed here? I think the main one is just downstream implementers as usual. I'm happy to own checking the impact this will have on chromium.

annevk: I guess the risk is getting the changes wrong. I don't see any downstream risks personally, but I'm not aware of all downstream usage either.

annevk:
Note that I was partially responsible for this mess in the first place. https://github.com/web-platform-tests/wpt/issues/4210 and https://github.com/web-platform-tests/wpt/pull/6792 have history. It seems that we didn't really take a step back at any point to consider what would be pragmatic and useful here.

rfcs/any.js-globals.md
jgraham: The downstream risk would be implementors who are not using the upstream infrastructure who could have reimplemented part of the logic here. I'm not specifically aware of anyone doing that (gecko doesn't) but it's at least possible.

rfcs/any.js-globals.md
```diff
@@ -0,0 +1,25 @@
...
+This RFC will make these changes:
+
+1. Remove `default` (though keep `window,dedicatedworker` as the default if `// META: global=` is not present).
+2. Remove `!` (negation). Instead test developers are expected to either rely on the default or explicitly list the globals they want.
+
+A test supposed to run in dedicated and shared workers would use `dedicatedworker,sharedworker` and not `!window,sharedworker`, `!window,worker,!serviceworker`, etc.
+
+## Details
+
+1. Update all `*.any.js` tests that use `// META: global=` to list all the relevant globals explicitly when possible. (Some necessarily rely on `!default` or `!`.
```
stephenmcgruer, line 16:
```suggestion
1. Update all `*.any.js` tests that use `// META: global=` to list all the relevant globals explicitly when possible. (Some necessarily rely on `!default` or `!`).
```

annevk: The closing parenthesis should go after the dot in this case.

stephenmcgruer: (Fixed.)

PR #53, opened 2020-05-11
RFC 53: testharness.js timeout detection

jugglinmike:
[Rendered](https://github.com/bocoup/wpt-rfcs/blob/timeout-deferral/rfcs/timeout-deferral.md)

### DIFF
--- a/rfcs/timeout-deferral.md
+++ b/rfcs/timeout-deferral.md
@@ -1,0 +1,61 @@
+# RFC 53: Introduce more robust mechanism for testharness.js timeout detection
+
+## Summary
+
+Authors of long-running testharness.js tests are able to prevent their tests
+from being classified as "timed out," but the available methods of achieving
+this are suboptimal. A new mechanism for describing long-running tests would
+address much of the drawbacks and give authors more flexibility than they have
+today. Specifically: a new value for the `timeout` test metadata describing the
+number of subtests which are expected to run.
+
+```diff
+-<meta name="timeout" content="long">
++<meta name="timeout" content="400 subtests">
+```
+
+## Details
+
+Currently, testharness.js cancels tests and reports their result as "timed out"
+if they do not finish running within one of [two possible durations: "normal"
+and
+"long"](https://github.com/web-platform-tests/wpt/blob/31da90bff3d6fd5cd8169b924125b3253b438ec4/resources/testharness.js#L22-L25).
+This behavior has two negative consequences.
+
+Actual testing durations are not as consistent as these two options suggest,
+and careful authors are likely to prefer the "long" option to avoid false
+negatives. At scale, this tendency may have a non-trivial impact on overall
+test execution times (where non-conforming browsers require more time to fail).
+
+Some authors want to run many subtests organized under one logical test.
+Although the individual tests may be relatively quick, their number may push
+the test's execution time over the limit. Here, authors have to split their
+subtests across many files, creating subtests based on anecdotal evidence of
+what seems to be "fast enough" to satisfy the duration restriction in their
+environment. In the interest of limiting duplication, [this is sometimes
+implemented using methods which detract from
+readability](https://github.com/web-platform-tests/wpt/pull/11028).
+
+The harness's current restriction serves to detect and recover from cases where
+testing has unexpectedly halted. However, the harness currently ignores a
+signal that testing is indeed active: the reporting of subtest results. If the
+harness restarted its internal "timeout" countdown following every subtest
+result, it would be no less capable of detecting timeouts, and authors would no
+longer need to manage this expectation. It would also reduce waste (since a
+contextual timeout value of, e.g. "10 seconds after the last subtest" is likely
+shorter than the static timeout value of "60 seconds").
+
+Many subtests in WPT are generated procedurally-generated. Bugs in test code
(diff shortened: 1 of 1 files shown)


jugglinmike:
[When this feature was first suggested](https://github.com/web-platform-tests/wpt/issues/23326), @zcorpan wrote

> Why did we go with 10s / 60s originally? The proposal here seems more robust, and removes the need for splitting tests with many subtests.
>
> We could still have the `long` indicator, for cases where a single subtest takes a long time.
>
> To tackle the infinite subtests scenario, maybe there could be a fixed maximum timeout of a few minutes?

I can't speak to the first question, but as for the second: a generic solution would be better insofar as it requires less of the test author (they wouldn't have to maintain an explicit count of subtests). It would have a couple drawbacks, though:

- it would make the harness slower to recognize errors (since it would have to wait for the fixed maximum timeout rather than halting immediately after the first extraneous test was declared)
- by reporting the problem with a "TIMEOUT" status rather than an "ERROR" status, it would obscure the problem with the test's design (since the status of "TIMEOUT" is just as likely to be interpreted as a browser bug rather than a test bug)

jgraham:
I see the appeal of this design, but I do have some concerns.

We currently aren't streaming subtest results back to the harness; we just sen back a single JSON blob once the tests are done. Without such streaming the worst case for a configuration with timeout multiplier N is we wait N minutes (plus padding) for a test to timeout. For a debug android build that might already be 4 minutes which is itself already problematic; we have some tests that timeout on every run and it wastes a lot of machine time. But with this design it can be arbitrarily long; if we say that there are 1000 subtests and each subtest is allowed 10s and it happens that the first subtest hangs the browser (e.g. by spawning a print dalog) so that the harness is waiting 10000N seconds for the test to finish, that's clearly a problem.

I also see that this can in some cases make things better. In an example where we have 10 chained promise tests which all depend on some event that isn't fired in all implementations, and set `timeout=long` because each test can take >1s in the case where the test is working, we could timeout in 10N seconds rather than in 60N seconds.

I'm also generally worried by adding more test-author controlled parameters. "Subtest count" isn't a good metric because on large tests like encoding there can be tens or hundreds of thousands of subtests which all take a very short time, whereas tests where you have a lot of windows opening and navigations may take more than 10s per subtest. So in either case it's an implied proxy to time. The reason we don't allow authors to set per-test timeouts is simply that it's too hard to get it right across a large range of browsers and platforms; having "this is relatively quick" vs "this takes a long time" is a coarse model but one that doesn't require us to make so many adjustments as we run tests beyond the developer's local environment.

So in summary:
* I'm in favour of some way to make tests that have multiple short-running subtests timeout faster than the maximum timeout if they don't get subtest results. This should probably be opt-out (e.g. by having `subtest-long` as the meta value) and we should adjust existing tests that are affected (although I see this will be hard to do).
* I'm worried by the idea of allowing the test timeout to be arbitrarily long. I think if we want a longer maximum timeout we also need to change the runner to stream back subtest results (or have some kind of keepalive ping).
* I'm worried by proposals to make the maximum timeout directly author controlled rather than category-based.

jugglinmike:
Thanks for the feedback, @jgraham! I definitely see the value of streaming subtest results, but I'm having trouble interpreting your comments in light of that extension. I'm wondering which of your concerns are intended to motivate streaming (and would therefore be addressed by extending this proposal) and which would stand regardless.

> - I'm worried by the idea of allowing the test timeout to be arbitrarily long. I think if we want a longer maximum timeout we also need to change the runner to stream back subtest results (or have some kind of keepalive ping).

It seems like streaming would enable a more substantial change than a longer maximum timeout. Wouldn't it actually address your concern about a more dynamic heuristic?

> - I'm worried by proposals to make the maximum timeout directly author controlled rather than category-based.

"Number of subtests" is quantitative, but it's still an indirect control for timeout value. Just like "long" and "short," the runner is in a position to interpret the exact duration based on its performance characteristics. What if the author could write "10000 short subtests" or "10 long subtests"? It's more configuration, but it seems intuitive enough and maybe not too much to ask.

jgraham:
I think streaming, or at least keepalive pings, is a necessary precondition to anything that increases the maximum timeout value. I'm very worried about allowing longer timeouts without any mechanism to detect that we're still actively running tests and getting results.

> "Number of subtests" is quantitative, but it's still an indirect control for timeout value. Just like "long" and "short," the runner is in a position to interpret the exact duration based on its performance characteristics. What if the author could write "10000 short subtests" or "10 long subtests"? It's more configuration, but it seems intuitive enough and maybe not too much to ask.

It's still adding an extra axis of variability, and I'm concerned about people choosing values that makes us end up spending too long on tests. For example if you have 10 subtests, 9 that take 100ms and 1 that takes 10s, how do you set the subtest count? If you choose "10 long subtests" to allow the 10s test to complete then in the worst case you can end up waiting 10xlong subtest timeout, which might be 10 minutes, whereas before you might only have waited 1 minute in the worst case. Of course there are other things you could do like split the test into two, but the main motivation for this proposal was avoiding having to split tests with very large numbers of subtests in two, so if the tradeoff is we end up having to split other tests, it may not be a significant improvement.

jgraham:
I think this issue has unresolved technical objections, and doesn't currently look like it's going to progress. I'm going to close the issue for now, but if there is a way forward we can reopen the discussions.

PR #56, opened 2020-06-01
Add RFC for wait_for[_callback] methods

jgraham:


rfcs/wait_for.md
```diff
@@ -0,0 +1,56 @@
...
+
+`wait_for_callback(cond, callback, description, timeout=3000,
+interval=100)` and `wait_for(cond, description, timeout=3000,
+interval=100)`. In each case the function `cond` is called with no
+arguments, immediately and then every `interval` milliseonds until it
+returns a value that evaluates as `true`. If this does not happen in
+`timeout` milliseconds, an `AssertionError` is raised, using
+`description` as the error description.
+
+These functions must methods on `Test` to allow any error that's
```
annevk, line 34: must be?

rfcs/wait_for.md
```diff
@@ -0,0 +1,56 @@
...
+
+To encourage this pattern we introduce two new methods on `Test`:
+
+`wait_for_callback(cond, callback, description, timeout=3000,
+interval=100)` and `wait_for(cond, description, timeout=3000,
+interval=100)`. In each case the function `cond` is called with no
+arguments, immediately and then every `interval` milliseonds until it
+returns a value that evaluates as `true`. If this does not happen in
+`timeout` milliseconds, an `AssertionError` is raised, using
+`description` as the error description.
```
annevk, line 32: When is callback called?

annevk:
FWIW, this seems really good. Moving away from step_timeout and instead making everything use this will lead to a lot less idle time if I understand things correctly. Both when running things locally and when running things on CI as we'll no longer be bound by the lowest common denominator.

rfcs/wait_for.md
```diff
@@ -0,0 +1,60 @@
...
+order to allow tests to run faster.
+
+In some cases it's possible to detect directly whether the thing being
+waited on has occured or not. In this case the performance/robustness
+tradeoff can be improved by having a relatively long total timeout but
+polling for success to allow the test to proceed as soon as possible.
+
+To encourage this pattern we introduce two new methods on `Test`:
+
+`wait_for_callback(cond, callback, description, timeout=3000,
```
Ms2ger, line 26: Bikeshed: `wait_for_callback` sounds like it waits for a callback to be called. Not sure what would be better… `callback_wait_for`, maybe?

foolip: I had the same thought, but also don't know what to call it instead. Maybe something involving "step" to make it clear (?) that the callback will be wrapped in a step func automatically. `t.wait_for_step`?

jgraham: Yeah, I also don't know what's really better. `callback_wait_for` perhaps, although it's a bit weird. `step` seems a bit weird because both this and the promise variant are implied steps at least when running the condition function. More verbose names would be like `step_wait_cond` `step_wait_cond_and_callback`. I guess if we dropped `_cond` it wouldn't be too bad? `step_wait` and `step_wait_and_callback`?

jgraham: Or I suppose we could assume that all modern browser devs and other test contributors should be familiar with Rust and so use `step_wait` and `step_wait_and_then` for the promise/callback versions respectively? (that's a serious proposal despite the non-serious assumption ;)

annevk:
I've started playing with this locally and found that having wait_for_done for async_test wrappers would be useful as a way to avoid having to pass () => t.done() to wait_for_callback. Would this require a new RFC or is this a small enough amendment that I can add it and it can be accepted together with the remainder this week?

jgraham:
I'd consider that a small enough change to be done as an addon to this RFC. AFAIK the main blocker here at the moment is some sense of consensus on API names. @Ms2ger and @foolip previously had feedback; do any of the alternatives suggested appeal to you?

Ms2ger:
step_wait and step_wait_and_then are fine by me.

annevk:
Maybe step_wait_func and step_wait_func_done? This would be a slight change to my proposed wait_for_done in that a callback can still be passed, but I think is more consistent with step_func and step_func_done and probably worthwhile therefore.

jgraham:
I think I'm OK with `step_wait_func` and `step_wait_func_done` although you don't really wait on `func`. If you want `step_wait_func_done` to be useful I guess the callback has to be optional.

annevk:
Yeah, I was thinking it would be optional. Usually you probably don't need it, but sometimes it might be useful for some additional asserts.

rfcs/wait_for.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Risks
+
+* Adding API surface to testharness.js requires long-term support or
+  is quite difficult to remove or change if the feature is deemed to
+  be a failure.
+* Features like this and `step_timeout` could be useful in support
+  files, but using testharness.js features from support files is
+  tricky.
+* This RFC doesn't propose a lint to flag uses of `step_timeout` /
```
stephenmcgruer, line 56: Maybe I misunderstand this risk, but we already have a lint for `setTimeout`: https://github.com/web-platform-tests/wpt/blob/master/tools/lint/rules.py#L383 . We could change this lint to encourage `wait_for` instead, with `step_timeout` as a fallback if necessary.

A harder question is whether to or how to lint `step_timeout` of course, since that may be necessary more often (or do we strongly feel you should usually be using `wait_for`?)

rfcs/wait_for.md
```diff
@@ -0,0 +1,60 @@
...
+`wait_for_callback(cond, callback, description, timeout=3000,
+interval=100)` and `wait_for(cond, description, timeout=3000,
+interval=100)`. In each case the function `cond` is called with no
+arguments, immediately and then every `interval` milliseonds until it
+returns a value that evaluates as `true`. At this point `callback` is called.
+If this does not happen in `timeout` milliseconds, an `AssertionError` is
+raised, using `description` as the error description.
+
+`wait_for(cond, description, timeout=3000, interval=100)` works like
+`wait_for_callback` except a promise is resolved once the function
```
stephenmcgruer, line 35:
```suggestion
`wait_for_callback` except it returns a promise that is resolved once the function
```

jgraham:
> I'm not sure if this is bad practice, but my initial feeling was to always make the callback optional and just have the behavior depend on that,

I'm not a huge fan of this kind of overloading and it seems quite unidiomatic for testharness.js in general (e.g. we have a distinction between `test()` and `promise_test()`, although the latter could be "any `test()` where the function returns a thenable", and we could even have said that `async_test()` was a `test()` with no function passed in since in the early days it didn't take a function; fortunately we didn't since it turned out to be better to allow an initial function).

stephenmcgruer:
> I'm not a huge fan of this kind of overloading and it seems quite unidiomatic for testharness.js in general

Ok, seems reasonable to avoid it since its unidiomatic for testharness. I personally quite like it, but outside of some literature searching I think we're doing to personal opinion there :)

I can't come up with a better name for this, so I guess this LGTM. I wonder slightly if we should be helping folks who are listening for actual events (rather than a condition change), but maybe that should be an extension to EventWatcher instead. For example:

```
promise_test(async t => {
  let a = create_animation();
  // This is expected to happen in 1-2 frames. If the impl is broken, it will take
  // the entire test timeout time! 
  await a.ready;  
});
```

(Of course then you get into the fact that *technically* on a very slow machine 1-2 frames could take a long time, so we shouldn't be encouraging authors to define their own timeouts and should just depend on the whole test timeout, I guess...)

Anyway, long story short, LGTM but the PR needs updated with new names (and note my comment on the risks section). It would also be great if we could give a handful of examples in the PR of where people are doing things today that would be improved by this.

annevk:
For examples you could link https://github.com/web-platform-tests/wpt/pull/24038 I suppose. (Though that also needs updating for the new names.)

jgraham:
I think we have enough approvals here to go ahead and merge.

PR #58, opened 2020-06-23
RFC 58: Add a new test property to control displaying test state updates

sefeng211:
test changes: https://github.com/web-platform-tests/wpt/pull/24170

### DIFF
--- a/rfcs/hide_test_state.md
+++ b/rfcs/hide_test_state.md
@@ -1,0 +1,22 @@
+# RFC 58: Add a new test property to control displaying test state updates
+
+## Summary
+
+Add a new test property called `hide_test_state` to control displaying
+test state updates while tests are running.
+
+This property will default to `false`, and can be set to `true` via the `setup()` function.
+
+## Details
+
+In testharness there are a few usage of `show_status()` function,
+which outputs the test state to the documents during test
+runs. Some tests are sensitive to the content of the documents,
+so the test state output may interfere the test results.
+
+We are adding a new setup property to prevent outputting the
+test state.
+
+## Risks
+
+No risks as this is an opt-in property.


rfcs/disable_notify_test_state.md
```diff
@@ -0,0 +1,20 @@
+# RFC ??: Add a new test property to not displaying test state
```
stephenmcgruer, line 1:
```suggestion
# RFC 58: Add a new test property to control displaying test states
```

rfcs/disable_notify_test_state.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Summary
+
+Add a new test property called `notify_test_state` to not displaying
```
stephenmcgruer, line 5:
```suggestion
Add a new test property called `notify_test_state` to control displaying
```

rfcs/disable_notify_test_state.md
```diff
@@ -0,0 +1,20 @@
...
+test state when tests are running.
+
```
stephenmcgruer, line 7:
```suggestion
test state when tests are running.

This property will default to `true`, and can be set to `false` via the `setup()` function.
```

rfcs/disable_notify_test_state.md
```diff
@@ -0,0 +1,20 @@
+# RFC ??: Add a new test property to not displaying test state
+
+## Summary
+
+Add a new test property called `notify_test_state` to not displaying
+test state when tests are running.
```
stephenmcgruer, line 6:
```suggestion
test state updates while tests are running.
```

rfcs/disable_notify_test_state.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Details
```
stephenmcgruer, line 8: How does this interact with the `test_done_callbacks`? I believe those too cause `WindowTestEnvironment` to update the DOM, right?

sefeng211: I don't think this impacts the `test_done_callbacks`, this PR just prevents `notify_test_state` to be invoked. We invoke the `test_done_callbacks` in `notify_result` function which is not impacted by this PR.

jgraham:
It's kind of unclear if blocking `notify_test_state` is actually the right approach, since in theory other things could be using that (e.g. if we changed the harness to stream events back and used that to build an interactive view of the test state). It might make more sense to just make the Output handler not register a callback in case this property is set.

rfcs/disable_notify_test_state.md
stephenmcgruer: Yes, what I'm asking is if they could affect the tests! E.g. if a test has 10 subtests, aren't the `test_done_callbacks` invoked for each individual subtest finishing, and then each of those 10 updates causes DOM changes?

Put another way, I suspect that blocking just `notify_test_state` may not be sufficient.

sefeng211:
@stephenmcgruer @jgraham Yeah....good point! 

How about we just not register the `show_status()` function to `test_state_callbacks` and `test_done_callbacks`? 

So this test property could be called `register_show_status_callback`?

jgraham:
I think we should name the property for the desired effect rather than the implementation strategy. Maybe `notify_test_state` seemed OK because we would stop doing in-content notifications, but I see that it's confusing if you think of it as blocking all kinds of notifications. So we could maybe have `output_test_state: false` or `allow_harness_dom_updates: false` or something.

sefeng211:
@jgraham sounds good, what do you think about the implementation strategy, Stephen mentioned a good point that we don't want to output anything between tests.

jgraham:
I don't `show_results` in the output function is only called at most once when the tests are complete. We don't do interactive update of the table of results. So I think as long as `show_status` isn't called it's probably fine.

sefeng211:
I updated the WPT patch to change the implementation from checking the flag in `notify_test_state` to not calling the `show_status` function in the callback. Detail: https://github.com/web-platform-tests/wpt/pull/24170/commits/255548294450d02c4465e62c8a434dc12faf55a5, @stephenmcgruer Please let me know what you think :)

sefeng211:
I left the flag name as `hide_test_state` because we want the default value to be false.

rfcs/hide_test_state.md
```diff
@@ -0,0 +1,22 @@
+# RFC 58: Add a new test property to control displaying test states
+
+## Summary
+
+Add a new test property called `hide_test_state` to control displaying
+test states update while tests are running.
+
+This property will default to `true`, and can be set to `false` via the `setup()` function.
```
Ms2ger, line 8:
```suggestion
This property will default to `false`, and can be set to `true` via the `setup()` function.
```

sefeng211:
Updated!

sefeng211:
@Ms2ger Does it look good to you? (pinging in case it falls off your radar)

stephenmcgruer:
Checked with jgraham that they're happy too, so that's the 1-week LGTM-no-objections time passed. Merging.

PR #63, opened 2020-09-29
Add support for running testdriver commands outside the test context

jgraham:
https://github.com/web-platform-tests/wpt/pull/25833 contains the initial implementation

### DIFF
--- a/rfcs/testdriver_cross_origin.md
+++ b/rfcs/testdriver_cross_origin.md
@@ -27,7 +27,7 @@ The specific proposed changes are:
   context.
 
 * Also provide a `message_test` function as a wrapper around
-  postMessge to the test context.
+  postMessage to the test context.
 
 * Add an optional `context` argument to test_driver commands that
   don't depend on a provided element, to specify the context in which
--- a/rfcs/testdriver_cross_origin.md
+++ b/rfcs/testdriver_cross_origin.md
@@ -1,0 +1,70 @@
+# RFC 63
+
+## Summary
+Support running testdriver commands in browsing contexts other than
+the test browsing context, including cross origin contexts where it's
+possible to get a handle to the browsing context containing the test
+(i.e. excluding the rel=noopener case).
+
+## Details
+
+For contexts in which it's possible to get a handle to the context
+containing testharness.js ("the test context"), we can support
+testdriver interactions by using postMessage to the top-level window
+with a message containing the command to run and the context to run it
+in. That approach works with the WebDriver-based implementation in
+wptrunner where WebDriver can only address a single frame at a time,
+and a WebDriver-driven event loop in the test context is used to
+invoke testdriver actions and also to report the test results.
+
+The specific proposed changes are:
+
+* Allow testdriver.js to be loaded in contexts other than the test
+  context.
+
+* In such contexts require use of testdriver to first call the
+  `test_driver.set_test_context()` function with a handle to the test
+  context.
+
+* Also provide a `message_test` function as a wrapper around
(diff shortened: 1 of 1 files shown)


rfcs/testdriver_cross_origin.md
```diff
@@ -0,0 +1,70 @@
...
+
+* Allow testdriver.js to be loaded in contexts other than the test
+  context.
+
+* In such contexts require use of testdriver to first call the
+  `test_driver.set_test_context()` function with a handle to the test
+  context.
+
+* Also provide a `message_test` function as a wrapper around
+  postMessge to the test context.
```
stephenmcgruer, line 30:
```suggestion
  postMessage to the test context.
```

rfcs/testdriver_cross_origin.md
```diff
@@ -0,0 +1,70 @@
...
+wptrunner where WebDriver can only address a single frame at a time,
+and a WebDriver-driven event loop in the test context is used to
+invoke testdriver actions and also to report the test results.
+
+The specific proposed changes are:
+
+* Allow testdriver.js to be loaded in contexts other than the test
+  context.
+
+* In such contexts require use of testdriver to first call the
```
stephenmcgruer, line 25: I wonder if we should not make this (and `message_test`) part of the public contract of testdriver.js. Instead, we would specify that `test_driver_internal` should work in non-test browsing contexts and leave it to the vendors how they would support that?

rfcs/testdriver_cross_origin.md
```diff
@@ -0,0 +1,70 @@
...
+
+The implementation details for wptrunner are as follows:
+
+* testdriver events are given a command id and a context argument. The
+  command id is used to match the response from webdriver to the
+  specific request. The context id is used to identify which frame or
+  window should be used for the request.
+
+* Ideally the context id would be the wptrunner window/frame handle
+  for the window or frame. But this is poorly supported so instead we
```
stephenmcgruer, line 44: Could you link or otherwise add a footnote specifying in what way this is poorly supported?

stephenmcgruer (approved):
Overall I think this LGTM.

Putting on my downstream consumer hat, it's likely that Chromium's downstream test runner will either not support these tests or that we will do it in a significantly different way, so I'm in favor of minimizing changes to the `testdriver.js` API itself (vs putting stuff in `testdriver-extras.js`). This RFC does seem to mostly do that.

Putting on my WPT hat - sure, this isn't perfect, but in the imperfect world of WebDriver it seems pretty good, and its an improvement many test authors have wanted. We may one-day have a better WebDriver and might want to walk back this model somewhat, but I don't see that as particularly insurmountable come that day.

rfcs/testdriver_cross_origin.md
stephenmcgruer: You're very correct, of course. I started with "I want to eliminate `message_test` from the public API" and then somehow let my brain increase the scope to `set_test_context` despite the fact that part has to be called by test authors -_-.

jgraham: I think `message_test` is useful because even with a perfect system you need some way to tell the test window that the testdriver action completed.

jgraham:
This has been a week and doesn't have any remaining negative feedback, so unless there's further comments expressing concerns I'll merge it tomorrow.

rfcs/testdriver_cross_origin.md
stephenmcgruer: Ack, sgtm.

PR #65, opened 2020-10-28
RFC 65: Switching to "Py3-first" for WPT tests on CI

ziransun:


rfcs/Py3-first.md
```diff
@@ -0,0 +1,53 @@
+# RFC 59: Switching to "Py3-first" for WPT tests on CI
```
stephenmcgruer, line 1:
```suggestion
# RFC 65: Switching to "Py3-first" for WPT tests on CI
```

jgraham (commented):
Is there a proposal for the date when these changes will occur?

Ms2ger (approved):
:+1:

stephenmcgruer:
> Is there a proposal for the date when these changes will occur?

Not yet. I would suggest that we make it part of the RFC to say we aim to switchover shortly after merge, but will announce the switchover in [PLACES] with X days notice.

Before we get there, we need to determine what are the list of blocking issues versus which aren't blocking. The blocking issues I know of are:

* [ ] The [threading issue](https://github.com/web-platform-tests/wpt/issues/25678) - this is potentially non-trivial.
* [ ] The flaky-failure when trying to [install Py3 on the Windows test-suite bots](https://github.com/web-platform-tests/wpt/pull/24952).
* [ ] The other 'switch to Py3' pull requests should probably all be in a reasonably reviewed state before we land the RFC (basically so there are no surprises when it comes to switchover day).

Are there any other blocking issues folks are aware of?

rfcs/Py3-first.md
```diff
@@ -0,0 +1,53 @@
...
+This is to propose switching test suite runs over to Python 3 on the WPT CI (Taskcluster and Azure). At the same time unit tests and infrastructure tests will still be running in Python 2 and Python 3. This is referred to as the “Py3-first" stage in the [Dropping Python 2 support plan](https://github.com/web-platform-tests/rfcs/issues/62).
+
+The minimum python 3 version supported is 3.6 with main focus on 3.8+. For the foreseeable future it is very likely to continue to run unit tests and infrastructure tests on both 3.6 and 3.8+ versions. The WPT test suite itself would run on 3.8+ in the CI.
+
+## Changes on upstream CI
+
+**1. Switch the test suite runs to use Python 3**
+*    [Switch wpt run on Taskcluster to python 3 only](https://github.com/web-platform-tests/wpt/pull/26252) (specifically, Python 3.8)
+
+*     Use Python 3.8.x in Azure Pipelines for [Safari Preview](https://github.com/web-platform-tests/wpt/pull/25044) and [Edge Stable/Dev runs](https://github.com/web-platform-tests/wpt/pull/24952)
```
stephenmcgruer, line 14:
```suggestion
**1. Switch the test suite runs to use Python 3**
* [Switch wpt run on Taskcluster to Python 3 only](https://github.com/web-platform-tests/wpt/pull/26252) (specifically, Python 3.8)

* Use Python 3.8.x in Azure Pipelines for [Safari](https://github.com/web-platform-tests/wpt/pull/25044) and [Edge](https://github.com/web-platform-tests/wpt/pull/24952) runs.
```

rfcs/Py3-first.md
```diff
@@ -0,0 +1,53 @@
...
+* Newly introduced `.py` test files or codes. It is noted that before we move to the stage of "Py3-only", there is a risk that newly introduced `.py` codes might not be Python 3 compatible. We have fixed issues found in the tracking sheets but it is very likely that this kind of issues will appear again during the transition period.
+
+Looking into every different result of the WPT test suite between Python 2 and Python 3, we believe that we have each known Python 3 issue either fixed or at least explained.
+
+## Rollback and success criteria
+
+Admins (smcgruer & Hexcles) will be on call on #testing at irc.w3.org on the flag day. In case of unexpected massive failures (e.g. a large number of tests fail to run or produce invalid results), they will revert the following PRs (overriding any checks if needed):
+* [Switch wpt run on CI to python 3 only](https://github.com/web-platform-tests/wpt/pull/26252)
+* [Use Python 3.8.x in Azure Pipelines for Safari Preview](https://github.com/web-platform-tests/wpt/pull/25044)
+* [Use Python 3.8.x in Azure Pipelines for Edge Stable/Dev runs](https://github.com/web-platform-tests/wpt/pull/24952)
```
stephenmcgruer, line 51:
```suggestion
* [[Taskcluster] Switch wpt run on CI to Python 3 only](https://github.com/web-platform-tests/wpt/pull/26252)
* [[Azure Pipelines] Use Python 3.8.x for Safari test-suite runs](https://github.com/web-platform-tests/wpt/pull/25044)
* [[Azure Pipelines] Use Python 3.8.x for Edge test-suite runs](https://github.com/web-platform-tests/wpt/pull/24952)
```

Hexcles:
>  The threading issue - this is our riskiest blocker.

With a tight-ish time frame in front us, I think we should take https://github.com/web-platform-tests/wpt/pull/26111 to work around this for now. We talked about experimenting with multiprocessing; I fully agree this is something we should try, but we probably don't want to block the migration on the experiment.

stephenmcgruer:
> With a tight-ish time frame in front us, I think we should take web-platform-tests/wpt#26111 to work around this for now. We talked about experimenting with multiprocessing; I fully agree this is something we should try, but we probably don't want to block the migration on the experiment.

Sorry, last I checked I thought the fix did not actually address the encoding problem. Did the locking added in https://github.com/web-platform-tests/wpt/pull/26111/commits/b7fa47294b8c4646b77a53287384b90554974fc2 definitely resolve the encoding import errors?

Hexcles:
Sorry, should've made that clear in the PR. The answer is **yes**, but see https://github.com/web-platform-tests/wpt/pull/26111#pullrequestreview-508523896 for caveats...

stephenmcgruer:
> Sorry, should've made that clear in the PR. The answer is yes, but see web-platform-tests/wpt#26111 (review) for caveats...

I see. That changes my opinion on this; if we have a valid (albeit scary) workaround **and** are committed to finding a proper path forward that will remove the workaround (which we are; I am happy to own that myself), I don't think the threading issue should block the RFC then.

@jgraham , WDYT about the threading issue given this?

jgraham:
Ideally I'd like us to try the absolute import thing first. If we can't make that work we could fall back to the existing patch, but I think we should be very wary of taking something that pokes at so many internals; if we drop the ball on fixing the underlying issue it seems almost certain we'll see breakage at some point.

stephenmcgruer:
> Ideally I'd like us to try the absolute import thing first. If we can't make that work we could fall back to the existing patch, but I think we should be very wary of taking something that pokes at so many internals; if we drop the ball on fixing the underlying issue it seems almost certain we'll see breakage at some point.

Ack, I can own experimenting with that over the next few (business) days.

stephenmcgruer:
> Ack, I can own experimenting with that over the next few (business) days.

To keep folks up-to-date on this, I am playing with it in https://github.com/web-platform-tests/wpt/pull/26328 . My next task is to diff a set of trigger runs that I did for Chrome Dev.

stephenmcgruer:
An update:

* We believe we have cleared the path for the Windows 3.8 installation issue.
* The Python file handler threading issue has a proposed PR by me out for review now, but it's a fairly invasive change and will likely need an RFC of its own anyway.

rfcs/Py3-first.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Risks
+
+To assess possible risks and provide evidence for  a move to Python 3 being safe, we have produced [a tracking spreadsheet](https://docs.google.com/spreadsheets/d/1cJcSW8PHe3A4m0dtBfjV4nrYLg67LzIznFbZIknDURc/edit#gid=1054524593). It compares the test results of wpt runs on Python2 and Python 3 for all the supported configurations including:
+* Linux+Chrome Dev
+* Linux+Firefox Nightly
+* Windows + Edge Dev
+* macOS + Safari Technology Preview
+
+Most different test results are caused by the fact that the test itself is flaky. For those flaky tests, links to [flakes](https://wpt.fyi/insights) or results from [recent runs](https://wpt.fyi/runs) are given.
```
stephenmcgruer, line 35: @ziransun can we update this with recent results from e.g. https://github.com/web-platform-tests/wpt/pull/26252#issuecomment-720607692 ?

That should remove the HTTP/2 support line since that has been added.

rfcs/Py3-first.md
```diff
@@ -0,0 +1,53 @@
...
+* Different test names in some `webaudio` tests. We noticed  `cors-check.https.html`produced different test names for both Chrome and Firefox, also `no-cors.https.html` for Firefox. This is due to the way the tests were written, which in some way is a bit strange while creating test names.
+* Threading issue. [Issue #25678](https://github.com/web-platform-tests/wpt/issues/25678) about earlier relative imports shadow later ones with the same name have been reported and discussed. For the moment there is no direct fix for it. However, for tests in infrastructure that failed due to this issue, [a workaround]( https://github.com/web-platform-tests/wpt/pull/26111/) that makes local absolute imports not incorrectly cache between handlers has been provided.
```
stephenmcgruer, line 41:
```suggestion
* Python file handler local-imports issue, aka [issue #25678](https://github.com/web-platform-tests/wpt/issues/25678. An [RFC has been filed](https://github.com/web-platform-tests/rfcs/pull/68) along with [a fix](https://github.com/web-platform-tests/wpt/pull/26328) for this.
```

stephenmcgruer (approved):
This depends on https://github.com/web-platform-tests/rfcs/pull/68 (and the related fix), but once that's in this LGTM.

rfcs/Py3-first.md
ziransun: @stephenmcgruer: Updated the track sheet and removed HTTP/2 statement. Thank you!

jgraham (commented):
So I generally agree with everything here, but I'd like a Python 3 meta RFC formalizing the timeline for all the steps, rather than just having an RFC for one part of the process and the overall framework in an issue. I can write that.

stephenmcgruer:
> So I generally agree with everything here, but I'd like a Python 3 meta RFC formalizing the timeline for all the steps, rather than just having an RFC for one part of the process and the overall framework in an issue. I can write that.

I mean, no objection since you're willing to write it, but isn't that two lines?

- **November 2020**: Py3-first
- **Feburary 2021**: Py3-only

jgraham:
I've written #69 to detail the overall transition plan.

The one thing I realised while writing that that isn't covered here is the behaviour of `wpt --py3`; I think we need a `--py2` option during the transition. I can make a PR for that.

jgraham:
https://github.com/web-platform-tests/wpt/pull/26507

gsnedders (approved):

PR #67, opened 2020-11-04
RFC #67: Add WebSockets over HTTP/2 support to wptserve

arenevier:


arenevier:
@jgraham @ricea

rfcs/websockets_http2.md
```diff
@@ -0,0 +1,24 @@
...
+- Detect Extended CONNECT Method and implement the WebSocket opening handshake.
+- Wrap H2Response and H2Request into a Request-like object that pywebsocket can operate on.
+
+## Risks
+
+Most WebSocket tests have a wss variant. We will need to add a new variant in order to use the http2 port when opening the WebSockets.
+
+No other risks are expected except those introduced by any feature e.g. complexity.
+
+This change does not introduce any new dependency.
```
Hexcles, line 24: But it requires updating some existing dependencies, right? If so, we should list them here just in case.

rfcs/websockets_http2.md
```diff
@@ -0,0 +1,24 @@
...
+## Details
+
+Allow tests to open a WebSocket over an existing http2 connection.
+
+wptserver needs to implement the [Bootstrapping WebSocket with HTTP/2 protocol](https://tools.ietf.org/html/rfc8441).
+
+It means that it needs to be modified to:
+- Set ENABLE_CONNECT_PROTOCOL settings to 1
+- Detect Extended CONNECT Method and implement the WebSocket opening handshake.
+- Wrap H2Response and H2Request into a Request-like object that pywebsocket can operate on.
```
Hexcles, line 16: Should point out which `Request` this is (the standard `http.client` library? our own in `wptserve`? or something else?).

rfcs/websockets_http2.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Risks
+
+Most WebSocket tests have a wss variant. We will need to add a new variant in order to use the http2 port when opening the WebSockets.
```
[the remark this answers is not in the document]
arenevier: > Dumb question: does WS over H2 also support both plaintext and TLS (and by the way what are their schemes)? If so, then we need two more variants.

Browsers don't support insecure http2, and wpt server also do not support it. I will mention that in this rfc.

ricea:
Non-owner lgtm

rfcs/websockets_http2.md
```diff
@@ -0,0 +1,24 @@
...
+wptserver needs to implement the [Bootstrapping WebSocket with HTTP/2 protocol](https://tools.ietf.org/html/rfc8441).
+
+It means that it needs to be modified to:
+- Set ENABLE_CONNECT_PROTOCOL settings to 1
+- Detect Extended CONNECT Method and implement the WebSocket opening handshake.
+- pywebsocket operates on Apache mod_python Request objects. So we need to wrap H2Response and H2Request into a Request-like object.
+
+## Risks
+
+WebSocket tests who source **websockets/constants.js** have a wss variant. We will need to add a new variant named **http2** in order to use the http2 port when opening the WebSockets. WTP server does not support insecure HTTP/2 so we don't need to expose an insecure HTTP/2 WebSocket variant.
```
zcorpan, line 20: There are also websocket tests at `websockets/Secure-*` using wss, but not using `websockets/constants.js`. Instead they use `websockets/websocket.sub.js`, which might need a similar change.

(Why this disparity between websocket tests? Some tests came from Opera, other tests came from Microsoft, and nobody has rewritten them to be consistent...)

arenevier: What would be the ideal way to fix it? There is a lot of duplication between Secure-* and non Secure- test. Maybe they can be rewritten to use variants. That would reduce the number of files, and make it easier to test over h2?

Also, should we have a separate discussion and/or rfc for those tests? I'm ready to help with the rewriting of those tests. But it can be decoupled from the task of enabling ws over h2 in wptserve.

zcorpan: > I'm ready to help with the rewriting of those tests.

Awesome! Thank you!

Yeah, I think: Merge `constants.js` and `websocket.sub.js` into one. Make the test files directly under `websocket/` be more like the tests in directories (e.g. using `variant`s). And move them into the directories. Maybe all tests could use some massage in the process, but at a high level.

There might be some duplication in the websocket handlers, too.

I'm not sure if an RFC is required for this kind of cleanup, or if an issue in wpt/ is sufficient.

jgraham: No RFC required for test cleanup.

rfcs/websockets_http2.md
```diff
@@ -0,0 +1,24 @@
...
+
+This change needs to update hyper libraries to a more recent version.
```
jgraham, line 22: Are there notable changes in the update that might either affect existing tests or require installing new third-party libraries (especially any that require non-Python code)

arenevier: nothing of that sort.

After double checking hyper changelog though, I notice that in the version we need, support for python 3.3 and python 3.7 is dropped.

Does wpt support those versions of python? If so, the alternative is to backport the 3 patches we need.

jgraham: I believe we are targeting the range 3.6-3.8 initially. I wonder why 3.7 support was dropped; it seems to still be an officially supported version of Python.

arenevier: Sorry my bad. I had misread hyper changelog. It actually says:
"Drop support for Python 3.3 and support 3.7"


Since wpt support 3.6-3.8, this is fine.

rfcs/websockets_http2.md
Hexcles:
```suggestion
WebSocket tests who source **websockets/constants.js** have a wss variant. We will need to add a new variant named **http2** in order to use the http2 port when opening the WebSockets. WPT server does not support insecure HTTP/2 so we don't need to expose an insecure HTTP/2 WebSocket variant.
```

stephenmcgruer (approved):
LGTM. This seems fairly self-contained; I can't see any reason this would give downstream consumers problems.

stephenmcgruer:
As a note: we are 2 days from the initial 7-day period, and there is no substantial disagreement. If anyone wants more time, they can request an additional 1 week extension.

jgraham (approved):

Ms2ger (approved):
Sounds fine.

stephenmcgruer:
Thanks for the RFC @arenevier ! We've reached the 7-day mark with no objections, so merged.

PR #69, opened 2020-11-12
RFC 69: Switch web-platform-tests to Python 3

jgraham:


Ms2ger:
I'm wondering if the timeline isn't too tight. I'd assume we lost at least the last two weeks of the year to various people's holidays, so that leaves about 5 weeks at best from today to get everything finished on our side and on the vendors's side. Moving the timeline by two weeks or a month might mean people don't have to rush as much. I don't know how urgently people need to move this along, though, so happy to be overruled on this if it seems feasible to the people who are going to do the work.

gsnedders (approved):

jgraham:
I want to switch Gecko over asap after upstream, but not before. I also don't want Gecko on 2.7 next year. So from my point of view doing the GitHub change on the last week of November and then doing the Gecko change the first week of December seems fine, although there could be breakage that doesn't happen upstream (I think I checked most things but not everything). So from my point of view this timetable is tight but doable. If we were going to slip dates I'd like it to be extending the vendor "deadline" to mid Jan or 1st Feb. and moving the end of Py 2 support to match. Obviously the longer the overall transition period is the more chance we get get cross-version breakage. So if others have concerns we can make that kind of change, but I think this timetable can work for gecko.

stephenmcgruer:
Similarly I am looking to switch Chromium's use of WPT to Python 3 in early December.

rfcs/py_3.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Summary
+
+This RFC sets out the overall plan and timetable for the Python
+2->Python 3 transition. It is anticipated that further details on the
+transition will be set out in other RFCs as appropriate.
+
+## Details
+
+ 1. **November / December 2020** XXX add actual date - Python 3 becomes default on GitHub CI
```
stephenmcgruer, line 11: Should the 'XXX add actual date' be removed?

stephenmcgruer (approved):

rfcs/py_3.md
stephenmcgruer: At a guess, first attempt at this will likely happen the week of the 23rd. The handlers change RFC can be landed on the 17th, but that's the week of BlinkOn for us so I don't want to do the Python 3 switchover at the same time (and I'd like to give the handlers time to bake as well).

jgraham: I updated the document to list that date. I think if we miss the week of the 23rd for whatever reason we should consider pushing the other dates back too.

jgraham:
This has now been >1 week without additional objects, so I'm going to merge it.

PR #70, opened 2020-11-12
RFC 70: Add Python3 WebSockets dependency for WebDriver Bidi tests

k7z45:
RFC for https://github.com/web-platform-tests/wpt/issues/26015.
@stephenmcgruer @Hexcles @foolip @gsnedders @jgraham @zcorpan

### DIFF
--- a/rfcs/websockets_py3.md
+++ b/rfcs/websockets_py3.md
@@ -10,7 +10,7 @@ Add Py3-only [WebSockets client](https://github.com/aaugustin/websockets) that u
 
 [WebDriver Bidi](https://w3c.github.io/webdriver-bidi/) builds and extends on [WebDriver](https://w3c.github.io/webdriver/).
 WebDriver Bidi uses WebSockets for bidirectional communication between remote end and local end.
-Existing pywebsocket3 library only serves as a websocket server. Therefore, a new WebSockets client dependency is needed in order to connect to WebDriver for Bidi communication.  
+Existing pywebsocket3 library only serves as a WebSockets server. Therefore, a new WebSockets client dependency is needed in order to connect to WebDriver for Bidi communication.  
 
 ### Proposal
 
--- a/rfcs/websockets_py3.md
+++ b/rfcs/websockets_py3.md
@@ -1,4 +1,4 @@
-# RFC : Add Python3 WebSockets dependency for WebDriver Bidi tests
+# RFC 70: Add Python3 WebSockets dependency for WebDriver Bidi tests
 
 ## Summary
 
--- a/rfcs/websockets_py3.md
+++ b/rfcs/websockets_py3.md
@@ -9,14 +9,14 @@ Add Py3-only [WebSockets client](https://github.com/aaugustin/websockets) that u
 ### Background
 
 [WebDriver Bidi](https://w3c.github.io/webdriver-bidi/) builds and extends on [WebDriver](https://w3c.github.io/webdriver/).
-WebDriver Bidi uses WebSockets for bidirectional communication between remote end and local end.
-Existing pywebsocket3 library only serves as a WebSockets server. Therefore, a new WebSockets client dependency is needed in order to connect to WebDriver for Bidi communication.  
+WebDriver Bidi uses WebSockets for bidirectional communication between the remote end (acting as the server) and the local end (WPT, acting as a client).
+The existing pywebsocket3 library that is vendored into WPT only serves as a WebSockets server. Therefore, a new WebSockets client library dependency is needed in order to connect to WebDriver for Bidi communication.  
 
 ### Proposal
 
-Add Py3-only WebSockets library linked in Summary.
+Add the Py3-only WebSockets client library linked in the Summary above.
 
-Add [pytest-asyncio](https://github.com/pytest-dev/pytest-asyncio) for convenient test marker.
+Additionally, add a dependency on [pytest-asyncio](https://github.com/pytest-dev/pytest-asyncio) for its convenient test marker.
 Note pytest-asyncio depends on later versions of pytest and pluggy so they also need to upgrade to compatible versions.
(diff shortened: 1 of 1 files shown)


rfcs/websockets_py3.md
```diff
@@ -0,0 +1,45 @@
...
+about missing _pytest._version module.
+I commented out the import and hardcoded the version to 6.1.1.
+1. I installed the WebSockets and pytest-asyncio dependencies locally instead of including them in tools/third_party library.
+
+## Risks
+
+The current plan for python migration(I believe) is Py3-first by the end of the year and Py3-only Feb 2021.
+The patch is Py3-only and there is a gap of 3 months from now and then.
+Do we want to wait until then to add the dependencies?
+Or do we want to be able to exclude these dependencies for Py2?
```
stephenmcgruer, line 45: The February 2021 date is correct. Reference: the Py3 overview RFC, https://github.com/web-platform-tests/rfcs/pull/69/files

So it really comes down to how quickly we want/need webdriver bidi testing support, which I think is really something for webdriver folks to answer :). Internally to WPT, we likely won't be developing a webdriver bidi executor until late 2021 I expect (once bidi has standardized better).

Who are the correct decision makers for whether webdriver bidi testing support is needed before Feb 2021? @jgraham ?

jgraham: I think we will want to land this support before feburary, but I also think that we can make it opt-in (c.f. the quic support), or only enabled for the correct Python version. In practice it's expected that vendors and GH infra will be running Py3 by EoY so the gap is more like a few weeks.

jgraham (approved):
In principle this plan looks good; I haven't yet read the implementation PR.

rfcs/websockets_py3.md
k7z45: I talked to @Hexcles offline about "quick support" and "enabling for Py3".
I believe we can conditionally import "async" using six.PY3 flag and skip WebDriver tests that use async.
However, It's tricky to handle the third party dependency upgrade for pytest as the upgraded version is Py3 only.
I might just wait for "a few weeks" for the gap to fill.

foolip:
@sadym-chromium and I added some tests in https://github.com/sadym-chromium/WebDriverBiDiServerExperiments/tree/5be157aaaaada7ca8bf44ac2ffcb1c275b01da8e/bidiClient using Python 3, and happily we picked the same library being proposed here, https://github.com/aaugustin/websockets.

jgraham:
I think on the basis that we're going to allow Py3-only code in a couple of weeks we can merge this but defer the implementation until then. I'll leave a couple of days in case there are objections to that plan.

rfcs/websockets_py3.md
```diff
@@ -0,0 +1,45 @@
+# RFC 70: Add Python3 WebSockets dependency for WebDriver Bidi tests
+
+## Summary
+
+Add Py3-only [WebSockets client](https://github.com/aaugustin/websockets) that uses asyncio for WebDriver Bidi tests.
```
stephenmcgruer, line 5:
```suggestion
Add a Py3-only [WebSockets client](https://github.com/aaugustin/websockets) library that uses `asyncio`, to enable WebDriver Bidi tests. 
```

rfcs/websockets_py3.md
```diff
@@ -0,0 +1,45 @@
...
+## Summary
+
+Add Py3-only [WebSockets client](https://github.com/aaugustin/websockets) that uses asyncio for WebDriver Bidi tests. 
+
+## Details
+
+### Background
+
+[WebDriver Bidi](https://w3c.github.io/webdriver-bidi/) builds and extends on [WebDriver](https://w3c.github.io/webdriver/).
+WebDriver Bidi uses WebSockets for bidirectional communication between remote end and local end.
```
stephenmcgruer, line 12:
```suggestion
WebDriver Bidi uses WebSockets for bidirectional communication between the remote end (acting as the server) and the local end (WPT, acting as a client).
```

rfcs/websockets_py3.md
```diff
@@ -0,0 +1,45 @@
...
+Existing pywebsocket3 library only serves as a WebSockets server. Therefore, a new WebSockets client dependency is needed in order to connect to WebDriver for Bidi communication.
```
stephenmcgruer, line 13:
```suggestion
The existing pywebsocket3 library that is vendored into WPT only serves as a WebSockets server. Therefore, a new WebSockets client library dependency is needed in order to connect to WebDriver for Bidi communication.  
```

rfcs/websockets_py3.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Background
+
+[WebDriver Bidi](https://w3c.github.io/webdriver-bidi/) builds and extends on [WebDriver](https://w3c.github.io/webdriver/).
+WebDriver Bidi uses WebSockets for bidirectional communication between remote end and local end.
+Existing pywebsocket3 library only serves as a WebSockets server. Therefore, a new WebSockets client dependency is needed in order to connect to WebDriver for Bidi communication.  
+
+### Proposal
+
+Add Py3-only WebSockets library linked in Summary.
```
stephenmcgruer, line 17:
```suggestion
Add the Py3-only WebSockets client library linked in the Summary above.
```

rfcs/websockets_py3.md
```diff
@@ -0,0 +1,45 @@
...
+
+Add [pytest-asyncio](https://github.com/pytest-dev/pytest-asyncio) for convenient test marker.
```
stephenmcgruer, line 19:
```suggestion
Additionally, add a dependency on [pytest-asyncio](https://github.com/pytest-dev/pytest-asyncio) for its convenient test marker.
```

stephenmcgruer (approved):

PR #72, opened 2020-11-17
RFC 72: Address space overrides.

letitz:
Addresses issue web-platform-tests/wpt#26166.

cc @mikewest @camillelamy @hemeryar


rfcs/address_space_overrides.md
```diff
@@ -0,0 +1,111 @@
...
+
+For the purposes of testing, declare that the following subnets are not `local`
+but instead:
+
+* 127.1.0.0/16: `private`
+* 127.2.0.0/16: `public`
+
+### Browser changes
+
+A new command-line flag (or a collection thereof) is added to each browser (it
```
jgraham, line 30: It would be useful to file bugs on various browsers to ensure that this plan is actually plausible. Unlike most RFCs, the majority of the implementation complexity here is outside wpt itself.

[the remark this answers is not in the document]
letitz: That's fair. It can be left to the browser implementers to offer some configuration surface for this.

letitz: Also fair. I'm curious if you have any feedback on this approach vs the "Use individual IP addresses" alternative?

rfcs/address_space_overrides.md
```diff
@@ -0,0 +1,111 @@
...
+which might come in handy [in the
+future](https://github.com/WICG/cors-rfc1918/pull/1#issuecomment-721110250).
+
+### Use real IP addresses
+
+If we need to test user agent behavior in the face of a variety of IP addresses,
+then... just do it!
+
+This approach would have the web platform test runner configure additional
+loopback network devices and assign them `private` and `public` IP addresses.
```
antosart, line 84: more easily, couldn't you redirect the IPs to 127.0.0.1 (or some other localhost address) via iptables/netfilter?

rfcs/address_space_overrides.md
```diff
@@ -0,0 +1,111 @@
...
+Then we could configure the hosts file to resolve specific domains to those IP
+addresses, and test the whole shebang end to end.
+
+This approach's main pro is that it truly is an end to end test of the feature,
+exercising production code to its full extent. By virtue of requiring no
+test-only hooks, it also requires no work on the part of browser developers.
+
+Its main con is that it requires the test runner to have system privileges.
+Additionally, it requires work to support for each test execution environment.
+Finally, it arguably requires the purchase of a public IP address for our use,
```
antosart, line 94: Does it really require purchasing a public IP address? I am not sure why. Would it be so unacceptable to choose some arbitrary public IP address and redirect it, only for the purpose of the test execution, to localhost? Executing WP tests does not need access to public IP addresses, right? So I don't see room for conflicts.

rfcs/address_space_overrides.md
```diff
@@ -0,0 +1,111 @@
...
+This approach would have the web platform test runner configure additional
+loopback network devices and assign them `private` and `public` IP addresses.
+Then we could configure the hosts file to resolve specific domains to those IP
+addresses, and test the whole shebang end to end.
+
+This approach's main pro is that it truly is an end to end test of the feature,
+exercising production code to its full extent. By virtue of requiring no
+test-only hooks, it also requires no work on the part of browser developers.
+
+Its main con is that it requires the test runner to have system privileges.
```
antosart, line 92: That is true, but is it so unacceptable? In the automated test execution environment, we for sure can do it. For developers needing to test this locally, we can have instructions/an automated script to do it (similarly as the make /etc/hosts script). And we can document that those tests will fail otherwise if run locally.

rfcs/address_space_overrides.md
```diff
@@ -0,0 +1,111 @@
...
+
+Same as above, but allow overriding address spaces per IP address instead of
+per subnet.
+
+The marginal difficulty of overriding per subnet seems low, and provides plenty
+of test IP addresses to exercise intra-address-space, cross-ip-address requests,
+which might come in handy [in the
+future](https://github.com/WICG/cors-rfc1918/pull/1#issuecomment-721110250).
+
+### Use real IP addresses
```
antosart, line 78: I am curious why this is not a better alternative, since the main proposal seems to require quite some implementation effort for the browsers. Some comments below.

rfcs/address_space_overrides.md
Hexcles: Yes, but that'd also require root in the running environment. It also doesn't remove the requirement of owning the real IPs; otherwise, we'd essentially block those two IPs which may break things in unexpected/confusing ways (e.g. what if some real internet service depended by the runner resolves to one of these IPs?).

rfcs/address_space_overrides.md
Hexcles: Although the tests themselves don't need to access internet, the runner (especially in setup/teardown phases, e.g. to download pip packages, browser binaries, test prefs, etc.) usually does.

I guess we could carefully set up the redirection after all other setup finishes and right before tests start, and remove it at the end, but that'd require giving root to the test runner; see my reply to the other comment.

rfcs/address_space_overrides.md
Hexcles: AFAIK, LUCI Swarming bot doesn't give test code root permissions, but I'm not sure if that's a strong constraint. Also cc @jgraham on Gecko infra.

jgraham: I'm incredibly reluctant to have any requirement that the test harness runs with root permissions. Historically we've avoided this requirement even when there are tradeoffs in terms of testability (e.g. we don't run the server on ports 80/443, which limit our ability to test the different URL serializations in those cases). 

On gecko infra it *might* be possible to get root, although I'm pretty sure nothing does at the moment. We use a gecko pref for the DNS rather than editing the system `/etc/hosts`. For local gecko development there's no requirement to get root access except for installing the build tools. For web-platform-tests in particular you might need to add the hosts if you want to test in a non-Firefox browser or use the server standalone, but it isn't a requirement.

Requiring people to reconfigure their system network setup, or requiring the test runner to have root access so it can fiddle with the system setup, both seem like they have a high chance of ending up with user complaints either at the principle, or because the changes break their system in some way. The `/etc/hosts` modifications are borderline acceptable in that the actual modification is straightforward (just appending some lines to the hosts file) and works the same way on all systems we support. I'm extremely reluctant to add anything else which requires root, especially when it's not so easy to understand or debug.

antosart: I was not proposing to have the test runner run as root, but rather to have this setup as a separate step, and to have the infrastructure already set up for this.

Regarding local setup, I imagine not every developer wants to run every test on his machine, so this would be a problem of a very small number of people. But I do understand there are some cons here.

Hexcles: See my other comment for the downside of having this setup as a separate step (outside of test runner). The test runner still needs internet access, so we'd need some refactoring to pull that out; otherwise, we run the risk of obscuring legit IPs if we don't own them.

stephenmcgruer:
To keep this RFC up-to-date, at this point I believe WPT's stance is that this should be done via browser flags rather than have the testrunner (or some setup done by the testrunner) have system privileges. Both Blink and Gecko infra have said that having privileges in this way is at least not commonly done (but not explicitly banned).

We would like to see the aforementioned issues against browsers opened and discussions occur there (please link to them in this discussion). If a reasonable number of browsers think the level of support is unacceptably high, we can revisit the system-privilege angle.

letitz:
I will open issues against other browsers to seek their input. Thanks for the discussion so far!

letitz:
Filed [bug 1681316](https://bugzilla.mozilla.org/show_bug.cgi?id=1681316) against Firefox. Any help locating appropriate contacts within Mozilla to assign the bug to or cc on the bug is appreciated!

letitz:
Filed [bug 219642](https://bugs.webkit.org/show_bug.cgi?id=219642) against WebKit.

letitz:
Thanks @jgraham for helping triage the Firefox bug. Does anyone on this thread have a contact at WebKit I could ask to get some traction on that bug?

jgraham:
@gsnedders perhaps

letitz:
Thanks!

rfcs/address_space_overrides.md
```diff
@@ -0,0 +1,111 @@
...
+
+The goal of this RFC is to enable web platform tests to cover the
+[CORS-RFC1918](https://wicg.github.io/cors-rfc1918) specification entirely.
+More specifically, to enable tests to exercise user agent behavior in the face
+of responses served from `local`, `private` and `public`
+[address spaces](https://wicg.github.io/cors-rfc1918#address-space).
+
+This is achieved by passing command-line flags to the browser under test forcing
+it to artificially consider specific IP addresses `local`, `private` or
+`public`.
```
ddragana, line 13: What about:
have a command line flag or pref that overwrites an address. Browsers would overwrite the IP address just before creating a socket. For example map 100.10.10.10 to 127.0.0.1. The rest of the browser will only see 100.10.10.10, and only the socket will be created using 127.0.0.1

I think the approach will be less promte to implementation errors because it is simpler to implement.

letitz: Interesting! Thanks for reviewing.

That sounds OK to me! As long as the socket then reports back, when asked, that it is connected to the fake address, then I think it works. It would be injecting logic deep into the network stack for a feature that is otherwise situated at a layer far above the network.

It's not clear to me that this would be easier to implement, since it requires persisting the faked state somewhere in the network stack. The approach I suggested instead only requires injecting a lookup in some override map at address space computation time. That computation should already happen in some CORS-RFC1918-specific code.

WDYT?

letitz: @ddragana friendly ping :)

letitz: @Hexcles this would also alter the meaning of specific IP addresses in the context of the test, but not for the whole system. Do you have similar reservations about this approach as to the `/etc/hosts` approach or would this be fine with you?

Hexcles: This approach doesn't seem to require root AFAICT, so it seems fine, but let me pass it to @stephenmcgruer as I'm no longer working on WPT.

stephenmcgruer: Apologies for the delay; yes, from the WPT perspective I think that would be fine as it fits our preferences (no root, pass a command line flag). I think the big question would still be one of persuading other browser vendors :)

letitz: Thanks! I just need a signal from @ddragana now to fix up this PR and get to work :)

rfcs/address_space_overrides.md
letitz: I think this comment is addressed by replies below.

rfcs/address_space_overrides.md
letitz: Resolving this thread - discussion with browser vendors continues on the PR and in linked issues.

rfcs/address_space_overrides.md
letitz: I'll sketch out that approach as equivalent.

rfcs/address_space_overrides.md
letitz: I'll add some detail to this section to reflect this discussion.

rfcs/address_space_overrides.md
letitz: I'll add some detail to this section to make this point.

rfcs/address_space_overrides.md
letitz: Added a section to describe this alternative. @ddragana can you please take a look?

rfcs/address_space_overrides.md
letitz: Done.

rfcs/address_space_overrides.md
letitz: Friendly ping @ddragana - I'd like to get this RFC finalized.

ddragana: Sorry, I have missed the ping.
It looks fine.

letitz: Thank you, resolving.

letitz:
@stephenmcgruer I believe we have consent from Gecko (see @ddragana's [comment](https://github.com/web-platform-tests/rfcs/pull/72#discussion_r597200599) above) and WebKit (see [bug 219642](https://bugs.webkit.org/show_bug.cgi?id=219642).

OK with you to proceed?

stephenmcgruer:
> OK with you to proceed?

Unfortunately I'm no longer working on WPT matters, so I don't want to say yes when the consequences are on others. Instead, passing this to foolip@ and jgraham@ to sign off. For the record, I think this should be good to go based on https://github.com/web-platform-tests/rfcs/pull/72#issuecomment-737487271 and the subsequent work letitz has done.

letitz:
Thanks for the redirect! @jgraham @foolip WDYT?

jgraham (approved):
lgtm

foolip (approved):
I haven't followed the discussion leading up to this, but reading through the whole proposal it makes good sense to me. It seems like a good trade off among the alternatives considered.

letitz:
Thanks for the reviews! I'm not familiar with this workflow: how does this PR get merged now?

foolip:
As a newly returning member of the wpt core team there's a risk of embarrassing myself here, but per https://github.com/web-platform-tests/rfcs#the-rfc-process the latest "new information" would be your ping in https://github.com/web-platform-tests/rfcs/pull/72#issuecomment-805719591 and we can consider this accepted if there's no further feedback by March 31.

letitz:
Oh, I believed the previous discussion and your approvals were enough. It seems to me all stakeholders have had ample time to review this since November. Anyway, I won't haggle for 2 days after 5 months of waiting. Glad this is finally getting merged :)

foolip:
@letitz I'm only unsure about the status of WebKit and wanted to be sure @gsnedders had the full 7 days after this became ready to go. https://bugs.webkit.org/show_bug.cgi?id=219642 doesn't load for me right now, but if the discussion has already happened there, then really we could go ahead.

Please feel free to prepare an implementation PR in the meantime.

foolip:
Alright, two more days have passed, I'm merging this now.

PR #73, opened 2020-11-24
RFC 73 : Add testdriver.js support for the `timeZone` WebDriver command.

stephenmcgruer:


rfcs/testdriver_set_time_zone.md
```diff
@@ -0,0 +1,21 @@
+# RFC 73 : Add testdriver.js support for the `timeZone` WebDriver command.
+
+## Summary
+Add testdriver.js support for the `timeZone` WebDriver extension, which will
+allow setting the current timezone in the browser. This extension is specced as
+part of the [HTML spec](https://github.com/whatwg/html/pull/3047).
```
stephenmcgruer, line 6: The proposed text is currently in this comment: https://github.com/whatwg/html/pull/3047#issuecomment-732158684

This RFC should not be merged until the comment text is (a) part of the PR, and (b) the PR is either landed or fully LGTM'd and about to be landed.

mathiasbynens: My comment text is now part of the PR: https://github.com/whatwg/html/pull/3047

rfcs/testdriver_set_time_zone.md
```diff
@@ -0,0 +1,21 @@
+# RFC 73 : Add testdriver.js support for the `timeZone` WebDriver command.
+
+## Summary
+Add testdriver.js support for the `timeZone` WebDriver extension, which will
+allow setting the current time zone in the browser. This extension is specced
+as part of the [HTML spec](https://github.com/whatwg/html/pull/3047).
+
+## Details
+Changing the time zone is not available via normal web APIs. To test the effect
+of changing time zone on the `ontimezonechange` event there needs to be a way
```
mathiasbynens, line 10: Terminology nit:

```suggestion
of changing time zone on the `timezonechange` event there needs to be a way
```

`timezonechange` is the event. `ontimezonechange` is the corresponding event handler. https://annevankesteren.nl/2011/08/on-events

rfcs/testdriver_set_time_zone.md
```diff
@@ -0,0 +1,21 @@
...
+to instruct the browser to act as if the time zone is changing: this WebDriver
+extension provides that ability.
```
mathiasbynens, line 12: Worth mentioning that this command is also useful to test PWAs that take into account the user's local time zone? IOW, it's useful even if `timezonechange` didn't exist.

stephenmcgruer: Are you thinking from the context of WPT still, or from the context of a web-developer (or PWA developer) who might want to test their site/app?

mathiasbynens: Oh yeah, I was talking about the WebDriver command, not the testdriver.js change. Ignore me! :)

stephenmcgruer: Not at all - that info helps inform us that this WebDriver extension is probably useful outside of the scope of WPT. Those tend to be even easier to accept, since there's a valid use for them in the 'real world' :)

mathiasbynens:
The upstream HTML PR now includes the WebDriver command. Is anything else blocking this?

mathiasbynens:
Ah, per @stephenmcgruer's earlier comment:

> This RFC should not be merged until the comment text is (a) part of the PR, and (b) the PR is either landed or fully LGTM'd and about to be landed.

a) happened, but we're still working on b).

jgraham:
As per current policy this can be added to testdriver without an RFC, so I'm going to close this issue, but that shouldn't be interpreted as a rejection; just file a PR on testdriver when the feature lands and it can go through the normal code review process.

PR #76, opened 2021-03-02
RFC 76: http://wptpr.live turndown

stephenmcgruer:


jgraham:
I also have the impression this isn't much used, and don't have an alternate proposal to continue maintance, so I don't object to this plan.

stephenmcgruer:
This has been open with no objections for >7 days. I'm not sure whether to be worried by that or take it as a sign that nobody has serious concerns... In either case, it does need approved by someone (other than me) before its merged.

jgraham (approved):

Ms2ger (approved):
Nobody admitted to even knowing this existed, so I suppose it's fine.

domfarolino:
I just saw a link to this PR from another issue that Stephen commented on recently. Just FYI, I'm pretty sad to see it go. I got some good use out of it, but if there's no turning back, oh well I guess

PR #85, opened 2021-06-17
RFC 85: WebTransport over HTTP/3 test server (#85)

bashi:
Hi web-platform-tests,

We would like to introduce a new test server for WebTransport over HTTP/3 so that we can write WPTs for WebTransport. [Rendered version](https://github.com/bashi/rfcs/blob/08cd404ee62df4dd93b11175310d44fa98187c96/rfcs/webtransport_h3_test_server.md).

cc/ @yutakahirano @ricea

### DIFF
--- a/rfcs/webtransport_h3_test_server.md
+++ b/rfcs/webtransport_h3_test_server.md
@@ -4,7 +4,7 @@
 
 Add a [WebTransport over HTTP/3](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3-01) server to wpt. The server is used to test the [WebTransport Web APIs](https://w3c.github.io/webtransport/).
 
-The server will use a new [file name flag](https://web-platform-tests.org/writing-tests/file-names.html) `.webtransport` (`.webtransport.py`) for handlers. Handlers are python scripts which are used to respond WebTransport events.
+The server will use a new [file name flag](https://web-platform-tests.org/writing-tests/file-names.html) `.wt.h3` (`.wt.h3.py`) for handlers. Handlers are python scripts which are used to respond WebTransport events.
 
 This RFC shares the same background/motivation with [RFC 42](https://github.com/web-platform-tests/rfcs/blob/master/rfcs/quic.md) but focuses on WebTransport over HTTP/3. See RFC 42 for more background.
 
@@ -68,6 +68,12 @@ TODO: Spend more time on the API design. People may prefer this alternative appr
 When the server receives an [extended CONNECT method](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3-01#section-3.2) for WebTransport, it looks up a corresponding handler from the directory specified by a configuration. The configuration is passed by `wptserve`. If the server finds a corresponding handler the server creates a new WebTransport session with the handler and starts passing events to the handler.
 
 The server uses `":path"` header field of an extended CONNECT request to look up a handler. For example, if `":path"` is `/echo`, the server looks up `echo.webtransport.py` from the handler directory.
+
+### File name flags
+
+A new file name flag `.wt` is used to indicate a Python script is a WebTransport handler.
+
+`.h2` and `.h3` flags indicate that the underlying protocol (HTTP/2 or HTTP/3).
 
 ### `wptserve` integration
 
--- a/rfcs/webtransport_h3_test_server.md
+++ b/rfcs/webtransport_h3_test_server.md
@@ -1,0 +1,88 @@
+# RFC 85: WebTransport over HTTP/3 Test Server
+
+## Summary
+
+Add a [WebTransport over HTTP/3](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3-01) server to wpt. The server is used to test the [WebTransport Web APIs](https://w3c.github.io/webtransport/).
+
+The server will use a new [file name flag](https://web-platform-tests.org/writing-tests/file-names.html) `.webtransport` (`.webtransport.py`) for handlers. Handlers are python scripts which are used to respond WebTransport events.
+
+This RFC shares the same background/motivation with [RFC 42](https://github.com/web-platform-tests/rfcs/blob/master/rfcs/quic.md) but focuses on WebTransport over HTTP/3. See RFC 42 for more background.
+
+## Details
+
+### Implementation
+
+The WebTransport over HTTP/3 test server is built on top of `aioquic`, which is already added under `tools/third_party`. As of version 0.9.11, `aioquic` lacks some mechanisms to implement a WebTransport server (e.g. configuring SETTINGS HTTP/3 frame) but we plan to update `aioquic` once it supports necessary mechanisms.
+
+The server will be implemented under `tools/webtransport` directory.
+
+### Handlers
+
+A WebTransport handler is a Python script which contains a function `handle_event()`. Here is an example handler which echos back received data.
+
+```python
+from webtransport import (
+    WebTransportSession,
+    BidirectionalStreamDataReceived,
+    DatagramReceived,
+    WebTransportEvent
+)
+
+def handle_event(session: WebTransportSession,
+                 event: WebTransportEvent) -> None:
+    if isinstance(event, DatagramReceived):
+        session.send_datagram(data=event.data)
+    elif isinstance(event, BidirectionalStreamDataReceived):
+        session.send_stream_data(
+            stream_id=event.stream_id, data=event.data)
+```
+
+`session` is an object that represents a WebTransport over HTTP/3 session. It provides APIs to handle the session. Current proposed APIs can be found [here](https://bashi.github.io/wpt-wt-test-server-api/session.html).
+
+`event` represents a WebTransport event such as a reception of stream data. Current proposed events can be found [here](https://bashi.github.io/wpt-wt-test-server-api/events.html).
+
+#### Alternative handler design
+
+Alternatively, we can define a set of functions which are called upon each event. This API pattern is similar to existing [H2 handlers](https://web-platform-tests.org/writing-tests/h2tests.html). The equivalent of the above example might look like:
+
+```python
+def datagram_received(session: WebTransportSession,
+                      data: bytes):
+    session.send_datagram(data=data)
+
+
+def bidirectional_stream_data_received(session: WebTransportSession,
+                                       stream_id: int,
+                                       data: bytes):
+    session.send_stream_data(stream_id_id=stream_id, data=data)
+```
+
+A major downside of this approach is that keeping API documents and implementations in sync would require extra works, as H2 handler API document duplicates comments from the source.
+
+Another downside of this approach is that we won't have editor/IDE supports on what kind of callback functions are available.
+
+TODO: Spend more time on the API design. People may prefer this alternative approach.
+
+### Handler lookup
+
+When the server receives an [extended CONNECT method](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3-01#section-3.2) for WebTransport, it looks up a corresponding handler from the directory specified by a configuration. The configuration is passed by `wptserve`. If the server finds a corresponding handler the server creates a new WebTransport session with the handler and starts passing events to the handler.
+
+The server uses `":path"` header field of an extended CONNECT request to look up a handler. For example, if `":path"` is `/echo`, the server looks up `echo.webtransport.py` from the handler directory.
+
+### `wptserve` integration
+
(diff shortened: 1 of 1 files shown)


rfcs/webtransport_h3_test_server.md
```diff
@@ -0,0 +1,88 @@
+# RFC 85: WebTransport over HTTP/3 Test Server
+
+## Summary
+
+Add a [WebTransport over HTTP/3](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3-01) server to wpt. The server is used to test the [WebTransport Web APIs](https://w3c.github.io/webtransport/).
+
+The server will use a new [file name flag](https://web-platform-tests.org/writing-tests/file-names.html) `.webtransport` (`.webtransport.py`) for handlers. Handlers are python scripts which are used to respond WebTransport events.
```
yutakahirano, line 7: We may want to have handlers for WebTransport over HTTP/2 and map the same handler directory. How about adding .h3 to the script? The suffix will be : `.webtransport.h3.py` `.wt.h3.py`.

For the context, we are planning to have some tests as "generic" tests, running the same tests for WebTrasnport over HTTP/3 and WebTransport over HTTP/2. For example, the reliable streams behaviors are very similar with each other. We can write something like:

```
// This can be WebTransport over HTTP/3 or WebTransport
// over HTTP/2, depending on `server`.
const wt = new WebTransport(`${server}/echo`);
const stream = await wt.createBidirectionalStream();
// Test `stream`.
...
```

and we want to have `echo.wt.h3.py` and `echo.wt.h2.py`.

rfcs/webtransport_h3_test_server.md
```diff
@@ -0,0 +1,88 @@
...
+
+```python
+from webtransport import (
+    WebTransportSession,
+    BidirectionalStreamDataReceived,
+    DatagramReceived,
+    WebTransportEvent
+)
+
+def handle_event(session: WebTransportSession,
```
yutakahirano, line 31: Is session establishment an event? [tools/quic](https://github.com/web-platform-tests/wpt/tree/master/tools/quic) had a different function for the client indiciation.

rfcs/webtransport_h3_test_server.md
bashi: Thank you for the context. `wt.h3.py` and `wt.h2.py` look reasonable to me. I guess the content of handlers would be the same for most cases so we might also want `wt.any.py`.

rfcs/webtransport_h3_test_server.md
bashi: Yes, in this approach. Updated the [document](https://bashi.github.io/wpt-wt-test-server-api/events.html#webtransport.events.WebTransportConnectReceived).

In the alternative approach, we would have a function for session establishment.

rfcs/webtransport_h3_test_server.md
```diff
@@ -0,0 +1,94 @@
...
+                                       stream_id: int,
+                                       data: bytes):
+    session.send_stream_data(stream_id_id=stream_id, data=data)
+```
+
+A major downside of this approach is that keeping API documents and implementations in sync would require extra works, as H2 handler API document duplicates comments from the source.
+
+Another downside of this approach is that we won't have editor/IDE supports on what kind of callback functions are available.
+
+TODO: Spend more time on the API design. People may prefer this alternative approach.
```
jgraham, line 64: It feels like a seperate RFC is needed for the API design.

rfcs/webtransport_h3_test_server.md
```diff
@@ -0,0 +1,94 @@
+# RFC 85: WebTransport over HTTP/3 Test Server
+
+## Summary
+
+Add a [WebTransport over HTTP/3](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3-01) server to wpt. The server is used to test the [WebTransport Web APIs](https://w3c.github.io/webtransport/).
+
+The server will use a new [file name flag](https://web-platform-tests.org/writing-tests/file-names.html) `.wt.h3` (`.wt.h3.py`) for handlers. Handlers are python scripts which are used to respond WebTransport events.
```
jgraham, line 7: So, usually the filename flags are used to indicate that the test file itself is served over some particualr protocol. But I assume that webtransport tests will actually be testharness tests. So the `h3` flag in the test filename might be needed so the test can be same origin with the http3 server. But it's unclear why it would be needed in the name of the the `.py` file; that will just be loaded on whichever server the test requests.

Conversely the `.wt` part is presumably so that the server code knows to provide the correct python API (i.e. `handle_event` rather than `main`), so that seems reasonable to include in the name of the python file, but it's not needed in the name of the test itself.

I know the QUIC setup currently uses `.quic` in the name of Python files, but I think that's largely because the Python 3 requirement wasn't standard at the time it was introduced, so running the quic tests introduced an additional dependency that wasn't met in all environments.

rfcs/webtransport_h3_test_server.md
```diff
@@ -0,0 +1,94 @@
...
+
+Another downside of this approach is that we won't have editor/IDE supports on what kind of callback functions are available.
+
+TODO: Spend more time on the API design. People may prefer this alternative approach.
+
+### Handler lookup
+
+When the server receives an [extended CONNECT method](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3-01#section-3.2) for WebTransport, it looks up a corresponding handler from the directory specified by a configuration. The configuration is passed by `wptserve`. If the server finds a corresponding handler the server creates a new WebTransport session with the handler and starts passing events to the handler.
+
+The server uses `":path"` header field of an extended CONNECT request to look up a handler. For example, if `":path"` is `/echo`, the server looks up `echo.webtransport.py` from the handler directory.
```
jgraham, line 70: Why is this not just the wpt root, and the path contains the literal path to the Python file? e.g. if you want a handler in a file at `$WPT_ROOT/webtransport/handlers/echo.wt.py` you set the path to `/webtransport/handlers/echo.wt.py`? That seems easier for people debugging tests to understand (I think the websockets support acts more like you suggest, but that doesn't feel like a good design to copy since debugging those tests is always a pain for exactly this reason).

 Also note that your example here seems to be using a different naming convention to the section below.

rfcs/webtransport_h3_test_server.md
```diff
@@ -0,0 +1,94 @@
...
+
+`wptserve` runs the WebTransport over HTTP/3 test server in a way similar to [pywebsocket](https://github.com/web-platform-tests/wpt/blob/246a32576020cb9c4241b7cfbc296f92d944ff6b/tools/serve/serve.py#L713): `wptserve` launches the server as a daemon and passes relevant configurations (port number, handler directory etc) to the server. The daemon is monitored by `wptserve` like other daemons.
+
+### Dependencies
+
+As of writing this RFC, the only dependency is `aioquic`.
+
+### `tools/quic` deprecation
+
+There are no ongoing standardlization efforts which require a general QUIC server. We might remove `tools/quic` directory at some point.
```
jgraham, line 88: I think I'd like to encourage this to happen sooner rather than later. The QUIC support is some additional complexity that we'd ideally do without.

rfcs/webtransport_h3_test_server.md
yutakahirano: @jgraham FYI

jgraham: This still doesn't make sense to me :) There's no mechanism in wpt to ensure that something with `.h2` in the name is loaded over HTTP2 or that `.h3` in the name is loaded over HTTP3. Those are just different servers running on different ports. The only thing that cares about the flags is the test manifest generation and, indirectly, the test runner which uses the manifest data to decide which server to use to load the initial HTML test file in the browser.

I can see that the `.webtransport` flag could be used to decide which Python API should be used (although until now we just check which functions exist in the file rather than using a specific naming convention). But `echo.wt.py` is something that can be loaded via any supported HTTP version at the discretion of the HTML test, just by specifying the required server address.

I suppose `.any.js` tests look like they're an exception to this rule about only putting flags in the HTML test filename, but that's because we're (effectively) autogenerating the initial HTML file from the js file. If the plan is to autogenerate tests from just the handler then it would make sense to put the filename flags into the handler file, but I don't think that's correct?

yutakahirano: There are multiple points to discuss I think.

 1. Where is the document root for the WebTransport over HTTP/3 server?
 2. Should we have the name completion functionality (e.g., "echo" => "echo.wt.h3.py")?
 3. Should we have a convention for WebSocket over HTTP/3 script (e.g., .webstransport, .wt, .wt.h3, .wth3)?

You answers for the questions are:

 1. Where is the document root for the WebTransport over HTTP/3 server?
    The document root of wptserve.
 2. Should we have the name completion functionality (e.g., "echo" => "echo.wt.h3.py")?
    No, for simplicity.
 3. Should we have a convention for WebSocket over HTTP/3 script (e.g., `.webstransport`, `.wt`, `.wt.h3`, `.wth3`)?
   Maybe, fine with `.wt` but not fine with `.wt.h3`.

Is my understanding correct?

jgraham: Yes, I think that's correct.  terms of filename flags, I think we should only use them where they're functional, so that if you get them wrong the test doesn't work. So the question for the `.webtransport` flag on `.py` files is "is that required for the file to be handled correctly". As I understand the proposal, that would be required for the server to identify how to handle the file, so it does make sense. If we need to know whether something is a webtransport handler before the client tries to use it then a filename flag required, if we don't we could introspect the file when it's first loaded like we do for HTTP/2 handlers. But that's a conversation about tradeoffs rather than a hard requirement.

yutakahirano: I'm fine with the first two answers, but for the last I still want to have at least a convention to distinguish them from normal HTTP handlers, and from each other (WebTransport over HTTP/2 and WebTransport over HTTP/3).

ricea: I like the way all the WPT WebSocket handlers are organised in a single directory. For comparison, in the Blink WebSocket tests they are scattered among the html files and it's much harder to keep track of them.

We can still put the handlers in a separate directory even if the document root is the same as wptserve.

**Benefits of using the same document root:**
* The handlers are easy to find if the URL contains the directory name.
* Occasionally we've found it useful to do a normal HTTP fetch to the WebSocket server in a test. If it turns out we need to do this in a WebTransport test it would be easier if all the files were already visible under the document root.

**Disadvantages of using the same document root:**
* Every handler URL ends up having the same boilerplate path in it.
* The server will try to load random Python files if the URL points to them, which could have weird side effects.
* The principle of least privilege implies we should make as little as possible be visible from the WebTransport server to reduce the attack surface.

bashi: The current proposal doesn't require flag names to be functional. I feel we can submit a separate RFC for HTTP/2 vs HTTP/3. I think `.py` is sufficient for now and it's not difficult to add/modify handler names as needed.

Regarding discoverability, I think using the same root as wpt makes sense, and I agree with Adam that it's nice to put handlers in a single directory. It might be cumbersome having the boilerplate paths but I guess that's acceptable compromise.

Updated the PR to reflect above points.

rfcs/webtransport_h3_test_server.md
bashi: I decided to use this approach (callback function style) and abandoned the event style approach. I guess we don't need a separate RFC?

rfcs/webtransport_h3_test_server.md
bashi: Thank you for the context. I used `.wt` or `.webtransport` largely because `.quic` is used for QuicTransport. Name flags are removed from the current PR.

rfcs/webtransport_h3_test_server.md
bashi: Makes sense. Updated to set the root directory as the same as wpt root.

rfcs/webtransport_h3_test_server.md
bashi: I see. We will remove them once we introduce the WebTransport over HTTP/3 test server.

foolip:
@past could you take a look at this?

jgraham (approved):
THis lgtm now. Thanks for addressing the comments!

bashi:
Thank you for review, @jgraham!

@foolip It seems @past is OOO. Should I wait for his response, or do you think it's OK to merge this PR and start implementing the server?

foolip:
@bashi sorry I didn't check that first. I can take a look then, tomorrow. But I almost certainly won't spot anything major, so you could start preparing next steps.

foolip (approved):
Great writeup, thank you!

foolip:
A week has passed since the last updates here, so per [our process](https://github.com/web-platform-tests/rfcs#the-rfc-process) I declare this RFC accepted and will merge.

bashi:
Thank you foolip@ and others for reviews and feedback!

PR #89, opened 2021-07-27
RFC 89: [testdriver] Add an execute_script function to testdriver.

jgraham:
This provides a direct way to execute script in another browsing context.

### DIFF
--- a/rfcs/execute_script.md
+++ b/rfcs/execute_script.md
@@ -1,0 +1,82 @@
+# RFC 89: `testdriver` Add an `execute_script` method.
+
+## Summary
+
+Add an `execute_script` method to testdriver that takes a js function,
+arguments to pass to the function, and the context id to execute it it
+in a remote context. If the function returns a promise we return the
+resolved value of the promise, otherwise we return the return value of
+the function.
+
+## Details
+
+Sometimes a test window wants to check the status of something in a
+remote window. It's possible to do this in a number of ways, for
+example by posting messages from the remote window to the test
+window. However in cross-origin cases messge passing on the js side
+can be difficult or impossible. In addition when tests are defined in
+the test window, it's useful to avoid putting complex logic in other
+browsing contexts because these won't have the same error handling
+properties as the test window.
+
+A simple way to solve these problems is to provide a mechansim to
+execute script in a different context. This isn't possible using
+content APIs alone, but is possible in WebDriver, and we can reuse
+that to define a `execute_script` function that works across contexts:
+
+```
+execute_script(fn, args, context)
+```
+
+`body` is a js function for the script to execute. To pass it to the
+remote context, this is converted to a string using
+`toString()`. `args` is null or an array of arguments to pass to the
+function on the remote side. Arguments are passed as JSON. `context`
+is an object that can be resolved as a testdriver context id to
+identify the remote context (see [RFC
+88](https://github.com/web-platform-tests/rfcs/pull/88). If the return
+value of the function when executed in the remote context is a promise
+the promise returned by `execute_script` resolves to the resolved
+value of that promise. Otherwise the `execute_script` promise resolves
+to the return value of the function.
+
+The wptrunner implementation uses [Execute Async
+Script](https://w3c.github.io/webdriver/#execute-async-script). We
+wrap the provided script text and arguments into an `Execute Async
+Script` call as:
+
+```
+let callback = arguments[arguments.length - 1];
+let rv = ({function_string}).apply(null, {json.dumps(args)});
+return Promise.resolve(rv).then(callback)
+``
+
+## Example
+
+```
+<iframe src="child.html"></iframe>
+<script>
+setup({single_test: true});
+onload = {
+  let value = test_driver.execute_script(async (elemId) => {
+  await new Promise(resolve => onload(resolve))
+  return document.getElementById(elemId).textContent
+}, ["test"], iframe.contentWindow);
+  assert_equals(value, "PASS");
+  done();
+}
+</script>
+```
+
+## Risks
+
+WebDriver is single-threaded, so all the WebDriver-based actions need
+to be queued. That means that waiting for a promise in a remote
+context blocks all testdriver functionality that uses WebDriver, and
+also blocks the harness from returning results. This implies a high
+risk of timeouts.
+
+## References
+
+[PR 29803](https://github.com/web-platform-tests/wpt/pull/29803)
+contains a prototype implementation of this.


jgraham:
Gecko has something like this, but instead of passing in source text you pass in a function. Maybe we could have an API like
```
test_driver.execute_script(callable, {args: []})`
```
e.g.
```
test_driver.execute_script(async (elem_id) => {await new Promise(resolve => setTimeout(resolve, 0)); return document.getElementById(elem_id).innerHTML}, {args: ["test"]})
```

If the function returns a promise, we always await the promise, if it returns a value we always just return the value.

On the backend, we'd `toString()` the function and json-encode the args and do something like

```
webdriver.execute_async_script(f"""
let callback = arguments[arguments.length - 1];
let result = ({function_string)).call(null, ${json.loads(args_array)})
if (Promise.resolve(result) == result) {
    result.then(callback)
} else {
   callback(result)
}""")
```

rfcs/execute_script.md
```diff
@@ -0,0 +1,69 @@
...
+A simple way to solve these problems is to provide a mechansim to
+execute script in a different context. This isn't possible using
+content APIs alone, but is possible in WebDriver, and we can reuse
+that to define a `execute_script` function that works across contexts:
+
+```
+execute_script(body, async, context)
+```
+
+`body` is a function body for the script to execute. `async` controls
```
Ms2ger, line 30: `async` as an argument seems somewhat painful. Most likely, it'll always be a bare boolean constant at the call sites, which doesn't tend to be readable. Would adding a separate `execute_async_script` method be much more work?

jgraham: Agreed. I'm going to update the RFC with the proposal in https://github.com/web-platform-tests/rfcs/pull/89#issuecomment-888369412

jgraham:
I updated this based on https://github.com/web-platform-tests/rfcs/pull/89#issuecomment-888369412 But the more I think about it, the more worried I am that the fact that WebDriver is single-threaded and waiting on some complex script running in a remote context seems like it comes with a high risk of tests that have to be killed by the harness.

rfcs/execute_script.md
```diff
@@ -0,0 +1,82 @@
...
+The wptrunner implementation uses [Execute Async
+Script](https://w3c.github.io/webdriver/#execute-async-script). We
+wrap the provided script text and arguments into an `Execute Async
+Script` call as:
+
+```
+let callback = arguments[arguments.length - 1];
+let rv = ({function_string}).apply(null, {json.dumps(args)});
+return Promise.resolve(rv).then(callback)
+``
```
gsnedders, line 52: needs to be ```

rfcs/execute_script.md
```diff
@@ -0,0 +1,82 @@
...
+</script>
+```
+
+## Risks
+
+WebDriver is single-threaded, so all the WebDriver-based actions need
+to be queued. That means that waiting for a promise in a remote
+context blocks all testdriver functionality that uses WebDriver, and
+also blocks the harness from returning results. This implies a high
+risk of timeouts.
```
gsnedders, line 77: This also implies that recursive calls will never work, right?

jgraham: Yes.

rfcs/execute_script.md
```diff
@@ -0,0 +1,82 @@
...
+A simple way to solve these problems is to provide a mechansim to
+execute script in a different context. This isn't possible using
+content APIs alone, but is possible in WebDriver, and we can reuse
+that to define a `execute_script` function that works across contexts:
+
+```
+execute_script(fn, args, context)
+```
+
+`body` is a js function for the script to execute. To pass it to the
```
hiroshige-g, line 31: References in `fn` are evaluated on `context`, while `args` are evaluated on the caller context, right?
e.g. 

```
execute_script(
    (arg) => [location.href, `${location.href}`, arg],
    [location.href, `${location.href}`],
    context)
```
would result in [the URL of `context`, the URL of `context`, the URL of the caller of `execute_script`, the URL of the caller of `execute_script`].

jgraham: Yes, that's right.

hiroshige-g:
I tested BFCache test migration to jgraham's draft PR.

It works mostly similarly to the JavaScript-based COEP dispatcher framework, and I made a wrapper class that keeps most of the tests as-is.
Diffs/wrappers needed are uploaded at https://chromium-review.googlesource.com/c/chromium/src/+/3082215.

Functionality/implementation needed in the TestDriver version:

- A mechanism to wait for target context to be created or to become active (i.e. navigation completion triggered by `window.open`, back-navigation and restoration from BFCache, etc.).
  I made WPTRunner's `switch_to_window()` to retry until the target is found in my draft as a temporary workaround.
- Chromium-specific [testdriver-vendor.js](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/resources/testdriver-vendor.js) implementation (not implemented).
  Chromium's `run_web_tests.py` doesn't work with the current PR yet.
  I tested my BFCache draft tests using `testing/scripts/run_wpt_tests.py` that uses official wpt runner and the `tools/wptrunner/wptrunner/testdriver-extra.js` etc. implementation in this PR.

Differences that can handled by pure JS code changes (mainly around navigations):

- Races between injected script evaluation and page load (e.g. inline script evaluation, window load events, pageshow events, etc.).
    - In the TestDriver version, injected scripts can be evaluated even before inline script evaluation.
- Promises returned by `execute_script` should not be resolved after navigation.
    - The COEP framework can wait for a Promise created before navigation and then resolved after navigation/back-navigation from BFCache.
- When to (and when not to) wait for script evaluation results
    - In the COEP framework, I intentionally did not send the injected script result back to the caller if the injected script initiates navigation, because sending the result back via server stash + Fetch API after triggering navigation might disable BFCache. Perhaps there are no such things in the TestDriver version?
    - I added a little more `await` for `execute_script()` results, because I encountered timeouts when I called `execute_script()` twice without `await`ing for the first call (not investigated yet, but perhaps related to the single-threadedness of the WebDriver?).

Others:
- Chromium's `testing/scripts/run_wpt_tests.py` uses headless mode but BFCache is disabled in headless mode.
- Removing the headless flag enabled BFCache, but instead causes many more visibilitychange events, because switching tabs to find the target by WebDriver caused visibility changes.

rfcs/execute_script.md
fergald: what is "body" here? Should this be "fn"?

hiroshige-g:
Now I think I've figured out sufficient API specification to run BFCache tests at https://github.com/web-platform-tests/wpt/pull/28950 (See `/common/dispatcher/README.md`).
This is basically equivalent with `execute_script()`-related parts of RFCs 88/89/91, plus additional clarifications around navigation (like https://github.com/web-platform-tests/rfcs/pull/89#issuecomment-896327379), minus `testdriver` integration.

If it looks good, then I'd like to land the Pure-JS-based `RemoteContext.execute_script()` and BFCache tests in

- https://github.com/web-platform-tests/wpt/pull/29684 and
- https://github.com/web-platform-tests/wpt/pull/28950

and continue discussions on `test_driver` integration, as I think we will have sufficient concensus about the API surface itself, except for whether it is integrated with `test_driver`, and BFCache tests based on this can be migrated between JS-based and test_driver-integrated versions of `RemoteContext` with minimum changes.

Does this sound good?

([underlying design doc](https://docs.google.com/document/d/1vX_9vSDZJfzCFq4rjKbMoROpjabJM4ZX5Q0ELjR3b0I/edit?usp=sharing) for visibility, but most of the content is now included here and in https://github.com/web-platform-tests/wpt/pull/28950)

jgraham:
I don't have any objections to that. Note that https://github.com/web-platform-tests/rfcs/pull/90#issuecomment-914222361 has a new proposal (which I'll write up properly this week) to provide cross-context execute script using only js. I think it's basically compatible with what you're doing, so it should be easy to update your code to that model in the future if desired.

hiroshige-g:
Yeah it looks mostly compatible (I added some comments there but they don't affect much).
I'll aim to land the PRs relatively soon while I'll continue to follow the updates to the PRs.

hiroshige-g:
cc/ @ArthurSonzogni.

My proposal is at https://github.com/web-platform-tests/rfcs/pull/89#issuecomment-917260985, i.e. to have consensus around RFCs 88/89/91 plus BFCache-specific clarifications **except for `test_driver` integration and RFC 90**, which is summarized in https://github.com/web-platform-tests/wpt/pull/28950.

I'd like to land this part soon and after that continue discussion around `test_driver` integration and RFC 90 (including jgraham@'s upcoming RFC/impl updates at https://github.com/web-platform-tests/rfcs/pull/90#issuecomment-914222361) separately, because BFCache tests don't directly depend on these two (BFCache tests are using `send()`/`receive()` indirectly, but they are encapsulated under `RemoteContext.execute_script()`).

jgraham:
I'm withdrawning this in favour of https://github.com/web-platform-tests/rfcs/pull/98 which has a non-WebDriver based approach to the same problem.

PR #90, opened 2021-07-27
RFC 90: [testdriver] Primitives for cross-context messaging

jgraham:
Add send()/poll()/recv() functions that can be used to pass arbitary messages to
other testdriver contexts, to help with test scenarios requiring multiple origins

### DIFF
--- a/rfcs/cross_context_messaging.md
+++ b/rfcs/cross_context_messaging.md
@@ -2,7 +2,7 @@
 
 ## Summary
 
-Add an `send`/`poll`/`recv` methods to testdriver that can be used to
+Add `send`/`poll`/`recv` methods to testdriver that can be used to
 send messages to other browsing contexts, and receive messages for
 the current browsing context.
 
--- a/rfcs/cross_context_messaging.md
+++ b/rfcs/cross_context_messaging.md
@@ -1,0 +1,122 @@
+# RFC 90: `testdriver` Add a cross-context messaging primitive
+
+## Summary
+
+Add an `send`/`poll`/`recv` methods to testdriver that can be used to
+send messages to other browsing contexts, and receive messages for
+the current browsing context.
+
+## Details
+
+In tests that involve multiple origins, or multiple browsing context
+groups, there's often a need to make asserts about the behaviour of
+browsing contexts other than the test context. This problem has been
+solved by the COOP tests with
+[dispatcher.js](https://github.com/web-platform-tests/wpt/blob/7b0ebaccc62b566a1965396e5be7bb2bc06f841f/html/cross-origin-opener-policy/resources/dispatcher.js). This
+provides functions `send(uuid, message)` and `receive(uuid,
+maybe_timeout)` which are used to send and recieve messages between
+contexts given a shared key `uuid`. Typically the way this works is a
+specific resource load is given a `uuid` via a query parameter in the
+url. That resource will then `recieve()` messages using that `uuid`
+whereas other contexts, such as that containing the main test, can
+send messages using `uuid`.
+
+The backend transport layer is the wptserve stash, which stores a list
+of messages per `uuid`. A Python script `pop()`s the message from the
+from of the queue when it handles a `GET` request with `uuid` as a
+query parameter, and pushes the request body to the stash when it
+handles a `POST` request, also with a `uuid` parameter.
+
+This functionality is useful enough that it was adopted for Chrome's
+bfcache tests, which suggests it's likely to be useful as a primitive
+in many cases where tests involve multiple contexts or navigations.
+
+It is proposed that we lift this functionality into testdriver itself,
+integrating it with the context system used to target. The underlying
+communication is poll-based, so as a generalisation we expose three
+functions on `test_driver`, each of which returns a promise that is
+resolved when the relevant action is complete:
+
+```
+test_driver.send(message, context);
+test_driver.poll();
+test_driver.recv(timeout=null)
+```
+
+`test_driver.send(mesage, context)` Sends `message` to the context
+`context`. `message` must be JSON-encodable. Unlike the existing
+implementation we don't send `message` directly, but instead wrap it
+in a js object like:
+
+```
+{
+  src: <uuid of sending context>,
+  data: message,
+}
+```
+Including `src` in the wrapper means we're always able to reply to a
+message without needing to explictly communicate the uuid for the
+source context.
+
+`test_driver.poll()` polls the server for a message with the `uuid`
+for the current context. The return value is `null` if there is no
+message, or an object representing the wrapped message, if there is.
+
+`test_driver.recv(timeout=null)` waits at most `timeout` ms to receive
+a message, and throws an error if no message is received. If the
+timeout is null, the function polls indefinitely.
+
+As with the existing implementation, the backend is based on the
+server stash. Since this is standard wpt infrastructure not tied to
+wptrunner, it's possible to provide an implementation directly in
+`testdriver.js` as part of `test_driver_internal`, meaning that it can
+be overridden with a different transport if required, but there is no
+downstream work required for implementors as part of this feature.
+
+The use of the stash does present some promises; to avoid races if
+multiple processes try to access the same stash item we need to take a
+lock. This implies a performance cost to using the feature since all
+requests are serialised, even if there's multiprocessing. This could
+be reduced by using a per-key lock rather than a single global. The
+existing code also has a limit to the number of concurrent requests on
+the js side to avoid overwhelming the Python server.
+
+## Example
+
+```
+let uuid = "{{uuid()}}";
+window.open(`child.html?uuid=${uuid}`);
+await test_driver.send("start_test", uuid);
+let result = await test_driver.recv(uuid);
+assert_equals(result, "PASS");
+```
+
+## Risks
+
+The model in which we send messages to other contexts to get them to
+do work leaves authors with a lot of challenges for e.g. error
+handling. If the remote end throws an unexpected exception and fails
+to reply as expected it's likely we end up with a hard-to-debug
+timeout. It would be beneficial to supply some higher-level
+functionality for running parts of tests outside the main test window
+and communicating the result back to the text context.
+
+The promise-based polling model means that the default style of
+communication is command/response, where we `send()` a message to a
+remote context and `await recv()` a result. To implement something like
+events from the remote it would be necessary to build a custom event
+loop that continually polls and updates local state as
+necessary. Although this is possible, it would require repeated
+polling of the server, which is inefficient. This could be addressed
+by using Server-Sent-Events or websockets as the transport
+mechanism. These also expose incoming messages as DOM events rather
+than as promises. This probably makes simple cases more complex, but
+is easier in cases where there are an unknown number of events.
+
+## References
+
+[RFC 86](https://github.com/web-platform-tests/rfcs/pull/86) contains
+some discussion of the existing implementation and its reuse for bfcache.
+
+[PR 29803](https://github.com/web-platform-tests/wpt/pull/29803)
+contains a prototype implementation of this.


rfcs/cross_context_messaging.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Details
+
+In tests that involve multiple origins, or multiple browsing context
+groups, there's often a need to make asserts about the behaviour of
+browsing contexts other than the test context. This problem has been
+solved by the COOP tests with
+[dispatcher.js](https://github.com/web-platform-tests/wpt/blob/7b0ebaccc62b566a1965396e5be7bb2bc06f841f/html/cross-origin-opener-policy/resources/dispatcher.js). This
+provides functions `send(uuid, message)` and `receive(uuid,
+maybe_timeout)` which are used to send and recieve messages between
```
ArthurSonzogni, line 17: nit: "receive"

(×2)

rfcs/cross_context_messaging.md
```diff
@@ -0,0 +1,122 @@
...
+It is proposed that we lift this functionality into testdriver itself,
+integrating it with the context system used to target. The underlying
+communication is poll-based, so as a generalisation we expose three
+functions on `test_driver`, each of which returns a promise that is
+resolved when the relevant action is complete:
+
+```
+test_driver.send(message, context);
+test_driver.poll();
+test_driver.recv(timeout=null)
```
ArthurSonzogni, line 43: Shouldn't those specify the uuid of the queue your are receiving messages?

It is highly important to be able to listen to different queues. You might want to control multiple context or run multiple subtest in parallel. Having multiple queues is very important.

rfcs/cross_context_messaging.md
```diff
@@ -0,0 +1,122 @@
...
+```
+
+`test_driver.send(mesage, context)` Sends `message` to the context
```
ArthurSonzogni, line 46: nit: "message"

ArthurSonzogni: Optional: I like specifying the target context/uuid before the messages.:
```js
test_driver.send(context, `
  Hello from a different context
`);
```

rfcs/cross_context_messaging.md
jgraham: The way it's set up at the moment, you can only listen for messages sent to your queue. But each message has a `src` property that tells you where it came from. So you don't supply a UUID here because it's implied that it's the one for the current context.

rfcs/cross_context_messaging.md
jgraham: All the other testdriver commands take the context as the final argument, which is why I went with that order, even though I agree that this is more natural.

rfcs/cross_context_messaging.md
ArthurSonzogni: How can this be used to run several subtest in parallel?

Having a single queue per context is a strong limitation. As is, I don't believe this solution could replace code from:
- https://github.com/web-platform-tests/wpt/blob/master/html/cross-origin-embedder-policy/credentialless/resources/dispatcher.py
- https://github.com/web-platform-tests/wpt/blob/master/html/cross-origin-opener-policy/resources/dispatcher.py

rfcs/cross_context_messaging.md
ArthurSonzogni: ACK. That's unfortunate, but not a big deal.

rfcs/cross_context_messaging.md
jgraham: Presumably you could have some code like:

```
let awatingContexts = new Map();
fn messageFromContext(uuid) {
   new Promise(resolve => awaitingContexts.set(uuid, resolve));
};
async function dispatchMessages() {
  while (true) {
    let msg = test_driver.recv();
    let waiter = waitingContexts.get(msg.src);
    if (waiter) {
        waiter(msg.data)
    } 
  }
};
dispatchMessages();
```
And then a test would do something like
```
// Assume ctx is a uuid
await test_driver.send(msg , ctx);
let response = await messageFromContext(ctx);
```

So I don't think it's impossible, but I agree there's some additional overhead from the fact that the message queues are per context. But it feels like the per-context model is closer to the model for other testdriver commands. Do you think that adding some additional built-in API surface for this kind of message dispatch would be enough to make the solution work for you, or do you have another suggested approach?

ArthurSonzogni: Why does it has to belong to testdriver? I am not sure to understand the benefits. Maybe if I knew, they could potentially offset the drawback mentioned above?

I liked a lot hiroshige@ patches. Keep the message queue API very simple. Every context can pop/push to any arbitrary queue. Messages are simple string.
Then, once you have the capability to communicate, you can build second layers object like the Executor/ExecutorTarget, providing remote code execution, and returning result or errors.
1. https://chromium-review.googlesource.com/c/chromium/src/+/3033199
2. https://chromium-review.googlesource.com/c/chromium/src/+/3055392/2

ArthurSonzogni: I forgot... Most COEP:credentials are built this way:
1. Setup: Create multiple context with various different policies.
2. Subtest in parallel: interacting with the contexts from (1).

So they multiple message queues to communicate in between the driver and the helper contexts.

With the current solution, it is limited to one queue per receiver context. As you [said](https://github.com/web-platform-tests/rfcs/pull/90/files#r678412842), we can simulate more by dispatching per sender context. So at most we have one queue per (receiver, sender) pair, but this is not enough.

jgraham: It doesn't have to of course. But the reason for proposing this is to end up with a unified way of working with remote contexts, rather than having the concept of a remote context with a bunch of functionality like invoking user gestures, or running script, and a semi-decoupled model of messaging where the message queue ids are kind of coupled to context ids (because we give contexts an id through the URL), but anyone with the id has full read/write access. The model here seems closer to other things like `postMessage` where anyone (more or less) can write but only the target can read.

 I suppose another solution to the multi-test problem above would be introducing the ability to create non-context-bound message queues like the `MessagePort` DOM API. Then a test would create a context, and send a message to it containing the channel id to reply on. The test could await a message on that specific channel rather than doing dispatch on the src on the context-global channel.

rfcs/cross_context_messaging.md
```diff
@@ -0,0 +1,122 @@
+# RFC 90: `testdriver` Add a cross-context messaging primitive
+
+## Summary
+
+Add an `send`/`poll`/`recv` methods to testdriver that can be used to
```
Ms2ger, line 5:
```suggestion
Add `send`/`poll`/`recv` methods to testdriver that can be used to
```

rfcs/cross_context_messaging.md
```diff
@@ -0,0 +1,122 @@
...
+timeout is null, the function polls indefinitely.
+
+As with the existing implementation, the backend is based on the
+server stash. Since this is standard wpt infrastructure not tied to
+wptrunner, it's possible to provide an implementation directly in
+`testdriver.js` as part of `test_driver_internal`, meaning that it can
+be overridden with a different transport if required, but there is no
+downstream work required for implementors as part of this feature.
+
+The use of the stash does present some promises; to avoid races if
```
Ms2ger, line 76: Promises? Or challenges?

jgraham: It promises some challenges :p (but yes, good catch).

rfcs/cross_context_messaging.md
```diff
@@ -0,0 +1,122 @@
...
+the current browsing context.
+
+## Details
+
+In tests that involve multiple origins, or multiple browsing context
+groups, there's often a need to make asserts about the behaviour of
+browsing contexts other than the test context. This problem has been
+solved by the COOP tests with
+[dispatcher.js](https://github.com/web-platform-tests/wpt/blob/7b0ebaccc62b566a1965396e5be7bb2bc06f841f/html/cross-origin-opener-policy/resources/dispatcher.js). This
+provides functions `send(uuid, message)` and `receive(uuid,
```
ArthurSonzogni, line 16: Where does the UUID come from?
You said we can't generate them at will, but there will be instead one per context predefined.
Do we have an UUID for every type of context? (document, sharedworker, dedicatedworker, serviceworker). 
How do you convey the uuid from one context to the other?

ArthurSonzogni: I guess they come from the uuid params of the URL?

rfcs/cross_context_messaging.md
ArthurSonzogni: I am thinking, what about:
- Make the UUID of "receive" optional. By default it will listen to the current context predefined UUID.
- For other use case, like all mine, we can specify an arbitrary UUID.

rfcs/cross_context_messaging.md
jgraham: See https://github.com/web-platform-tests/rfcs/pull/88

jgraham:
So, to be clear I'm not wedded to the proposed setup. But I'm also skeptical of making the API surface a multi-producer, multi-consumer queue in which we expose a way for anyone with the token to both read and write. The problem with that design is that I fear it will lead to hard-to-debug flakiness where for timing reasons one context reads something that was supposed to be read by a different context, and the whole implicit protocol breaks down. Constraining who can read from a queue, even if multiple people can write to it, seems to encourage more robust designs.

That said, I agree that the proposal doesn't address the use case of multiple tests running concurrently, and all wanting to communicate back to the test context without needing complex dispatch. One way to extend the proposal to support that might be to allow something like `[receiver, sender] = test_driver.channel(); remote_context.send(sender); await reciever.recv()` i.e. you'd get seperate objects implementing the `send()` and `poll()/recv()` parts of the protocol, and it would be possible to `send()` the sender (but not the receiver) to a remote context so you'd get a unique channel per test to communicate back results. The test window, and and scriptable resource loaded with a uuid query parameter, would automatically implement the send and receive parts of the protocol.
 
 One thing that might be useful here is to look at use cases. For example:
* Test wants to get some data from a browsing context in a different browsing context group in order to make asserts about it
* Test wants to get some data from a worker running in a different origin to make some asserts about it
* Multiple subtests on the same page want to create browsing contexts/workers and get results back from those workers
* Test wants to observe events happening in a remote context/worker (e.g. check that a load event happens)

And maybe some higher level ones:
* Test wants to send over some script to run in a different context/worker and get the result of the execution back
* Want to run `assert()` functions outside the test window (in a different browsing context group) and communicate back the asserts run and their results (c.f. fetch_tests_from_window etc.).

Does that sound like a reasonable start? What am I missing?

ArthurSonzogni:
> Does that sound like a reasonable start? What am I missing?

I think we both understand each other.

The additional `test_driver.Channel()` can help, but this adds more work for you and for developers to use. Allowing a `uuid` params (even optional) would be simpler IMO.
I think you should implement all of your proposals and try them it as a replacement for the two `send()` `receive()` functions and `executor.html` file from:
https://github.com/web-platform-tests/wpt/blob/master/html/cross-origin-embedder-policy/credentialless/resources/dispatcher.py
This might help you figure out what is best, and helps me to realize this is a good fit.

Alternatively, you can also make COOP/COEP tests not to use this proposal, that's also a possibility. This proposal is good on its own, even if we don't manage to use it for the existing tests.

> one context reads something that was supposed to be read by a different context.

I would argue that with multiple queues, this precisely avoid having multiple readers reading the same queue.

jgraham:
OK, so I'm approximately ready to replace this PR with one that proposes a channel-based model.

I've updated the branch at https://github.com/web-platform-tests/wpt/pull/29803 with a WebSockets-backed channel-based proposal and converted a [bfcache test](https://github.com/web-platform-tests/wpt/pull/29803/commits/a5a6e871c478153abb3c26e6833040b2f580e81b) and a [COOP test](https://github.com/web-platform-tests/wpt/pull/29803/commits/3ce1c82f4b65ac269b27d3d2c1a77706daeb7ba4) to the new model. 

The COOP test in particular is very similar in the new model, but I think there are a few quality of life improvements:
* Because it supports serialzing complex data types rather than just strings, we can send functions with parameters to be executed on the remote rather than having to manually construct strings to eval. This means that syntax highlighting &c. works better so it's easier to write tests.
* The use of websockets backed by queues means we don't have to contually poll the server, which ought to reduce resource usage.
* The model is really close to WebDriver-BiDi, so hopefully at some point in the future we can replace the WebSockets backend with a WebDriver backend with minimal changes.

The bfcache test required some additional work, mostly because haing ongoing requests / open sockets prevents pages being added to the bfcache. This required a mechanism to stop new messages being read on the channels before navigation (so that the client websocket doesn't buffer a message, only for it to be lost after navigation). Obviously this is easier with a polling approach since the user code is in control of when we poll for new messages. If it turns out that this finegrained control is necessary, we could use the same server-side message storage approach but add a client API in which the client has to explictly `poll()` for a new message, causing a HTTP request which reads a message off the queue, if any is present. That could reuse most of the same frontend code.

If people think this is a good option to pursue I'll write up a full RFC for the new proposal, but some more technical details to help evaluate what's going on:
* Messages are stored on queues which are constrained to be multiple producer, single consumer.
* Queues are identified by a UUID.
* On top of the queues we build a channel abstraction. Channels can either send messages or receive messages, but not both. Only only message reader channel may be created per queue.
* Channels can send complex js objects, using a serialization format based on WebDriver-BiDi. It's also possible to send a `SendChannel`, giving the owner of the destination read channel the ability to send messages to the `SendChannel`'s corresponding `RecvChannel`.
* Layered on top of channels is a higher-level "testdriver" API (this doesn't actually depend on any other parts of testdriver, so might be badly named). This is accessed via a `TestDriverRemote` object with `executeScript(fn, ...args)`, `executeScriptNoResult(fn, ...args)` and `postMessage(msg)` methods. These are designed for performing actions in a remote context.  
* A page with a `uuid` parameter in the URL can call `ctx_channel` to create a `TestDriverCommandRecvChannel` for that uuid. This handles recieving messages sent over the "testdriver" API and actually executing the actions (e.g. running the script and returning the result).
* `TestDriverRemote` also has a `pause()` action which causes the reader channel to disconnect until explictly reconnected. This is useful for "send a command but don't allow the remote to process any more commands until after it completes some async action (e.g. navigation) and reconnects".
* There's a `closeAllChannelSockets()` method which is a low-level tool to ensure no websocket connections remain open. It's kind of a hack for bfcache tests and could probably be improved.

Anyway the general pattern is that in the main test file you have something like
```html
<title>Demo of executeScript</title>
<script src="/resources/testharness.js"></script>
<script src="/resources/testharnessreport.js"></script>
<script src="/resources/channel.sub.js"></script>
<script>
promise_test(async t => {
    let page = new TestDriverRemoteContext();
    window.open(`child.html?uuid=${page.uuid}`, "_blank", "noopener");
    assert_equals(await page.executeScript(() => document.body.textContent), "PASS");
})
</script>
```
And in `child.html` you have something like
```html
<!-- This is currently needed for reading the uuid, but could maybe be removed -->
<script src="/resources/testdriver.js"></script>
<script src="/resources/testdriver-vendor.js"></script>
<script src="/resources/channel.sub.js"></script>
<p>PASS</p>
<script>
// Listen for messages
ctx_channel();
</script>
```

ArthurSonzogni:
I took a quick look. I like the API and how it looks on the existing COOP test above.
Thanks for working on this! This will help writing many tests about navigations / iframes / popups.

I would be curious to see how `SendChannel` would look like if implemented.

jgraham:
I'm withdrawing this in favour of https://github.com/web-platform-tests/rfcs/pull/98 which has an alternate approach to the same problem, not based on polling.

PR #91, opened 2021-07-27
RemoteContext object for testdriver.

jgraham:
This provides convenient access to the testdriver API bound to a different context.

### DIFF
--- a/rfcs/testdriver_remote_context.md
+++ b/rfcs/testdriver_remote_context.md
@@ -1,0 +1,38 @@
+# RFC 91: `testdriver` RemoteContext object
+
+## Summary
+
+Provide `test_driver.RemoteContext` which is a convenient wrapper over
+directing commands to another browsing context.
+
+## Details
+
+Lots of testdriver functions take `context` as a final parameter. This
+is used when the test file wants to cause some action to happen in a
+different browsing context
+e.g. `test_driver.delete_all_cookies(context)`. If a context is being
+used repeatedly it's tedious and error-prone to have to keep supplying
+the context id explictly. Instead it's more convenient to work with an
+object with the context id already correctly bound.
+
+`RemoteContext` is an object that wraps the part of the `test_driver` API
+which takes a `context` parameter, and calls the underlying API with
+a context set at object creation:
+
+```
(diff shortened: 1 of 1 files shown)


Ms2ger (approved):

rfcs/testdriver_remote_context.md
```diff
@@ -0,0 +1,38 @@
...
+used repeatedly it's tedious and error-prone to have to keep supplying
+the context id explictly. Instead it's more convenient to work with an
+object with the context id already correctly bound.
+
+`RemoteContext` is an object that wraps the part of the `test_driver` API
+which takes a `context` parameter, and calls the underlying API with
+a context set at object creation:
+
+```
+let ctx = new test_driver.RemoteContext(frames[0]);
```
ArthurSonzogni, line 23: Often, there are no frame or WindowProxy for the remote context we want to execute data.

For instance, all of COOP is about switching browsing context group. This make the openee and the opener to belong to two group that can't access each other at all.

I would be nice being able to pass a UUID here.

jgraham: You can do that; it's anything that testdriver can treat as a context, which is basically either a UUID, a window handle, or an element that contains a nested browsing context.

jgraham: See RFC #88 for details on the proposal there.

jgraham:
I think this still makes sense, but should be done as a followup to https://github.com/web-platform-tests/rfcs/pull/98 as e.g.`RemoteWindow.testdriver.click` and similar.

rfcs/testdriver_remote_context.md
```diff
@@ -0,0 +1,38 @@
...
+As proposed creating the context might be quite verbose. Maybe
+`RemoteContext` should be on window instead (or have a shorter name).
+
+This is more code to maintain without adding much additional
+functionality.
+
+## References
+
+[PR 29803](https://github.com/web-platform-tests/wpt/pull/29803)
+contains a prototype implementation of this.
```
Ms2ger, line 38: This PR seems to have been merged without the `RemoteContext` bit.

PR #95, opened 2021-08-31
RFC 95: End support for Python 3.6 on macOS and Windows

foolip:


rfcs/py36.md
```diff
@@ -0,0 +1,28 @@
...
+- Between Dec 6 and Dec 17: Taskcluster is upgraded from Python 3.6 to 3.7
+- After Dec 23: Support for testing Python 3.6 is removed throughout WPT
+
+## Risks
+
+Some downstream users of WPT may still be using Python 3.6. TODO, confirm with:
+
+- [ ] Chromium
+- [ ] Gecko
+- [ ] WebKit
```
gsnedders, line 28: I can't speak for the non-Apple ports, but I believe the situation for the Apple one should be that 3.7 at a minimum is used (as that is what is included with Xcode 11, as used on Catalina).

foolip: That's great to know, thanks! Do you know which non-Apple ports also run WPT and should be checked? I know about WebKitGTK, but others?

jgraham:
I believe Gecko is currently using 3.6 in CI. @mitchhentges or @ahal might know if we have plans to require 3.7+.

foolip:
@jgraham that's great to know. Not what I was hoping, but nonetheless :)

gsnedders:
One option is to just move to testing on Python 3.7 on macOS, but keep Python 3.6 on Linux?

mitchhentges:
> I believe Gecko is currently using 3.6 in CI.

FWIW, Gecko uses 3.6 only for Linux CI work - Mac and Windows work uses 3.7+ (unsure if WPT is using Gecko Mac/Win workers or not, though :smile:)

There's two developments here which may affect this ticket:
* [Future work this year](https://bugzilla.mozilla.org/show_bug.cgi?id=1724277) will create Gecko CI tasks that fetch a target version of Python on-demand, rather than the one installed at the system level. Perhaps that can be utilized for WPT tasks.
* As Python 3.6 gets EOL'd this December, I'm relatively confident that we'll be updating our system-level Python in our Linux workers. This is a larger task, since Python's "minor" releases have backwards-incompatibilities, and we'd want our existing Linux work to still be "green", so I'm not sure how long it'll take. I don't know if there's a bug for this yet.

jgraham:
So it sounds like for Gecko bumping the Mac version to 3.7 in wpt CI should be fine; that will match Gecko CI. But we should keep Linux on 3.6 for now, with the hope of requiring 3.7+ by the time 3.6 is EOL (@mitchhentges feel free to CC me on any bugs related to this work, and I can be proactive about ensuring wpt tasks are using a more modern Python :) ).

rfcs/py36.md
Ms2ger: For WebKitGTK/WPE, I understand we promise to support Ubuntu Focal Fossa, which is on 3.8, and Debian Buster, which is on 3.7. I'm vaguely aware of Sony maintaining a port as well, but I don't know if they run WPT.

Ms2ger (approved):

foolip:
OK, so it sounds like what we should do is stop supporting 3.6 on macOS and Windows now (for convenience in Azure Pipelines, really) and then stop supporting 3.6 entirely in early 2022.

foolip:
I have now updated this to reduce the scope to just macOS and Windows. r?

jgraham (approved):

foolip:
Closing this, I no longer want to upgrade just macOS and Windows.

PR #96, opened 2021-09-14
Conditional requirements support in wptrunner

bashi:
Hi web-platform-tests,

I have a proposal to support conditional requirements in wptrunner.
Context: https://github.com/web-platform-tests/wpt/pull/30570

### DIFF
--- a/rfcs/conditional_requirements.md
+++ b/rfcs/conditional_requirements.md
@@ -1,0 +1,76 @@
+# Conditional requirements support in wptrunner
+
+## Summary
+
+Add a new field `conditional_requirements` in commands.json so that conditional
+requirements can be specified for `wpt` subcommands.
+
+## Details
+
+### Context
+
+`wpt` uses commands.json to set up subcommands (see `load_commands()` in
+tools/wpt/wpt.py). A subcommand can specify
+[requirements.txt](https://pip.pypa.io/en/stable/user_guide/#requirements-files)
+to install its dependencies in a virtualenv environment. These dependencies are
+unconditionally installed before running the subcommand.
+
+There is a demand to install additional dependencies conditionally. For example,
+a subcommand may need a third party library only when a specific command line
+flag is specified.
+
+```sh
+# This doesn’t require `aioquic` library
+$ ./wpt run
+# This requires `aioquic` library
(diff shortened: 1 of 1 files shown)


bashi:
@jgraham, @foolip What do you think?

@yutakahirano, @nidhijaju: FYI

annevk:
It would help me if this had a more detailed explanation as to why this is a good thing. It says "there is a demand", but I don't see the demand being explained.

bashi:
Thank you for feedback!

Does that mean the following example needs more explanations to clarify the benefits (not installing may-not-be-needed dependencies unconditionally)? Or are you suggesting that I should add alternatives considered (e.g. installing potential dependencies unconditionally) then discuss pros and cons (e.g. keeping wptrunner simple vs making wptrunner flexible for extension)?

annevk:
I was thinking about the former, but both sounds great. In particular without such information my bias would be toward making it easy to run all tests in all browsers.

bashi:
Thank you for the clarification. I updated the RFC and included these points.

+1 to making it easy to run all tests in all browsers. I'm happy to withdraw this proposal if adding aioquic dependency unconditionally makes sense :)

jgraham:
There are a couple of things to consider here:
* We already support installing per-browser dependencies in wptrunner e.g. we don't insist on installing mozprofile to run tests except in Firefox, because having access to the correct versions of dependencies in CI is not always trivial (e.g. browser CI systems are typically not able to install stuff directly from PyPI for reliability reasons). We could solve this by vendoring everything, but we don't have good infrastructure to keep vendored deps up to date (and actually it's worse than that because in gecko CI we want to use the packages in mozilla-central, not vendored copies). So optional dependencies in general are considered OK.
* In the specific case of aioquic, I think the problem is the requirement for a compiled component somewhere in the chain. That means we can't simply vendor in the dep and expect things to work out; we either need other CI systems to be able to compile from source or to provide a binary wheel using whatever package source they have. Making the dep optional in the first place helps vendors to roll out support without risking widespread breakage.

bashi:
Hi, any concerns on this proposal? If there is no substantive disagreement on this, I'd like to request merging this PR since 1 week has passed.

jgraham (approved):

PR #98, opened 2021-09-21
RFC 98: Remote channels for cross-browsing-group communication

jgraham:
[Rendered](https://github.com/web-platform-tests/rfcs/blob/remote_channel/rfcs/remote_channel.md)
[Implementation](https://github.com/web-platform-tests/wpt/pull/29803)

### DIFF
--- a/rfcs/remote_channel.md
+++ b/rfcs/remote_channel.md
@@ -526,8 +526,6 @@ class RemoteObject {
  * Close all websockets in the current global that are being used for channels.
  */
 function closeAllChannelSockets() {}
-
-///
 ```
 
 ### Resource Management
--- a/rfcs/remote_channel.md
+++ b/rfcs/remote_channel.md
@@ -22,7 +22,7 @@ the prior context. Typically in implementations each browsing context
 group is assigned a unique OS-level process.
 
 This creates some problems for testing; because testharness tests are
-running in a single browsing context, and test that involves a context
+running in a single browsing context, and tests that involves a context
 that is isolated from the context containing the test will be unable
 to communicate test results back to the harness using web platform
 APIs.
@@ -35,7 +35,7 @@ browser-specific techniques.
 
 ### Prior Art
 
-* Gecko tests able to provide cross-context communication using the
+* Gecko tests are able to provide cross-context communication using the
   parent (i.e. browser UI) process as an intermediary. The
   [SpecialPowers.spawn](https://searchfox.org/mozilla-central/source/testing/specialpowers/content/SpecialPowersChild.jsm#1547-1584)
   API available to gecko tests allows running a function in another
@@ -89,7 +89,7 @@ combines some of these strengths.
 
 #### Addressing
 
-Contexts that want to participate in messaging must have a parameter
+Contexts that want to participate in messaging must have a query parameter
 called `uuid` in their URL, with a value that's a UUID. This will be
 used to identify the channel dedicated to messages sent to that
 context. If the context is navigated the new document may reuse the
@@ -225,7 +225,7 @@ The low-level API provides required primitives, but it's difficult to
 use directly. To achieve the aim of making tests no harder to write
 than SpecialPowers-based Gecko tests, there is also a higher-level API
 initially providing two main capabilities: the ability to
-`postMessage` a remote context, and the ability to `executeScript` so
+`postMessage` to a remote context, and the ability to `executeScript` so
 that the script runs in a remote context.
 
 This API is provided by a `RemoteWindow` object. The `RemoteWindow`
@@ -237,19 +237,19 @@ created first and its `uuid` property used when constructing the URL.
 Inside the remote browsing context itself, the test author has to call
 `window_channel()` in order to set up a `RecvChannel` with UUID given
 by the `uuid` parameter in `location.href`. By default this is not
-connected until the `async connect()` method is called. This allows
+connected until the async `connect()` method is called. This allows
 message handlers to be attached before processing any messages. For
 convenience `await start_window_channel()` returns an already
 connected `RecvChannel`.
 
 The `RecvChannel` object offers an `addMessageHandler(callback)` API
 to receive messages sent with the `postMessage` API on `RemoteWindow`,
-and `async nextMessage()` to wait for the next message.
+and async `nextMessage()` to wait for the next message.
 
 ##### Script Execution
 
 `RemoteWindow.executeScript(fn, ...args)` serializes the function
-`fn`, using `Function.toString()`. Each argument is
+`fn`, using `Function.prototype.toString()`. Each argument is
 serialized using the remote value serialization algorithm. Along with
 the function string and the arguments, a `SendChannel` is sent for the
 command response (only one such channel is created per `RemoteWindow`
@@ -469,7 +469,7 @@ class RemoteWindow {
    *
    * Arguments and return values are serialized as RemoteObjects.
    */
-   async executeScript(fn: (args: ...any) => any, ..args: any): Promise<any> {}
+   async executeScript(fn: (args: ...any) => any, ...args: any): Promise<any> {}
 }
 
 /**
@@ -547,7 +547,7 @@ promise_test(async t => {
   let result = await remote.executeScript(id => {
     return document.getElementById(id).textContent;
   }, "test");
-  assert_equals("result", PASS);
+  assert_equals("result", "PASS");
 });
 </script>
 ```
@@ -608,7 +608,7 @@ testdriver integration is possible. For example we could add
 `set_permission` command in the remote context (and similarly for the
 remainder of the testdriver API). This would desugar to
 `testdriver.set_permission(params, uuid)` and testdriver would be
-update to identify the target window from the `uuid` parameter in its
+updated to identify the target window from the `uuid` parameter in its
 `location.href`.
 
 ## Risks
--- a/rfcs/remote_channel.md
+++ b/rfcs/remote_channel.md
@@ -158,16 +158,16 @@ functions shutdown when the socket is closed.
 
 #### Remote Object Serialization
 
-In order to allow passing complex JavaScript objects between contexts,
+In order to allow passing complex JavaScript values between contexts,
 a serialization format loosely based on the [WebDriver
 BiDi](https://w3c.github.io/webdriver-bidi/#type-common-RemoteValue)
 proposal is used. An object is represented using a JSON object as
 follows:
 ```js
 {
-  type: Name of the object type,
-  value: A representation of the object in a JSON-compatible form
-         where possible,
+  type: Name of the value type,
+  value: A JSON-compatiable representation of the value. Nested values are themselves
+         represented using this serialization.
   objectId: A unique id assigned to the object in case of circular references.
 }
 ```
--- a/rfcs/remote_channel.md
+++ b/rfcs/remote_channel.md
@@ -37,7 +37,7 @@ browser-specific techniques.
 
 * Gecko tests able to provide cross-context communication using the
   parent (i.e. browser UI) process as an intermediary. The
-  [SpecialPowers.spwan](https://searchfox.org/mozilla-central/source/testing/specialpowers/content/SpecialPowersChild.jsm#1547-1584)
+  [SpecialPowers.spawn](https://searchfox.org/mozilla-central/source/testing/specialpowers/content/SpecialPowersChild.jsm#1547-1584)
   API available to gecko tests allows running a function in another
   context, using structured clone to pass the arguments
   ([example](https://searchfox.org/mozilla-central/source/dom/tests/browser/browser_data_document_crossOriginIsolated.js#17)).
@@ -101,7 +101,7 @@ an error to use the same UUID for multiple contexts.
 The message passing backend is based on a Python
 [Queue](https://docs.python.org/3/library/queue.html) object per
 channel, stored in the server stash. The advantage of using queues
-here is that they are designed allow blocking reads. This means we
+here is that they are designed to allow blocking reads. This means we
 don't have to use a polling API but can, on the server side, wait for
 a message to be added to a channel, and immediately forward the
 message to the client. Because the stash itself runs in a single
@@ -488,7 +488,7 @@ function closeAllChannelSockets() {}
 
 With the stash-based approach, it's important to clean up the stash
 once no channels remain. Otherwise we end up leaking queues. The
-general approach is to explicitly store a refcounf as part of the
+general approach is to explicitly store a refcount as part of the
 stash value. Each socket that connects increments the refcount, and
 decrements it once the socket is closed. Access to the stash is
 protected by a lock, so only one socket may touch the refcount at a
@@ -520,7 +520,7 @@ test.html
 
 <script>
 promise_test(async t => {
-  let remote = new RemoteContext();
+  let remote = new RemoteWindow();
   window.open(`child.html?uuid=${remote.uuid}`, "_blank", "noopener");
   let result = await remote.executeScript(id => {
     return document.getElementById(id).textContent;
@@ -612,5 +612,5 @@ contains a prototype implementation of this.
 
 <!--  LocalWords:  UUID WebDriver wptrunner testharness APIs UI
  -->
-<!--  LocalWords:  testdriver websockets bfcache refcounf
+<!--  LocalWords:  testdriver websockets bfcache
  -->
--- a/rfcs/remote_channel.md
+++ b/rfcs/remote_channel.md
@@ -1,0 +1,616 @@
+# RFC 98: Cross-Window Channels
+
+## Summary
+
+Provide a channel API which provides support for server-mediated
+messaging between browsing contexts, including those which are in
+different browsing context groups, and so cannot communicate purely
+from the client.
+
+## Details
+
+### Use Case
+
+The modern web platform has features which allow isolating different
+browsing contexts and scripting environments from each other such that
+they are unable to communicate. For example windows opened with the
+`noopener` attribute will not have a handle to their parent, nor will
+the parent have a handle to the child. Similarly, cross-origin
+navigations with the `cross-origin-opener-policy` header appropriately
+set create a new browsing context group which is totally isolated from
+the prior context. Typically in implementations each browsing context
+group is assigned a unique OS-level process.
+
+This creates some problems for testing; because testharness tests are
+running in a single browsing context, and test that involves a context
+that is isolated from the context containing the test will be unable
+to communicate test results back to the harness using web platform
+APIs.
+
+Given the importance of isolation in the modern web platform, it's
+important to improve support for tests involving multiple browsing
+context groups in web-platform-tests. Otherwise test authors will
+either not write tests for these scenarios, or write them using
+browser-specific techniques.
+
+### Prior Art
+
+* Gecko tests able to provide cross-context communication using the
+  parent (i.e. browser UI) process as an intermediary. The
+  [SpecialPowers.spwan](https://searchfox.org/mozilla-central/source/testing/specialpowers/content/SpecialPowersChild.jsm#1547-1584)
+  API available to gecko tests allows running a function in another
+  context, using structured clone to pass the arguments
+  ([example](https://searchfox.org/mozilla-central/source/dom/tests/browser/browser_data_document_crossOriginIsolated.js#17)).
+
+* The
+  [cross-origin-opener-policy](https://github.com/web-platform-tests/wpt/tree/7b0ebaccc62b566a1965396e5be7bb2bc06f841f/html/cross-origin-opener-policy)
+  tests currently in web-platform-tests use a bespoke
+  [dispatcher.py](https://github.com/web-platform-tests/wpt/blob/7b0ebaccc62b566a1965396e5be7bb2bc06f841f/html/cross-origin-opener-policy/resources/dispatcher.py)
+  to allow cross-origin communication. This uses a queue (implemented
+  as a list) stored in the wpt server stash object as a message
+  channel. Each context is given a UUID in a URL parameter named
+  `uuid`, and uses HTTP GET requests to check for any message in the
+  queue stored in the stash entry with that UUID. Other contexts that
+  know the UUID can POST a message that's appended to the queue.
+
+* WebDriver also has the ability to communicate between different
+  contexts, and [RFC
+  89](https://github.com/web-platform-tests/rfcs/pull/89) proposed
+  adding an `execute_script` method to testdriver to enable running a
+  script in a different context. However the design of WebDriver and
+  wptrunner means that only one context can communicate using
+  WebDriver at a time, and for testharness we must always return
+  control to the test window, since WebDriver is also used to
+  communicate test results back to the harness. The single threaded
+  nature of WebDriver, and requirement that all testdriver actions go
+  via the main test window, make it hard to build an ergonomic
+  cross-context messaging API.
+
+### Proposal
+
+All the above examples of prior art have some attractive features, and
+it's possible to combine them in a way that should provide an
+ergonomic API for web-platform-tests
+
+* The Gecko SpecialPowers API appears to offer the best ergonomics,
+  providing a relatively high level API for executing script in a
+  remote context. However obviously the implementation depends on
+  gecko specifics.
+
+* The dispatcher framework provides an addressing mechanism (the
+  `uuid` parameter in URLs) that is workable in the context of
+  web-platform-tests, and the use of the server stash as the backend
+  makes sense for wpt. However the requirement to poll the server for
+  messages, and the relatively low level API based on passing strings
+  around, seem like areas for improvement.
+
+The following subsections will set out a proposal for an API that
+combines some of these strengths.
+
+#### Addressing
+
+Contexts that want to participate in messaging must have a parameter
+called `uuid` in their URL, with a value that's a UUID. This will be
+used to identify the channel dedicated to messages sent to that
+context. If the context is navigated the new document may reuse the
+same UUID if it wants to share the same message queue. Otherwise it's
+an error to use the same UUID for multiple contexts.
+
+#### Backend
+
+The message passing backend is based on a Python
+[Queue](https://docs.python.org/3/library/queue.html) object per
+channel, stored in the server stash. The advantage of using queues
+here is that they are designed allow blocking reads. This means we
+don't have to use a polling API but can, on the server side, wait for
+a message to be added to a channel, and immediately forward the
+message to the client. Because the stash itself runs in a single
+Python process and communicates with the server processes via IPC, it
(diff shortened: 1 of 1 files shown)


rfcs/remote_channel.md
```diff
@@ -0,0 +1,616 @@
...
+important to improve support for tests involving multiple browsing
+context groups in web-platform-tests. Otherwise test authors will
+either not write tests for these scenarios, or write them using
+browser-specific techniques.
+
+### Prior Art
+
+* Gecko tests able to provide cross-context communication using the
+  parent (i.e. browser UI) process as an intermediary. The
+  [SpecialPowers.spwan](https://searchfox.org/mozilla-central/source/testing/specialpowers/content/SpecialPowersChild.jsm#1547-1584)
```
Ms2ger, line 40:
```suggestion
  [SpecialPowers.spawn](https://searchfox.org/mozilla-central/source/testing/specialpowers/content/SpecialPowersChild.jsm#1547-1584)
```

rfcs/remote_channel.md
```diff
@@ -0,0 +1,616 @@
...
+context. If the context is navigated the new document may reuse the
+same UUID if it wants to share the same message queue. Otherwise it's
+an error to use the same UUID for multiple contexts.
+
+#### Backend
+
+The message passing backend is based on a Python
+[Queue](https://docs.python.org/3/library/queue.html) object per
+channel, stored in the server stash. The advantage of using queues
+here is that they are designed allow blocking reads. This means we
```
Ms2ger, line 104:
```suggestion
here is that they are designed to allow blocking reads. This means we
```

rfcs/remote_channel.md
```diff
@@ -0,0 +1,616 @@
...
+function closeAllChannelSockets() {}
+
+///
+```
+
+### Resource Management
+
+With the stash-based approach, it's important to clean up the stash
+once no channels remain. Otherwise we end up leaking queues. The
+general approach is to explicitly store a refcounf as part of the
```
Ms2ger, line 491:
```suggestion
general approach is to explicitly store a refcount as part of the
```

rfcs/remote_channel.md
```diff
@@ -0,0 +1,616 @@
...
+```html
+<!doctype html>
+<title>executeScript example</title>
+<script src="/resources/testharness.js">
+<script src="/resources/testharnessreport.js">
+<script src="/resources/channel.js">
+
+<script>
+promise_test(async t => {
+  let remote = new RemoteContext();
```
Ms2ger, line 523:
```suggestion
  let remote = new RemoteWindow();
```

rfcs/remote_channel.md
```diff
@@ -0,0 +1,616 @@
...
+  across multiple windows).
+
+## References
+
+[PR 29803](https://github.com/web-platform-tests/wpt/pull/29803)
+contains a prototype implementation of this.
+
+<!--  LocalWords:  UUID WebDriver wptrunner testharness APIs UI
+ -->
+<!--  LocalWords:  testdriver websockets bfcache refcounf
```
Ms2ger, line 615:
```suggestion
<!--  LocalWords:  testdriver websockets bfcache
```

jgraham:
Gentle ping here. This has been open for a week, so per the process I could go ahead and merge it, but I'm pretty sure you don't want me to do that ;)

rfcs/remote_channel.md
```diff
@@ -0,0 +1,616 @@
...
+On the remote side the function is deserialized and executed. If
+execution results in a `Promise` value, the result of that promise is
+awaited. The final return value after the promise is resolved is sent
+back and forms the async return value of the `executeScript` call. If
+the script throws, the thrown value is provided as the result, and
+re-thrown in the originating context. In addition an `exceptionDetails`
+field provides the line/column numbers of the original exception,
+where available.
+
+In addition there is a `RemoteWindow.executeScriptNoResult(fn,
```
hiroshige-g, line 249: Based on my experience on bfcache tests, I feel `executeScriptNoResult()` might be hard to use, because:

- Mixing `await executeScriptNoResult()` and `await executeScript()` might complicates the timeline. Having `await executeScript()` only would be clear, because it serializes the script evaluation on caller and remote context.
- The choice between `await executeScriptNoResult()` and `await executeScript()` might depend on internals of `executeScript()`. When I was migrating my bfcache tests from COOP-based framework to test_driver-based framework, I had to make changes around `executeScriptNoResult()`-equivalent that were hard to explain without knowing the internal structure of the framework, so I switched to the `executeScript()`-only interface.

So I'd prefer removing `executeScriptNoResult()`.
Still there would remain complexities around navigation: instead of `executeScriptNoResult()`, I created promises / async callbacks triggered by `executeScript()` but not blocking `async executeScript()` in bfcache tests. But by removing `executeScriptNoResult()` I expect we can make the complexities only navigation-specific, not affecting `RemoteWindow` interface.

rfcs/remote_channel.md
```diff
@@ -0,0 +1,616 @@
...
+`start_window()` is a pretty nondescript method name.
+
+#### Navigation and bfcache
+
+For specific use cases around bfcache, it's important to be able to
+ensure that no network connections (including websockets) remain open
+at the time of navigation, otherwise the page will be excluded from
+bfcache. In the current prototype this is handled as follows:
+
+* A `pause` method on `SendChannel`. This causes a server-initiated
```
hiroshige-g, line 267: As commented at https://github.com/web-platform-tests/wpt/pull/29803#discussion_r707551472, I'm not sure how `pause` works and whether `pause` is needed (the draft impl doesn't use `pause`).

rfcs/remote_channel.md
```diff
@@ -0,0 +1,616 @@
...
+  WebDriver at a time, and for testharness we must always return
+  control to the test window, since WebDriver is also used to
+  communicate test results back to the harness. The single threaded
+  nature of WebDriver, and requirement that all testdriver actions go
+  via the main test window, make it hard to build an ergonomic
+  cross-context messaging API.
+
+### Proposal
+
+All the above examples of prior art have some attractive features, and
```
hiroshige-g, line 71: nit: Maybe it would be clearer to move these blocks to "Prior Art" section, because the list here isn't a part of the proposal (e.g. this RFC doesn't propose SpecialPowers-integrated API).

rfcs/remote_channel.md
```diff
@@ -0,0 +1,616 @@
...
+for a `SendChannel` listens for incoming messages and writes them to
+the corresponding `Queue`, whilst the websocket handler for a
+`RecvChannel` essentially blocks on the `get()` method for the `Queue`
+and writes and message it receives to the socket. In detail some
+additional care is needed to ensure that the websocket handler
+functions shutdown when the socket is closed.
+
+#### Remote Object Serialization
+
+In order to allow passing complex JavaScript objects between contexts,
```
hiroshige-g, line 152: nit: It might be better to clarify that this serialization is implemented by pure JavaScript (IIUC `RemoteObject.serialize()` in the draft PR) and thus doesn't depend on WebDriver BiDi impl itself.

rfcs/remote_channel.md
```diff
@@ -0,0 +1,616 @@
...
+  script in a different context. However the design of WebDriver and
+  wptrunner means that only one context can communicate using
+  WebDriver at a time, and for testharness we must always return
+  control to the test window, since WebDriver is also used to
+  communicate test results back to the harness. The single threaded
+  nature of WebDriver, and requirement that all testdriver actions go
+  via the main test window, make it hard to build an ergonomic
+  cross-context messaging API.
+
+### Proposal
```
hiroshige-g, line 69: I'd like to clarify (somewhere in the RFC) what are the dependencies of this RFC in terms of implementation.

IIUC this RFC can be implementable by pure JavaScript outside WPT infra + existing server-side stash, which is good because we can run the tests by running wpt server and bare browsers, without further depending on WPT test infra.
This RFC mentions integration with WebDriver BiDi, test_driver and testharness, but these are out of scope of this RFC itself.

Is my understanding correct?
Why the changes to `resources/testharness.js` and `tools/wptrunner/wptrunner` are needed in https://github.com/web-platform-tests/wpt/pull/29803 ?

(Other changes, in JS/HTML files, `tools/wptserve/wptserve/stash.py`, and `websockets/handlers/msg_channel_wsh.py` look like "implementable by pure JavaScript outside WPT infra + existing server-side stash")

rfcs/remote_channel.md
```diff
@@ -0,0 +1,616 @@
...
+## Details
+
+### Use Case
+
+The modern web platform has features which allow isolating different
+browsing contexts and scripting environments from each other such that
+they are unable to communicate. For example windows opened with the
+`noopener` attribute will not have a handle to their parent, nor will
+the parent have a handle to the child. Similarly, cross-origin
+navigations with the `cross-origin-opener-policy` header appropriately
```
hiroshige-g, line 19: Is this RFC going to replace the framework in the COOP/COEP tests?

rfcs/remote_channel.md
```diff
@@ -0,0 +1,616 @@
...
+where available.
+
+In addition there is a `RemoteWindow.executeScriptNoResult(fn,
+...args)` method. This works the same way except no channel is passed,
+and so no result is returned. This can be useful in case the script
+does something like trigger a navigation, so there's no need to
+synchronize the navigation starting (which will close the socket) with
+writing the response.
+
+TODO: the naming here isn't great. In particular a `RemoteWindow`
```
hiroshige-g, line 256: Doesn't `RemoteContext` work (as proposed in https://github.com/web-platform-tests/rfcs/pull/91)?

rfcs/remote_channel.md
```diff
@@ -0,0 +1,616 @@
...
+without providing the API surface that only makes sense in a test
+window.
+
+It may be possible in the future to replace the backend with a
+WebDriver BiDi based message passing system that would avoid the need
+for websockets. By sticking close to WebDriver BiDi proposed semantics
+the transition may even be seamless.
+
+testdriver integration is possible. For example we could add
+`RemoteContext.testdriver.click` to execute a click in the remote
```
hiroshige-g, line 574: Would `RemoteContext.testdriver.click` be a pure JavaScript wrapper like `RemoteContext.executeScript(() => test_driver.click())`?

rfcs/remote_channel.md
```diff
@@ -0,0 +1,616 @@
...
+});
+```
+
+Currently this is possible if `file.html` is put in `resources` or
+`support` directory so it can't be detected as a test. But it would
+perhaps be better to have a minimal subset of testharness available
+without providing the API surface that only makes sense in a test
+window.
+
+It may be possible in the future to replace the backend with a
```
hiroshige-g, line 568: For further integration with WebDriver, testdriver etc., discussion about benefits and risks would be helpful (not necessarily now though, given that this is "Possible Future Additions").

One one hand ("Risk" side), I basically prefer not integrating the framework too much to WebDriver/test_driver etc., to make the framework work with minimum dependencies and work with bare browsers, and not to make debugging harder. We might want to `JSON.stringify()` instead of WebDriver BiDi format if we want to reduce complexity around serialization, given that the current tests considered so far don't need WebDriver BiDi capability beyond `JSON.stringify()`.
So I'm interested in hearing more about the benefits (from the point of view of test writers and WPT infra maintainers) when we consider the further integration.

rfcs/remote_channel.md
```diff
@@ -0,0 +1,616 @@
...
+In addition there is a `RemoteWindow.executeScriptNoResult(fn,
+...args)` method. This works the same way except no channel is passed,
+and so no result is returned. This can be useful in case the script
+does something like trigger a navigation, so there's no need to
+synchronize the navigation starting (which will close the socket) with
+writing the response.
+
+TODO: the naming here isn't great. In particular a `RemoteWindow`
+could actually be some other kind of global like a worker, and
+`start_window()` is a pretty nondescript method name.
```
hiroshige-g, line 258: I did a similar thing to `start_window()` in my bfcache tests and I was also not sure about the better naming, better semantics, etc. (and thus I just made it `const executor = new Executor(uuid);` and avoided explicit naming).
So I'm also curious how this should be named.

hiroshige-g:
Commented somehow verbosely to move discussion forward.

As for `executeScript()`, the RFC basically looks similar to what the current `RemoteContext.execute_script()` and bfcache tests do, while there are more primitives around navigations (`executeScriptNoResult`, `pause` and `closeAllChannelSockets` in this RFC while the bfcache tests only have `closeAllChannelSockets`-equivalent, as commented inline above) and I'd like to understand the differences (IIUC there aren't fundamental differences, because I expect switching the impl from Fetch API to websocket doesn't impact so much).
So in terms of `executeScript()`, this RFC looks like switching from Fetch API to websocket based implementation (compared to the previous RFCs and https://github.com/web-platform-tests/wpt/pull/28950), both pure JavaScript-based, and thus doesn't cause test bodies except for mechanical changes (correct?).

The RFC is more about replacing the existing `send()` and `receive()` primitives currently used in COEP/COOP tests, by introducing related API classses, adding more APIs in addition to `send()`/`receive()` etc., and I expect test writers (other than me) and WPT infra people might have more interests on these.

cc/ @ArthurSonzogni and @foolip.

rfcs/remote_channel.md
jgraham: So (for the context of others) the technical concern here is that for bfcache tests in particular you need all sockets to be closed before initiating the navigation. The problem is that if you send a message, execute some script that closes all sockets, and then send the result, you end up re-opening a socket for the result. So there are broadly two choices:
* Don't try to send a result at all. `postMessage` works like this, and `executeScriptNoResult` does the same: we simply don't provide a channel to send the result on so there's no concern about sockets being reopened.
* Send a result, but ensure we navigate after the result has been sent, so nothing will implictly reopen a socket. Practically this means something like `setTimeout(async url => {await closeAllSockets(); startNavigation(url)}, 0)`. The problem with that is that we are currently linting for `setTimeout` so the infra discourages even usage that's required like this.

I don't quite get the argument about serialization; in the case that you're putting the actual action of the remote script in a timeout the test window can guarantee the script has been sent, but the actual action of the script hasn't happened at the time the value is returned, so it doesn't seem like that method offers particuarly stronger guarantees around ordering. I suppose one advantage is that any exception that happens when evalutaing the script is reported through the return value, rather than just happening in the remote window.

So, I think I'm mostly OK with dropping `executeScriptNoResult` because, as you say, it's exposing an implementation detail and it seems cleaner to start with a smaller API. My concern is that for some kinds of tests (notably bfcache), understanding that implementation detail is very important, and there's an argument that making it an explicit part of the API will help people write correct tests. So I'd like to hear other opinions here.

rfcs/remote_channel.md
jgraham: So the problem pause is intended to solve is where you have something like:
```
await remote.executeScript(() => startNavigation());
await remote.executeScript(() => ensureNavigationHappened());
```
With a `poll()` based interface this can work as written; the remote simply doesn't call `poll()` after `startNavigation` and before the navigation is complete. But with websockets, we don't have any synchronisation between the navigation and sending the `ensureNavigationHappened` message. If the test sends the second message before the socket is closed on the remote, it can end up being lost. So pause solves this by effectively providing a synchronisation point around the socket being closed:
```
await remote.executeScript(() => startNavigation());
remote.pause();
// We now know the remote now won't listen for messages until it reconnects to the socket after navigation
await remote.executeScript(() => ensureNavigationHappened());
```

From the point of view of the remote, `closeAllSockets()` and `pause()` do race, but the effect of `pause()` is a server-initiated disconnect of the read channel for the remote, and the effect of `closeAllSockets` is a client-initiated disconnect of all channel-related sockets, so the ordering isn't very important.

We can avoid requiring pause by not sending requests to a remote after initiating a navigation, before getting a confirmation that the navigation is actually complete. In practice this means that the target page has to be a different `RemoteWindow` (i.e. have a different uuid) and has to process a message i.e. something like:
```
await remote.executeScript(() => startNavigation());
await navigationTargetRemote.executeScript(() => goBack());
await remote.executeScript(() => ensureNavigationHappened());
```
That's what the bfcache tests are doing now, and why `pause` isn't needed there, but either way the ergonomics are a bit tricky. I think we could drop it from the initial featureset, but these races around navigation are tricky to understand and debug.

rfcs/remote_channel.md
jgraham: I don't expect the RFC to require moving the COOP/COEP tests. However I expect the RFC to provide all the features needed for those tests, and would anticipate the actual migration happening as a followup once the implementation has landed.

rfcs/remote_channel.md
jgraham: No, because `RemoteContext` is already used by testharness.js to mean something slightly different:
```
    /*
     * A RemoteContext listens for test events from a remote test context, such
     * as another window or a worker. These events are then used to construct
     * and maintain RemoteTest objects that mirror the tests running in the
     * remote context.
     */
```

In the long term it would make sense to unify these features, but in the short term a naming conflict will cause breakage.

rfcs/remote_channel.md
jgraham: That doesn't work as written, because all testdriver commands have to go via the test window since that's the one that webdriver is using. So it would desugar into `testdriver.click(RemoteContext.uuid)`, and testdriver would learn to look up contexts from a `uuid` parameter in the URL, as well as from the internal identifier it uses today.

rfcs/remote_channel.md
jgraham: Yes, future additions here would be subject to the same RFC process. So the idea was just to sketch out some possibilites in case people felt there were specific parts that should be given additional consideration in the initial design, or would be high value to work on sooner.

In terms of the serialization format, we are already using extra complexity beyond JSON-stringify when serializing a `SendChannel` for message responses. We obviously could special case that, but in general I think using plain JSON as a serialization format for data where we might want to support non-JSON types is a bad idea (e.g. current WebDriver mostly uses plain JSON but also has super-hacky support for serializing Element objects which doesn't generalise to other types and is generally a mess). Aiming for a featureset closer to structured clone seems more future proof, but structured clone doesn't actually define a serialization. So, given that WebDriver BiDi has very similar requirements, and actually defines the wire format, it seems sensible to use that as a reference point. The fact that, going forward, we could use WebDriver BiDi instead of the stash as the backend in the wptrunner implementation is good, but it's not the primary motivation here.

jgraham:
@hiroshige-g Thanks for the detailed comments, it's much appreciated! I've responded to the technical concerns, hopefully in a way that also provides context for other reviewers. I'll update the RFC for the more editorial issues. Please let e know if I miss[ed] anything.

rfcs/remote_channel.md
jgraham: I've clarified this below, let me know if you think more is required.

jgraham:
Ping, again. This has now gone another week since the last round of updates without further review.

rfcs/remote_channel.md
```diff
@@ -0,0 +1,650 @@
...
+* Having to manually pass around UUIDs to identify channels makes it
+  easy to author broken tests (e.g. trying to reuse the same UUID
+  across multiple windows).
+
+## References
+
+[PR 29803](https://github.com/web-platform-tests/wpt/pull/29803)
+contains a prototype implementation of this.
+
+<!--  LocalWords:  UUID WebDriver wptrunner testharness APIs UI
```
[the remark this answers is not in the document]
foolip: Do you want to keep them, then?

rfcs/remote_channel.md
```diff
@@ -0,0 +1,650 @@
...
+combines some of these strengths.
+
+#### Addressing
+
+Contexts that want to participate in messaging must have a parameter
+called `uuid` in their URL, with a value that's a UUID. This will be
+used to identify the channel dedicated to messages sent to that
+context. If the context is navigated the new document may reuse the
+same UUID if it wants to share the same message queue. Otherwise it's
+an error to use the same UUID for multiple contexts.
```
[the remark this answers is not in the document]
foolip: I see, I thought it might silently work.

rfcs/remote_channel.md
```diff
@@ -0,0 +1,650 @@
...
+
+The message passing backend is based on a Python
+[Queue](https://docs.python.org/3/library/queue.html) object per
+channel, stored in the server stash. The advantage of using queues
+here is that they are designed to allow blocking reads. This means we
+don't have to use a polling API but can, on the server side, wait for
+a message to be added to a channel, and immediately forward the
+message to the client. Because the stash itself runs in a single
+Python process and communicates with the server processes via IPC, it
+is sufficient to use a thread-based `Queue` rather than requiring a
```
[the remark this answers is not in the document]
foolip: I didn't realize it was a question about how `Queue` works, but I guess it was. I assumed reads were blocking, but it has arguments, and you're using `queue.get(True, 1)` in a loop, so the problem I imagined doesn't exist.

rfcs/remote_channel.md
```diff
@@ -0,0 +1,650 @@
...
+convenience `await start_window_channel()` returns an already
+connected `RecvChannel`.
+
+The `RecvChannel` object offers an `addMessageHandler(callback)` API
+to receive messages sent with the `postMessage` API on `RemoteWindow`,
+and `async nextMessage()` to wait for the next message.
+
+##### Script Execution
+
+`RemoteWindow.executeScript(fn, ...args)` serializes the function
```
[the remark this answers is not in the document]
foolip: Just `call` seems OK.

rfcs/remote_channel.md
```diff
@@ -0,0 +1,650 @@
...
+`fn`, using `Function.toString()`. Each argument is
+serialized using the remote value serialization algorithm. Along with
+the function string and the arguments, a `SendChannel` is sent for the
+command response (only one such channel is created per `RemoteWindow`
+for efficiency reasons), and a command id is sent to disambiguate
+responses from different commands.
+
+On the remote side the function is deserialized and executed. If
+execution results in a `Promise` value, the result of that promise is
```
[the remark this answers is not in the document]
foolip: There's a similar thing that happens in BiDi where I think it could be distinguished, but not here, I think you're right.

rfcs/remote_channel.md
```diff
@@ -0,0 +1,650 @@
...
+awaited. The final return value after the promise is resolved is sent
+back and forms the async return value of the `executeScript` call. If
+the script throws, the thrown value is provided as the result, and
+re-thrown in the originating context. In addition an
+`exceptionDetails` field on the response provides the line/column
```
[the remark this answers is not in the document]
foolip: I see. I think I must have misread this too, thinking that an `exceptionDetails` property would be added to the re-thrown error, but this is talking about a response.

rfcs/remote_channel.md
```diff
@@ -0,0 +1,650 @@
...
+
+  /**
+   * Send a message `msg`. The message object must be JSON-serializable.
+   */
+  async send() {msg: Object}
+
+ /**
+  * Disconnect the RecvChannel, if any, on the server side
+  */
+  async pause() {}
```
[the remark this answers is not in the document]
foolip: Resolving this, you renamed it `disconnectReader`.

rfcs/remote_channel.md
```diff
@@ -0,0 +1,650 @@
...
+
+ /**
+   * Close the channel and underlying websocket connection
+   */
+  async close() {}
+
+  /**
+   * Add a message handler function
+   */
+  addEventListener(fn: (msg: Object) => void) {}
```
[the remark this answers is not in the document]
foolip: With `type` added I think it's EventTarget-like enough for nobody to notice the difference.

rfcs/remote_channel.md
```diff
@@ -0,0 +1,650 @@
...
+stash value. Each socket that connects increments the refcount, and
+decrements it once the socket is closed. Access to the stash is
+protected by a lock, so only one socket may touch the refcount at a
+time. In practice the fact that there can only be one `RecvChannel`
+per queue means that we can just count `SendChannel` sockets and
+enforce an at-most-one rule for `RecvChannel`. The underlying queue is
+deleted when no socket connections remain. This does mean that if we
+were to e.g. put a message on a queue, then delete all the channels
+before forwarding the message, and then use the UUID to create a new
+`RecvChannel` the queue would be deleted and the message lost. One can
```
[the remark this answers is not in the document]
foolip: Well, I think the scheme you have here shouldn't leak more and more data as the tests run, so that's the most important thing. If the only scenario where the queue might be deleted too soon is when one deliberately reuses a UUID for a second `RecvChannel`, then that's not worth trying to fix.

Honestly I don't trust that I can see all the potential issues here, better to think about that when reviewing the implementation.

rfcs/remote_channel.md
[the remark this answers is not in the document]
jgraham: Well it's convenient for me when editing, but I could remove them once the overall RFC is approved, right before merging.

rfcs/remote_channel.md
```diff
@@ -0,0 +1,650 @@
...
+```html
+<!doctype html>
+<title>executeScript example</title>
+<script src="/resources/testharness.js">
+<script src="/resources/testharnessreport.js">
+<script src="/resources/channel.js">
+
+<script>
+promise_test(async t => {
+  let remote = new RemoteWindow();
```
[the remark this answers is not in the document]
foolip: Alright, enough bikeshedding :)

rfcs/remote_channel.md
```diff
@@ -224,77 +233,76 @@ as the result of deserialization.
...
+function in the remote global with given arguments.
+
 
-This API is provided by a `RemoteWindow` object. The `RemoteWindow`
-object doesn't handle creating the browsing context, but given the
-UUID for the remote window, creates a `SendChannel` which is able to
-send messages to the remote. Alternatively the `RemoteWindow` may be
-created first and its `uuid` property used when constructing the URL.
+This API is provided by a `RemoteGlobal` object. The `RemoteGlobal`
+object doesn't handle creating the browsing context (or other global
```
foolip, line 242: Good clarification!

rfcs/remote_channel.md
```diff
@@ -224,77 +233,76 @@ as the result of deserialization.
...
+  send a command that will initiate a navigation, then without knowing
+  when the navigation is done, send further commands that will be
+  processed when the `RecvChannel` reconnects. If the commands are
+  sent before the navigation, but not processed, they can be buffered
+  by the remote and then lost during navigation. An alternative here
+  would be a more explicit protocol in which the remote has to send an
+  explicit message to the test page indicating that it's done
+  navigating and it's safe to send more commands. But the way the
+  messaging works, it's hard for a random page that's loaded to
+  initiate a connection to the top-level test context For example,
```
foolip, line 298:
```suggestion
  initiate a connection to the top-level test context. For example,
```

foolip (approved):
Just some missing punctuation.

PR #99, opened 2021-10-07
RFC 99: Interop 2022

foolip:
[Rendered](https://github.com/web-platform-tests/rfcs/blob/compat_2022/rfcs/interop_2022.md) (easier to review than the raw Markdown)

Proposals, meeting agenda, etc., are in https://github.com/web-platform-tests/interop-2022 

This RFC is **ready for review** as of Jan 21.

### DIFF
--- a/rfcs/interop_2022.md
+++ b/rfcs/interop_2022.md
@@ -2,9 +2,11 @@
 
 ## Summary
 
-Identify focus areas and a list of tests for Interop 2022, a public effort and benchmark around which browser vendors can collaborate to improve browser compatibility in areas believed to be important to web developers.
+Interop 2022 is a public effort and benchmark around which browser vendors can collaborate to improve interoperability in areas believed to be important to web developers. 10 new focus areas have been selected, and 5 have been carried forward from Interop 2021.
 
-Proposed focus areas are tracked in the dedicated [Interop 2022 repo](https://github.com/web-platform-tests/interop-2022).
+For each area, a set of tests have been identified, and progress is tracked with scores for each area as well as for the overall effort. There will be a dashboard for this at https://wpt.fyi/interop-2022.
+
+See the [Interop 2022 repo](https://github.com/web-platform-tests/interop-2022) for all of the proposals and discussions that led to the current set of areas.
 
 ## Details
 
--- a/rfcs/interop_2022.md
+++ b/rfcs/interop_2022.md
@@ -2,25 +2,128 @@
 
 ## Summary
 
-Identify focus areas and a list of tests for Interop 2022, a public effort and benchmark around which browser vendors can collaborate to improve browser compatibility in areas believed to be important to web developers. Proposed focus areas are tracked in the dedicated [Interop 2022 repo](https://github.com/web-platform-tests/interop-2022).
+Identify focus areas and a list of tests for Interop 2022, a public effort and benchmark around which browser vendors can collaborate to improve browser compatibility in areas believed to be important to web developers.
 
-**Note: This RFC is initially a call for proposals and will be updated include the specific areas before being considered ready for final review.**
+Proposed focus areas are tracked in the dedicated [Interop 2022 repo](https://github.com/web-platform-tests/interop-2022).
 
 ## Details
 
-Following [MDN Web DNA 2019](https://insights.developer.mozilla.org/reports/mdn-web-developer-needs-assessment-2019.html), the [browser compat report](https://insights.developer.mozilla.org/reports/mdn-browser-compatibility-report-2020.html) narrowed in on a small number of concrete areas where web developers struggle with browser compatibility. Based on that and [other signals](https://web.dev/compat2021/#choosing-what-to-focus-on), [Compat 2021](https://wpt.fyi/compat2021) was created. Over the course of 2021, that effort has improved test results in the focus areas considerably, and the [mid-year update](https://web.dev/compat2021-midyear/) outlines what this means in practice for web developers.
+10 new focus areas are part of Interop 2022, as well as 5 from Interop 2022.
 
-While Compat 2021 was discussed among all major browser vendors, it didn't follow the web-platform-tests RFC process, and only Google and Microsoft have publicly committed to the effort. Interop 2022 is proposed the successor to Compat 2021, but following a public process of nomination and review, using the RFC process. The term "interop" is used instead of "compat" to avoid conflation with "site compat", see [terminology](#terminology) for details.
+### Cascade Layers
 
-No organization is required to make any public commitments.
+Proposal: https://github.com/web-platform-tests/interop-2022/issues/5
+Tests: [`interop-2022-cascade`](https://wpt.fyi/results/?label=master&label=experimental&product=chrome&product=firefox&product=safari&aligned&q=label%3Ainterop-2022-cascade)
+
+### Color Spaces & CSS Color Functions
+
+Proposal: https://github.com/web-platform-tests/interop-2022/issues/20
+
+Tests: [`interop-2022-color`](https://wpt.fyi/results/?label=master&label=experimental&product=chrome&product=firefox&product=safari&aligned&q=label%3Ainterop-2022-color)
+
+### Containment
+
+Proposal: https://github.com/web-platform-tests/interop-2022/issues/19
+
+Tests: [`interop-2022-contain`](https://wpt.fyi/results/?label=master&label=experimental&product=chrome&product=firefox&product=safari&aligned&q=label%3Ainterop-2022-contain)
+
+
+### `<dialog>` and `::backdrop`
+
+Proposal: https://github.com/web-platform-tests/interop-2022/issues/12
+
+Tests: [`interop-2022-dialog`](https://wpt.fyi/results/?label=master&label=experimental&product=chrome&product=firefox&product=safari&aligned&q=label%3Ainterop-2022-dialog)
+
+### Form Controls
+
+Proposal: https://github.com/web-platform-tests/interop-2022/issues/11
+
+Tests: [`interop-2022-forms`](https://wpt.fyi/results/?label=master&label=experimental&product=chrome&product=firefox&product=safari&aligned&q=label%3Ainterop-2022-forms)
+
+
+### Scrolling
+
+Proposals: https://github.com/web-platform-tests/interop-2022/issues/14 + https://github.com/web-platform-tests/interop-2022/issues/21
+
+Tests: [`interop-2022-scrolling`](https://wpt.fyi/results/?label=master&label=experimental&product=chrome&product=firefox&product=safari&aligned&q=label%3Ainterop-2022-scrolling)
+
+### Subgrid
+
+Proposal: https://github.com/web-platform-tests/interop-2022/issues/1
+
+Tests: [`interop-2022-subgrid`](https://wpt.fyi/results/?label=master&label=experimental&product=chrome&product=firefox&product=safari&aligned&q=label%3Ainterop-2022-subgrid)
+
+### Text
+
+Proposal: https://github.com/web-platform-tests/interop-2022/issues/13
+
+Tests: [`interop-2022-text`](https://wpt.fyi/results/?label=master&label=experimental&product=chrome&product=firefox&product=safari&aligned&q=label%3Ainterop-2022-text)
+
+### New Viewport Units
+
+Proposal: https://github.com/web-platform-tests/interop-2022/issues/4
+
+Tests: `interop-2022-viewport` (TODO)
+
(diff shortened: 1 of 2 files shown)


foolip:
In the meeting yesterday (agenda in https://github.com/web-platform-tests/interop-2022/issues/24) we decided to have a hard stop on new proposals. Many good proposals were filed after Nov 7 which was the original timeline in this RFC, and we'll consider those, but no new proposals filed after Nov 18. In other words, https://github.com/web-platform-tests/interop-2022/issues/31 is the last proposal.

Note that https://github.com/web-platform-tests/interop-2022/labels/compat-bug-proposal aren't standalone proposals but part of https://github.com/web-platform-tests/interop-2022/issues/9.

gsnedders:
I'd rather we actually defined a slightly more concrete scoping of what we want Interop 2022 to be.

At the moment, we have some proposals like https://github.com/web-platform-tests/interop-2022/issues/9 which stray from what I understand to be the [original notion of Compat 2021](https://web.dev/compat2021/#background):

> Compatibility on the web has always been a big challenge for developers. In the last couple of years, Google and other partners, including Mozilla and Microsoft, have set out to learn more about the top pain points for web developers, to drive our work and prioritization to make the situation better. This project is connected to Google's Developer Satisfaction (DevSAT) work, and it started on a larger scale with the creation of the MDN DNA (Developer Needs Assessment) surveys in 2019 and 2020, and a deep-dive research effort presented in the MDN Browser Compatibility Report 2020. Additional research has been done in various channels, such as the State of CSS and State of JS surveys.
>
> The goal in 2021 is to eliminate browser compatibility problems in five key focus areas so developers can confidently build on them as reliable foundations. This effort is called #Compat 2021.

See also the [introduction to Google's DevSAT](https://web.dev/developer-satisfaction/):

> Web Developer Satisfaction, shortened DevSAT, is a Google project to gather information about the needs of web developers. Our goal is to make sure developers are being heard, know what's happening with their feedback, and see issues being addressed. DevSAT is how we ensure we are focusing on the right areas with our Engineering and Developer Relations work. The goal is a more reliable, predictable, and interoperable web platform, enabling developers to invest in and trust it, and to adopt and use new features to grow the platform and their business.

How to mash this with Compat 2021/Interop 2022 is a slightly harder question, especially when there are cases when developers largely just complain about bugs in general (and I don't think "fix all browser-specific-failures in WPT" is a reasonable goal for Interop as a project!). As such we need to be selective about what we include, and for 2021 this was done largely by choosing five specific features. I don't think it's entirely infeasible to say "there are a whole load of important bugs we should target", but then that gets into questions about *how* we pick bugs, especially given we probably don't want to target _overly_ specific things.

Arguably there are two separate challenges, which I think the current process is showing:

1. In general, there's no broad triage done of developer feedback. A lot of the early parts of the process we're going through are doing fairly fundamental work trying to triage developer feedback, and it's not at all clear that WPT is the right place to be doing that. (In fact, it's almost certainly the wrong place.) \
\
This is work that arguably makes more sense to be done across SDOs, both in terms of prioritising standardisation work (which, admittedly, SDOs tend to have little direct control over) and in terms of advising those around the SDOs what the pain-points of the platform being standardised are when they are implementation concerns.\
\
A bunch of this is rooted in the fact that parts of what has come up from the State of CSS is very much things with ongoing standardisation work, and really should play into prioritises in that space as much as anything else.
2. Then, come out with a shortlist for Interop 2022 based on the output of this triage. Trying to figure out how small (or long) of a list of features we want is an open question, though questions about achievability within a year are reasonable to ask.

annevk:
I don't think that taking 2021 as the blueprint is the way to go, as that didn't have cross-browser agreement. The issue you point to above highlights issues where Firefox has seen some of the most significant webcompat challenges due to a lack of interop. It's our position that those are worthy of inclusion. I think we should strive for some healthy mix between challenges directly relayed by web developers and issues browsers have found as part of their webcompat efforts.

I'm not sure looking at web platform SDO is helpful as by-and-large they work on what browsers want to work on, which isn't necessarily related to the most pressing interop problems.

jgraham:
In addition to what @annevk said, I think the process so far has revealed a set of things where we aren't yet in a position to track implementation compliance against a set of tests, because the standards work isn't done. That includes both some new features that are still in development, and some existing features which have never been given a proper basis in standards and are causing interop problems.

For those things it won't be possible to include them in a 2021-style metric. But we can still agree that those are areas which require priority at the standards level, and track progress on that work with the aim of including them as test-pass targets in a future iteration of this process.

foolip:
I've deleted the timeline from the RFC and the description. The current "timeline" is that we'll meet every Thursday starting Jan 13 until we have finalized the test lists, and then we'll update this RFC.

m0dulev1 (approved):

m0dulev1 (approved):

rfcs/interop_2022.md
```diff
@@ -0,0 +1,133 @@
...
+
+Interop 2022 is a public effort and benchmark around which browser vendors can collaborate to improve interoperability in areas believed to be important to web developers. 10 new focus areas have been selected, and 5 have been carried forward from Interop 2021.
+
+For each area, a set of tests have been identified, and progress is tracked with scores for each area as well as for the overall effort. There will be a dashboard for this at https://wpt.fyi/interop-2022.
+
+See the [Interop 2022 repo](https://github.com/web-platform-tests/interop-2022) for all of the proposals and discussions that led to the current set of areas.
+
+## Details
+
+10 new focus areas are part of Interop 2022, as well as 5 from Interop 2022.
```
jgraham, line 13: 5 from Interop 2021

rfcs/interop_2022.md
```diff
@@ -0,0 +1,133 @@
...
+
+Scoring in detail:
+
+- Every test is scored between 0 and 1. For tests with subtests, the score is the number of passes divided by the number of subtests. This avoids tests with many subtests being given much larger weight than for example reftests.
+- Every area is scored between 0 and 1, adding up the scores and dividing by the number of tests.
+- The overall score is the sum of area scores divided by the number of areas (15).
+
+Scores are presented as percentages.
+
+TODO: decide on floating point vs. integer vs. rational numbers
```
jgraham, line 93: I don't think this kind of implementation detail needs to be in the RFC.

rfcs/interop_2022.md
```diff
@@ -0,0 +1,133 @@
...
+
+A public dashboard tracking the metrics over time will be available at https://wpt.fyi/interop-2022. The dashboard will include at least:
+
+- An explanation of the scope, which is the 15 focus areas, not the whole web platform.
+- A graph of the metric over time for at least Chrome, Firefox and Safari, which can show both the overall Interop 2022 metric, the Interop 2021 metric which makes up 1/3 of the score, and each of the 15 focus areas individually.
+- A brief description of each focus area, with links to specs, tests on wpt.fyi, and documentation on MDN.
+
+Additional features might be added, see https://github.com/web-platform-tests/interop-2022/issues/45 for discussion.
+
+### How to nominate an area
```
jgraham, line 105: This feels pretty weird. I guess we should have had an RFC for the process and an RFC for the outcome. But I think this section should be rewritten in the past tense: "The process for nominating areas was…", since in theory one might judge this RFC by whether one thought a reasonable process had been used.

rfcs/interop_2022.md
foolip: Sure, I'll rewrite it all to a 0-100 score.

rfcs/interop_2022.md
foolip: I didn't read this comment until now, but I deleted this bit yesterday.

foolip:
I everyone! After many meetings and lots of work by many people, this RFC is now ready for review. Per [our process](https://github.com/web-platform-tests/rfcs#the-rfc-process), the earliest it could be accepted is in one week, but we're expecting discussion and going with a two week comment period. As always, this could be extended based on feedback.

I will also create a new issue in the repo for repo watchers who would otherwise miss this comment.

rfcs/interop_2022.md
```diff
@@ -0,0 +1,127 @@
...
+- [`interop-2021-transforms`](https://wpt.fyi/results/?label=master&label=experimental&product=chrome&product=firefox&product=safari&aligned&q=label%3Ainterop-2021-transforms)
+
+[Query for `interop-2021-*`](https://wpt.fyi/results/?label=master&label=experimental&product=chrome&product=firefox&product=safari&aligned&q=label%3Ainterop-2021-aspect-ratio%20or%20label%3Ainterop-2021-flexbox%20or%20label%3Ainterop-2021-grid%20or%20label%3Ainterop-2021-position-sticky%20or%20label%3Ainterop-2021-transforms)
+
+### Updating the tests
+
+No test suite is perfect, and there will likely be a need to add, remove or rewrite tests during the course of the year. To suggest such non-trivial changes that would benefit from review, file an issue on the [interop-2022 repo](https://github.com/web-platform-tests/interop-2022). The [interop-2022 team](https://github.com/orgs/web-platform-tests/teams/interop-2022) will discuss at least monthly, and make decisions based on consensus. Anyone who would like to contribute to the Interop 2022 effort is welcome to join this team.
```
jgraham, line 83: For the final sentence I suggest instead: "The core team will nominate the initial partcipants to the Interop-2022 team; thereafter the team will be able to set its own criteria for membership and participation, subject to the wpt code of conduct."

Practically I think the effect will be similar, but I think it makes sense to explicitly delegate responibility for team membership to the team itself, rather than either making it a free-for-all (which assumes no possibility of bad actors), or requiring decisions to go through the core team.

rfcs/interop_2022.md
```diff
@@ -0,0 +1,138 @@
...
+
+Proposal: [interop-2022/#13](https://github.com/web-platform-tests/interop-2022/issues/13)  
+Tests: [`interop-2022-text`](https://wpt.fyi/results/?label=master&label=experimental&product=chrome&product=firefox&product=safari&aligned&q=label%3Ainterop-2022-text)
+
+### New Viewport Units
+
+Proposal: [interop-2022/#4](https://github.com/web-platform-tests/interop-2022/issues/4)  
+Tests: [`interop-2022-viewport`](https://wpt.fyi/results/?label=master&label=experimental&product=chrome&product=firefox&product=safari&aligned&q=label%3Ainterop-2022-viewport)
+
+See also the [Viewport Investigation project](https://github.com/web-platform-tests/interop-2022/issues/41) which is part of the Interop 2022 effort, but not part of the metric.
```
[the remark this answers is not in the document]
foolip: Good catch, fixed by dropping the untruth.

rfcs/interop_2022.md
```diff
@@ -0,0 +1,138 @@
+# RFC 99: Interop 2022
+
+## Summary
+
+Interop 2022 is a public effort and benchmark around which browser vendors can collaborate to improve interoperability in areas believed to be important to web developers. 10 new focus areas have been selected, and 5 have been carried forward from Interop 2021.
```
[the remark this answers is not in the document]
foolip: Looks good, I took your suggestion verbatim modulo typo fix.

rfcs/interop_2022.md
foolip: Oops, I never updated according to this suggestion. Now done!

rfcs/interop_2022.md
```diff
@@ -0,0 +1,141 @@
...
+
+- [Viewport Measurement](https://github.com/web-platform-tests/interop-2022/issues/4)
+- [Editing, contentEditable and execCommand](https://github.com/web-platform-tests/interop-2022/issues/17)
+- [Pointer Events](https://github.com/web-platform-tests/interop-2022/issues/15) + [MouseEvent.layerX/layerY](https://github.com/web-platform-tests/interop-2022/issues/37)
+
+There will be a group working on each area and reporting progress over the course of the year.
+
+### Metrics
+
+Scores (between 0% and 100%) will be computed from test suite pass rates for each focus area and the effort as a whole. 90% of the total score is determined by the test results of the 15 focus areas. All areas are given equal weight, meaning each contributes 6% to the total score. 10% of the total score is determined by the progress in the 3 areas of investigation. Each of the 3 investigation efforts are given equal weight, meaning each contributes 3⅓% to the total score.
```
jgraham, line 100: Could we make it clear here that the test pass rate scores are implementation-specific, but the investigate scores are common across all implementations.

foolip: Done in the "scoring in detail" section now.

jgraham (approved):

foolip:
I consider this accepted now.

foolip:
I've filed https://github.com/web-platform-tests/rfcs/pull/106 about moving the metrics generation code into the web-platform-tests org. I think this is important to make the governance/ownership of the Interop 2022 metrics clearer.

PR #106, opened 2022-02-18
RFC 106: Move wpt.fyi metrics generation into the WPT org

foolip:
[Rendered](https://github.com/web-platform-tests/rfcs/blob/metrics/rfcs/wpt-results-analysis.md)


rfcs/wpt-results-analysis.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Details
+
+The following metrics on wpt.fyi are generated by wpt-results-analysis:
+
+- Browser Specific Failures
+- Compat 2021
+- Interop 2022
+
+wpt-results-analysis depends on wpt-results, a highly compressed form of all full wpt.fyi runs ever, currently ~52k stored in a ~385M Git repository. This makes it possible to load thousands (or all) runs into memory at the same time, speeding up metrics generation.
```
gsnedders, line 15: Can we rename `wpt-results`, as I think we don't want people thinking it's a canonical data-source? I'm much happier with making everyone treat wpt.fyi's API as the canonical source, and treat wpt-results as an internal implementation detail of result-analysis. Maybe `wpt-results-analysis-cache`?

rfcs/wpt-results-analysis.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Risks
+
+wpt-results-analysis depends on nodegit, which only works on Node.js 14.
+
+## Alternatives
+
+We could merge the metrics generation code into the wpt.fyi repository, but the wpt-results would still need to be standalone.
```
gsnedders, line 23: This would put us in an odd state whereby we had wpt.fyi generating CSV files, saving them in its own repo, then wpt.fyi pulling those CSV files, right?

In a sense, having wpt.fyi owning this data makes a lot of sense (it is, ultimately, just another projection of data it already owns), but doing it via CSV files in the repo feels a bit odd.

foolip: I'd like to upload the data files to a storage bucket instead. That's complicated because of GCP rules around exporting service account keys, but we can probably make it work some way or another. It's weird that wpt.fyi fetching data from https://raw.githubusercontent.com/...

rfcs/wpt-results-analysis.md
foolip: `wpt-results-analysis-cache` sounds good to me.

Ideally wpt.fyi itself would maintain a representation of results that can be efficiently queried, but that's would be a large undertaking.

Ms2ger (approved):
Seems like a good idea

foolip:
More than two weeks have passed, so I consider this RFC accepted. When it comes to implementing this, I wanted to name it just "results-analysis", but discovered we already have https://github.com/web-platform-tests/results-analysis as an archived repo. So I'd like to:
- Unarchive that repo
- Copy over https://github.com/Ecosystem-Infra/wpt-results-analysis to that repo
- Move https://github.com/Ecosystem-Infra/wpt-results to https://github.com/web-platform-tests/wpt-results-cache
- Archive https://github.com/Ecosystem-Infra/wpt-results-analysis with an updated README

An upside of this is that it'll make the transition a bit easier, since both repos will exist, so switching GitHub pages can be done much more easily.

foolip:
This is done now:
- https://github.com/web-platform-tests/results-analysis-cache is moved and has new runs pushed after the move
- https://github.com/web-platform-tests/results-analysis is replaced by the contents of https://github.com/Ecosystem-Infra/wpt-results-analysis. Open issues and PRs were recreated, and the old repo archived.
- https://github.com/web-platform-tests/results-analysis/commit/b3c55634e00319dc09aef32410e62309f890c40e updates the metrics from the new location
- https://github.com/web-platform-tests/wpt.fyi/pull/2799 makes the necessary changes to wpt.fyi

Because the old repo still exists, as a side effect, the metrics on wpt.fyi will not change until https://github.com/web-platform-tests/wpt.fyi/pull/2799 is merged and deployed.

gsnedders:
> More than two weeks have passed, so I consider this RFC accepted. When it comes to implementing this, I wanted to name it just "results-analysis", but discovered we already have https://github.com/web-platform-tests/results-analysis as an archived repo. So I'd like to:
> 
> * Unarchive that repo
> * Copy over https://github.com/Ecosystem-Infra/wpt-results-analysis to that repo
> * Move https://github.com/Ecosystem-Infra/wpt-results to https://github.com/web-platform-tests/wpt-results-cache
> * Archive https://github.com/Ecosystem-Infra/wpt-results-analysis with an updated README
> 
> An upside of this is that it'll make the transition a bit easier, since both repos will exist, so switching GitHub pages can be done much more easily.

It feels slightly unfortunate that it didn't just get moved, thereby meaning all the old issues/PRs are in this old archived repo and links in commit messages don't work. The only upside given here is that it mitigates the GitHub Pages breakage from this, which could equally have been mitigated in other ways (e.g., reuploading a mirror to Ecosystem-Infra to maintain the Pages URLs there, at least in the short term).

foolip:
> The only upside given here is that it mitigates the GitHub Pages breakage from this, which could equally have been mitigated in other ways (e.g., reuploading a mirror to Ecosystem-Infra to maintain the Pages URLs there, at least in the short term).

Ah, I didn't consider that option, that probably would have worked, with at most a small amount of downtime. I agree that would have been better.

PR #107, opened 2022-03-18
RFC 107: Add support for Shadow Realms as a Javascript global

jjgriego:


Ms2ger (approved):
I've been working alongside Joseph on this, and his proposal makes sense to me.

PR #108, opened 2022-03-28
RFC 108: Add testdriver.js support for WebDriver "Get All Cookies" and "Get Named Cookie" commands

johannhof:


johannhof:
cc @miketaylr @briannagoldstein

rfcs/testdriver_get_cookie_methods.md
```diff
@@ -0,0 +1,40 @@
...
+`test_driver.get_cookies()` instead.
+
+This has a dependence on agreeing to the right strategy for [cookie serialization
+in WebDriver](https://github.com/w3c/webdriver/issues/1562), but there seems to
+be agreement on the issue that preserving raw bytes without intermediate UTF-8
+parsing seems like the right way to go.
+
+## Risks
+
+As mentioned, cookie serialization isn't fully defined (tough loosely agreed
```
gsnedders, line 29:
```suggestion
As mentioned, cookie serialization isn't fully defined (though loosely agreed
```

gsnedders (commented):
The majority of the risk here comes from https://github.com/w3c/webdriver/issues/1562, but I don't think we need to block ourselves on that. Even without that decided, we should be able to get the common-case working.

foolip (approved):
Discussed with @jgraham and @Ms2ger in a meeting just now.

This seems like a good addition, and OK to implement without first resolving https://github.com/w3c/webdriver/issues/1562. We should follow whatever the WebDriver WG decides on that in the future.

arichiv:
@johannhof do you intend to implement this? If not, I have capacity and could probably do it for M107. I see that the serialization issue for non UTF-8 isn't solved, but if that can be set aside I'd just send JSON with the same values as [Selenium](https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/Cookie.java#L27).

johannhof:
@arichiv I had this somewhere down my list of things to get to but if you have the cycles I'd be grateful for the help!

arichiv:
Awesome! I have a few tests where this would be handy (verifying the way max age is now capped).

arichiv:
Sorry for the delay, CL up for review here: https://chromium-review.googlesource.com/c/chromium/src/+/3891100

PR #111, opened 2022-04-05
Add RFC 111 for polyfill testing.

flackr:


rfcs/polyfill_testing.md
```diff
@@ -0,0 +1,53 @@
...
+   feature](https://github.com/w3c/webdriver-bidi/issues/65) once it is
+   available.
+
+## Risks
+
+Depending on the method used there are various risks.
+
+### Rewriting the response
+
+Rewriting the response is a fairly minimal change to the infrastructure, and we
```
jgraham, line 37: My concern with this is that injecting the polyfill into all HTML resource loads could lead to tests breaking unexpectedly when they rely on the particular structure of the document e.g. if they assume a certain number of `script` nodes, or that a document has an empty DOM, or similar.

This means that even though there isn't an explicit dependence on the polyfill in the tests, tests do have to be written with a polyfill in mind. I'm not sure this kind of implicit coupling is better than an explicit one, even though it's easier to remove later on.

rfcs/polyfill_testing.md
```diff
@@ -0,0 +1,53 @@
...
+identical to the test files in the repo.
+
+### Adding a canonical resource which includes the polyfill
+
+Using the approach of having a specific resource to include will result in a
+mixture of tests which support running with polyfills and tests which don't.
+If we later adopt a mechanism which doesn't require explicitly including them
+these artifacts may take some time to remove.
+
+### Bootstrap scripts feature
```
jgraham, line 49: Possibly mad idea: could this work with a Web Extension? Can one write a cross-browser web extension that injects the polyfill before content is loaded? We have some support for automatically loading extensions (at least in Firefox), so it's not impossible to imagine it working from a wptrunner point of view.

flackr: Having an extension inject the polyfill would also require it injecting a script tag which I suspect would have a lot of the same visibility concerns. I also don't like losing the ability to locally test the polyfill against wpt serve, but I guess that's a question of what the end goal of this support is.

flackr: The other downside is extensions don't have any path to mobile testing support for a while.

jgraham:
Based on the matrix discussion, I think this is a reasonable thing to land and injecting an inline script is probably the best approach for now. I hope we can update the RFC with some of the context we discussed on matrix about the various tradeoffs, observability of the injection, etc. so that it's clear to other readers why this is the right design choice.

flackr:
RFC is updated, let me know if there's anything else you'd like to see here. I also added a short conclusion to explain the decision and rationale.

jgraham (approved):

flackr:
FYI, I don't have write access to be able to merge this.

jgraham:
I would ideally like to see some approval from other members of the @web-platform-tests/wpt-core-team The last update was not yet a week ago, so per the process we do need to allow a little additional time for review.

zcorpan:
Please move the conclusion and the winning proposal into the "Details" section, and have the other alternative solutions in an "Alternatives considered" section.

I think there are some details missing. Which responses will get the rewrite? How do you opt in? Where is the script injected? Is the injection a classic script or a module script? (Module scripts have `defer` semantics by default, and tests may complete before deferred scripts are run.)

flackr:
Thanks for taking a look!

> Please move the conclusion and the winning proposal into the "Details" section, and have the other alternative solutions in an "Alternatives considered" section.

Done.

> I think there are some details missing. Which responses will get the rewrite? How do you opt in? Where is the script injected? Is the injection a classic script or a module script? (Module scripts have `defer` semantics by default, and tests may complete before deferred scripts are run.)

I've now covered all of these in the details section. Let me know if there's anything else you'd like to see.

rfcs/polyfill_testing.md
```diff
@@ -17,41 +17,67 @@ when developers are determining whether they can rely on it, just as they would
...
 
-There are a few methods by which we could support this but each of them
-ultimately is some mechanism for running the polyfill script before running
-tests.
+In order to support testing a polyfill, a runtime argument is added which
+when used specifies a single local Javascript file to be injected into
+responses. The contents of the file will be added as a classic script (to ensure
+the polyfill runs before any tests) automatically to responses with a
+`text/html` MIME type after the `doctype` but before the first tag in the
+document.
```
zcorpan, line 25: Injecting after the doctype will *work* in most cases but relies on error handling if the test uses `<html>` and/or `<head>` start tags. Further, some tests omit the doctype.

I think the following would work and avoid invalid HTML:

1. If there is a match for `/<head(\s+[^>]*)?>/i`, then inject after the first match.
2. Otherwise, if there is a match for `/<html(\s+[^>]*)?>/i`, then inject after the first match.
3. Otherwise, if there is a match for `/<!doctype[^>]*>/i`, then inject after the first match.
4. Otherwise, inject at the start of the response.

In theory, the regexps could match incorrectly if a test does something like `<head data-x=">">`. However, it looks like there are no such tests currently, and there's no reason to use it as the `>` can be escaped as `&gt;` and HTML parser tests are in a [custom format](https://github.com/html5lib/html5lib-tests/tree/master/tree-construction).

```
$ grep -rE "<h(tml|ead)\s+[^>']*'[^'>]+>" .
$ grep -rE '<h(tml|ead)\s+[^>"]*"[^">]+>' .
```

flackr: > Further, some tests omit the doctype.

A missing doctype is supported by https://github.com/web-platform-tests/wpt/pull/33338 but I will call it out explicitly in the rfc.

> 
> I think the following would work and avoid invalid HTML:
> 
> 1. If there is a match for `/<head(\s+[^>]*)?>/i`, then inject after the first match.
> 2. Otherwise, if there is a match for `/<html(\s+[^>]*)?>/i`, then inject after the first match.
> 3. Otherwise, if there is a match for `/<!doctype[^>]*>/i`, then inject after the first match.
> 4. Otherwise, inject at the start of the response.
> 
> In theory, the regexps could match incorrectly if a test does something like `<head data-x=">">`. However, it looks like there are no such tests currently, and there's no reason to use it as the `>` can be escaped as `&gt;` and HTML parser tests are in a [custom format](https://github.com/html5lib/html5lib-tests/tree/master/tree-construction).
> 
> ```
> $ grep -rE "<h(tml|ead)\s+[^>']*'[^'>]+>" .
> $ grep -rE '<h(tml|ead)\s+[^>"]*"[^">]+>' .
> ```

The PR for this already uses html5lib to parse the dom. I've updated it to keep parsing past the opening html and head tags if they exist over at https://github.com/web-platform-tests/wpt/pull/33338.

I hope this addresses your concerns, thanks for raising them.

rfcs/polyfill_testing.md
```diff
@@ -0,0 +1,151 @@
...
+  possible mitigations.
+* Developers would not be able to debug tests served from running `wpt serve` by
+  simply loading them in the browser. They could however install the extension
+  which injected the polyfill.
+* Mobile browsers have no current support for these kinds of extensions.
+
+## Conclusion
+
+Rewriting the responses from the wpt server is a solution that is available now,
+supports all existing tests, and supports testing both via wpt run and by
```
foolip, line 146: There are still some tests that don't work right, in particular XHTML and SVG? That seems fine, but worth spelling out still.

foolip (approved):
Looks like a good tradeoff to me!

rfcs/polyfill_testing.md
flackr: Injecting on xhtml mime types seems straightforward enough, I added it to the polyfill. I've updated this to call out specifically html and xhtml tests.

rfcs/polyfill_testing.md
```diff
@@ -0,0 +1,151 @@
...
+browser.
+
+In order to support testing a polyfill, a runtime argument is added which
+when used specifies a single local Javascript file to be injected into
+responses. The contents of the file will be added as a classic script (to ensure
+the polyfill runs before any tests) automatically to responses with a
+`text/html` MIME type after the `doctype` and `html` and `head` opening tags
+(if they exist) but before any other tags in the document.
+
+In order to reduce visible effects on running tests, the script is inlined (to
```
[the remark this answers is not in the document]
flackr: Thanks! Also thanks to @jgraham for bringing up these potential side effects which lead to coming up with these mitigations 😃

rfcs/polyfill_testing.md
foolip: Cool, thanks!

rfcs/polyfill_testing.md
```diff
@@ -0,0 +1,150 @@
...
+  possible mitigations.
+* Developers would not be able to debug tests served from running `wpt serve` by
+  simply loading them in the browser. They could however install the extension
+  which injected the polyfill.
+* Mobile browsers have no current support for these kinds of extensions.
+
+## Conclusion
+
+Rewriting the responses from the wpt server is a solution that is available now,
+supports html and xhtml tests, and supports testing both via wpt run and by
```
foolip, line 145:
```suggestion
supports HTML and XHTML tests, and supports testing both via `wpt run` and by
```

rfcs/polyfill_testing.md
```diff
@@ -0,0 +1,150 @@
...
+
+In order to reduce visible effects on running tests, the script is inlined (to
+prevent an external resource load which would show up in the resource timing
+API) and the script will remove itself from the DOM to ensure that tests which
+assume the shape of the DOM is unchanged will still function correctly (e.g.
+`document.getElementsByTagName('script')[0]` will still return the first script
+in the original test).
+
+E.g. when running `wpt run --inject-script=script.js [browser] [test]` or
+`wpt serve --inject-script=script.js` and loading the following html page:
```
foolip, line 36:
```suggestion
`wpt serve --inject-script=script.js` and loading the following HTML page:
```

foolip (approved):
A few capitalization suggestion on final (?) review.

rfcs/polyfill_testing.md
zcorpan: Sounds good!

rfcs/polyfill_testing.md
```diff
@@ -0,0 +1,150 @@
...
+to know which tests the polyfill passes or doesn't pass is useful information
+when developers are determining whether they can rely on it, just as they would
+use this information to know whether they could rely on the feature in the
+browser.
+
+In order to support testing a polyfill, a runtime argument is added which
+when used specifies a single local Javascript file to be injected into
+responses. The contents of the file will be added as a classic script (to ensure
+the polyfill runs before any tests) automatically to responses with a
+`text/html` or `application/xhtml+xml` MIME type after the `doctype` and `html`
```
zcorpan, line 24: If I understand correctly, the PR uses html5lib also for `application/xhtml+xml` responses. This will work if the test is sufficiently HTML-syntax compatible, e.g. doesn't use namespace prefixes. But it can also break the test.

The script that is injected can't use `<` or `&`. This could be fixed by replacing those characters with `&lt;` and `&amp;` when the response type is `application/xhtml+xml`, or always injecting `/*<![CDATA[*/` and `/*]]>*/` (but then the script can't use `]]>`).

zcorpan (approved):
Approving, % being more careful for `application/xhtml+xml` tests

rfcs/polyfill_testing.md
flackr: Done.

rfcs/polyfill_testing.md
foolip: Leaving this out for now sounds good to me!

foolip:
The next step will be review of https://github.com/web-platform-tests/wpt/pull/33338, which implements this RFC.

PR #113, opened 2022-05-03
Remove Origin Policy infrastructure

foolip:


foolip:
@domenic, as promised :)

jgraham (approved):

gsnedders:
While I don't have any objection to doing this, I do wonder if this is really worth the churn? There's risk here, and it's not clear there's much reward.

Ms2ger (approved):
No strong feelings here.

rfcs/origin_policy_revert.md
```diff
@@ -0,0 +1,13 @@
...
+
+Support for testing [Origin Policy](https://wicg.github.io/origin-policy/) was introduced in [RFC #44](origin_policy.md) + [wpt PR #21705](https://github.com/web-platform-tests/wpt/pull/21705). Revert this code, removing the 99 op* subdomains, for example op4.web-platform.test.
+
+## Details
+
+Origin Policy is no longer being pursued. The tests using this infrastructure were removed in [wpt PR #33875](https://github.com/web-platform-tests/wpt/pull/33875).
+
+## Risks
+
+There could be tests using the subdomains for something other than testing Origin Policy. To catch any such cases, all instances of "op" followed by a digit will be inspected, and full runs of wpt with the infrastructure removed will be compared to the current state.
```
zcorpan, line 13: It looks like there aren't any:
```
% git grep -El "[^a-zA-Z]op[0-9]+(\.|\])"
tools/certs/web-platform.test.pem
```

zcorpan (approved):

foolip:
> While I don't have any objection to doing this, I do wonder if this is really worth the churn? There's risk here, and it's not clear there's much reward.

Do you think the risk mitigation suggested is enough? On the reward side, since the output of `./wpt make-hosts-file` is something most contributors will see while setting thing up, I think shrinking it from 260 to 62 lines is pretty nice, gives you less of a "WTF?" feeling and can be copied by hand into your hosts file if you don't want to just append it.

foolip:
Discussed with @jgraham and @Ms2ger in the new RFCs/infra sync. We decided to go ahead with this, there's enough support and @gsnedders didn't quite object, but did wonder about the risk.

foolip:
I've filed https://github.com/web-platform-tests/wpt/issues/34707 to implement this.

PR #116, opened 2022-07-05
RFC 116: Interop 2023

foolip:
[Rendered](https://github.com/web-platform-tests/rfcs/blob/interop_2023/rfcs/interop_2023.md)

### DIFF
--- a/rfcs/interop_2023.md
+++ b/rfcs/interop_2023.md
@@ -30,7 +30,7 @@ During 2023:
   - Complete announcement plans. Prepare dashboard for launch.
 - January 23 to January 27: Launch.
 
-If necessary, additional changes to the test lists can be made throughout the year with the consensus of the team, see [updating the tests](#updating-the-tests).
+If necessary, additional changes to the test lists can be made throughout the year with the consensus of the team; see [updating the tests](#updating-the-tests).
 
 ### Governance
 
--- a/rfcs/interop_2023.md
+++ b/rfcs/interop_2023.md
@@ -4,7 +4,7 @@
 
 Interop 2023 is an effort to improve interoperability in areas believed to be important to web developers and users. This is the next iteration of [Interop 2022](./interop_2022.md), an effort which is still ongoing at the time of this RFC.
 
-The focus areas will be selected by a proposals period followed by a proposals review and selection process. The [Interop team](https://github.com/orgs/web-platform-tests/teams/interop-2022) will manage this process and make decisions based on consensus, see [governance](#governance).
+The focus areas will be selected by a proposals period followed by a proposals-review-and-selection process. The [Interop team](https://github.com/orgs/web-platform-tests/teams/interop-2022) will manage this process and make decisions based on consensus; see [governance](#governance).
 
 There will be a dashboard at https://wpt.fyi/interop-2023.
 
--- a/rfcs/interop_2023.md
+++ b/rfcs/interop_2023.md
@@ -36,7 +36,7 @@ If necessary, additional changes to the test lists can be made throughout the ye
 
 The [Interop team](https://github.com/orgs/web-platform-tests/teams/interop) will manage the overall Interop 2023 effort. The team will be formed by renaming the [’22 team](https://github.com/orgs/web-platform-tests/teams/interop-2022).
 
-The team makes decisions based on consensus, which is defined as support from at least two participating organizations and no opposition.
+The team makes decisions based on consensus, which is defined as support from at least two participating organizations, and no opposition.
 
 The team will meet regularly, with agenda issues created in the repo at least 48 hours in advance, and meeting minutes posted in the issue. Decisions can be made both in meetings and in the repository, as long as consensus can be established.
 
--- a/rfcs/interop_2023.md
+++ b/rfcs/interop_2023.md
@@ -40,8 +40,6 @@ The team makes decisions based on consensus, which is defined as support from at
 
 The Interop 2023 effort is subject to the [WPT code of conduct](https://github.com/web-platform-tests/wpt/blob/master/CODE_OF_CONDUCT.md).
 
-The [WPT core team](https://github.com/orgs/web-platform-tests/teams/wpt-core-team/) is not expected to be involved beyond this RFC, but could in exceptional cases intervene via the RFC process to amend the governance.
-
 ### Proposals
 
 During the proposals period (September 15 to October 15), proposals are submitted as issues in the [Interop 2023 repo](https://github.com/web-platform-tests/interop-2023). A proposal should be as specific as reasonably possible, but multiple proposals may be grouped later in the selection process.
--- a/rfcs/interop_2023.md
+++ b/rfcs/interop_2023.md
@@ -2,13 +2,15 @@
 
 ## Summary
 
-Interop 2023 is an effort to improve interoperability in areas believed to be important to web developers and users. This is the next iteration of [Interop 2022](https://github.com/web-platform-tests/interop-2022), an effort which is still ongoing at the time of this RFC.
+Interop 2023 is an effort to improve interoperability in areas believed to be important to web developers and users. This is the next iteration of [Interop 2022](./interop_2022.md), an effort which is still ongoing at the time of this RFC.
 
-The focus areas will be selected by a proposals period followed by a proposals review and selection process. The [Interop 2023 team](https://github.com/orgs/web-platform-tests/teams/interop-2023) will manage this process and make decisions based on consensus, see [governance](#governance).
+The focus areas will be selected by a proposals period followed by a proposals review and selection process. The [Interop team](https://github.com/orgs/web-platform-tests/teams/interop-2022) will manage this process and make decisions based on consensus, see [governance](#governance).
 
 There will be a dashboard at https://wpt.fyi/interop-2023.
 
 ## Details
+
+The Interop 2022 [repo](https://github.com/web-platform-tests/interop-2022) and [team](https://github.com/orgs/web-platform-tests/teams/interop-2022) will be renamed to omit the year and used for this effort as well.
 
 ### Timeline
 
@@ -32,7 +34,7 @@ If necessary, additional changes to the test lists can be made throughout the ye
 
 ### Governance
 
-The [Interop 2023 team](https://github.com/orgs/web-platform-tests/teams/interop-2023) will manage the overall Interop 2023 effort. The team will be formed with the same initial membership as the [’22 team](https://github.com/orgs/web-platform-tests/teams/interop-2022).
+The [Interop team](https://github.com/orgs/web-platform-tests/teams/interop) will manage the overall Interop 2023 effort. The team will be formed by renaming the [’22 team](https://github.com/orgs/web-platform-tests/teams/interop-2022).
 
 The team will meet regularly, with agenda issues created in the repo at least 48 hours in advance, and meeting minutes posted in the issue.
 
@@ -42,7 +44,7 @@ The Interop 2023 effort is subject to the [WPT code of conduct](https://github.c
 
 ### Proposals
 
-During the proposals period (September 15 to October 15), proposals are submitted as issues in the [Interop 2023 repo](https://github.com/web-platform-tests/interop-2023). A proposal should be as specific as reasonably possible, but multiple proposals may be grouped later in the selection process.
+During the proposals period (September 15 to October 15), proposals are submitted as issues in the [Interop repo](https://github.com/web-platform-tests/interop). A proposal should be as specific as reasonably possible, but multiple proposals may be grouped later in the selection process.
 
 There are two kinds of proposals: focus areas where progress is measured using test results, and investigation efforts where progress is measuring against a task list. For both kinds of proposals, finishing them should meaningfully advance the state of interoperability for the web platform.
 
@@ -83,11 +85,11 @@ The purpose of the second round of grouping and optionally dropping proposals is
 
 ### Updating the tests
 
-No test suite is perfect, and there will likely be a need to add, remove or rewrite tests during the course of the year. To suggest changes, file an issue on the [Interop 2023 repo](https://github.com/web-platform-tests/interop-2023). The team will make decisions based on consensus, see [governance](#governance).
+No test suite is perfect, and there will likely be a need to add, remove or rewrite tests during the course of the year. To suggest changes, file an issue on the [Interop repo](https://github.com/web-platform-tests/interop). The team will make decisions based on consensus, see [governance](#governance).
 
 ### Metrics
 
-Per-area and overall metrics will be defined in a similar fashion to [Interop 2022](https://github.com/web-platform-tests/rfcs/blob/master/rfcs/interop_2022.md#metrics). The number of areas and their contribution to the total score will be decided in the proposal selection phase.
+Per-area and overall metrics will be defined in a similar fashion to [Interop 2022](./interop_2022.md#metrics). The number of areas and their contribution to the total score will be decided in the proposal selection phase.
 
 ### Dashboard
 
--- a/rfcs/interop_2023.md
+++ b/rfcs/interop_2023.md
@@ -10,7 +10,7 @@ There will be a dashboard at https://wpt.fyi/interop-2023.
 
 ## Details
 
-The Interop 2022 [repo](https://github.com/web-platform-tests/interop-2022) and [team](https://github.com/orgs/web-platform-tests/teams/interop-2022) will be renamed to omit the year and used for this effort as well.
+The Interop 2022 [repo](https://github.com/web-platform-tests/interop-2022) and [team](https://github.com/orgs/web-platform-tests/teams/interop-2022) will be renamed to omit the year and used for this effort as well. Proposals from different years will be distinguished by labels or milestones.
 
 ### Timeline
 
@@ -36,17 +36,19 @@ If necessary, additional changes to the test lists can be made throughout the ye
 
 The [Interop team](https://github.com/orgs/web-platform-tests/teams/interop) will manage the overall Interop 2023 effort. The team will be formed by renaming the [’22 team](https://github.com/orgs/web-platform-tests/teams/interop-2022).
 
-The team will meet regularly, with agenda issues created in the repo at least 48 hours in advance, and meeting minutes posted in the issue.
-
 The team makes decisions based on consensus, which is defined as support from at least two participating organizations and no opposition.
+
+The team will meet regularly, with agenda issues created in the repo at least 48 hours in advance, and meeting minutes posted in the issue. Decisions can be made both in meetings and in the repository, as long as consensus can be established.
 
 The Interop 2023 effort is subject to the [WPT code of conduct](https://github.com/web-platform-tests/wpt/blob/master/CODE_OF_CONDUCT.md).
 
 ### Proposals
 
-During the proposals period (September 15 to October 15), proposals are submitted as issues in the [Interop repo](https://github.com/web-platform-tests/interop). A proposal should be as specific as reasonably possible, but multiple proposals may be grouped later in the selection process.
+During the proposals period (September 15 to October 15), proposals are submitted in the [Interop repo](https://github.com/web-platform-tests/interop). A proposal should be as specific as reasonably possible, but multiple proposals may be grouped later in the selection process.
 
-There are two kinds of proposals: focus areas where progress is measured using test results, and investigation efforts where progress is measuring against a task list. For both kinds of proposals, finishing them should meaningfully advance the state of interoperability for the web platform.
+To make proposal scope unambiguous, accepted proposals should be fully documented in the repository. To this end, and to make proposal review easier, the team may require that proposals are submitted as pull requests. Whether and how discussions, issues and pull requests are used in the proposals process is left to the team to decide before the proposal period begins.
+
+There are two kinds of proposals: focus areas where progress is measured using test results, and investigation efforts where progress is measuring against a task list.
 
 #### Focus areas
 
@@ -77,8 +79,8 @@ The following information should be provided in a proposal:
 Proposal selection (November 1 to November 30) will be done using a positions process similar to [Interop 2022](https://github.com/web-platform-tests/interop-2022/issues/38) but with some changes. The process is:
 
 - Each participating organization takes one of 3 positions on each proposal: support, neutral, or oppose.
-- A proposal is accepted if there is consensus, meaning support from at least two participating organizations and no opposition.
-- The team reviews the accepted proposals and can decide to group multiple proposals, or to drop proposals. Both actions require team consensus.
+- A proposal is tentatively accepted if there is consensus, meaning support from at least two participating organizations and no opposition.
+- The team reviews the proposals and can decide to group multiple proposals, or to drop proposals. Proposals that got only neutral positions can also be included. All changes at this point require team consensus; by default the outcome of the previous step stands.
 - The final list of accepted proposals is decided by team consensus no later than November 30.
 
 The purpose of the second round of grouping and optionally dropping proposals is to avoid sub-proposals as in the Interop 2022 process. Instead, proposals will be as specific as reasonably possible, and can be grouped into themes afterwards. The expectation is that proposals will only be dropped if there is consensus that they don't make sense in light of which other proposals were accepted.
--- a/rfcs/interop_2023.md
+++ b/rfcs/interop_2023.md
@@ -36,7 +36,7 @@ The [Interop 2023 team](https://github.com/orgs/web-platform-tests/teams/interop
 
 The team will meet regularly, with agenda issues created in the repo at least 48 hours in advance, and meeting minutes posted in the issue.
 
-The team makes decisions based on consensus, which is defined as support from at least two participating organizations and no objections. Additionally, during the [proposal selection](#proposal-selection) process, the positions of implementers in each browser engine are used to determine whether a proposal will be accepted, to ensure that the proposal is on track to interoperability across browser engines.
+The team makes decisions based on consensus, which is defined as support from at least two participating organizations and no opposition.
 
 The Interop 2023 effort is subject to the [WPT code of conduct](https://github.com/web-platform-tests/wpt/blob/master/CODE_OF_CONDUCT.md).
 
@@ -75,7 +75,7 @@ The following information should be provided in a proposal:
 Proposal selection (November 1 to November 30) will be done using a positions process similar to [Interop 2022](https://github.com/web-platform-tests/interop-2022/issues/38) but with some changes. The process is:
 
 - Each participating organization takes one of 3 positions on each proposal: support, neutral, or oppose.
-- A proposal is accepted if there is implementer support coming from at least 2 browser engines, and no opposition.
+- A proposal is accepted if there is consensus, meaning support from at least two participating organizations and no opposition.
 - The team reviews the accepted proposals and can decide to group multiple proposals, or to drop proposals. Both actions require team consensus.
 - The final list of accepted proposals is decided by team consensus no later than November 30.
 
--- a/rfcs/interop_2023.md
+++ b/rfcs/interop_2023.md
@@ -44,15 +44,31 @@ The Interop 2023 effort is subject to the [WPT code of conduct](https://github.c
 
 During the proposals period (September 15 to October 15), proposals are submitted as issues in the [Interop 2023 repo](https://github.com/web-platform-tests/interop-2023). A proposal should be as specific as reasonably possible, but multiple proposals may be grouped later in the selection process.
 
-The following information should be provided:
+There are two kinds of proposals: focus areas where progress is measured using test results, and investigation efforts where progress is measuring against a task list. For both kinds of proposals, finishing them should meaningfully advance the state of interoperability for the web platform.
 
-- Standard/specification for this feature
-- Current test results on wpt.fyi
-- How widely used is this feature? For example, [use counters](https://www.chromestatus.com/metrics/feature/popularity), [HTTP Archive](https://httparchive.org/) and [State of CSS](https://2021.stateofcss.com/en-US/features/) can help.
+#### Focus areas
+
+These proposals are for existing web platform features with good a spec and test suite which can be used to improve interoperability. Progress is measured by test results on wpt.fyi.
+
+The following information should be provided in a proposal:
+
 - Is this area a problem for web developers? For example, survey data and browser bug counts can help.
 - Is this area a problem for users?
+- How widely used is this feature? For example, [use counters](https://www.chromestatus.com/metrics/feature/popularity), [HTTP Archive](https://httparchive.org/) and [State of CSS](https://2021.stateofcss.com/en-US/features/) can help.
+- Standard/specification for this feature.
+- A list of tests on wpt.fyi which will be used to score progress towards interoperability on the feature.
 
-A proposals template will be created to help with this.
+#### Investigation efforts
+
+If an area doesn't have a good spec and test suite it's not possible to use wpt.fyi test results to judge progress. Instead, some other work is needed to bring the area to state where test results can be used.
+
+The following information should be provided in a proposal:
+
+- Is this area a problem for web developers? For example, survey data and browser bug counts can help.
+- Is this area a problem for users?
+- How widely used is this feature? For example, [use counters](https://www.chromestatus.com/metrics/feature/popularity), [HTTP Archive](https://httparchive.org/) and [State of CSS](https://2021.stateofcss.com/en-US/features/) can help.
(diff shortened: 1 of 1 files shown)


rfcs/interop_2023.md
```diff
@@ -0,0 +1,84 @@
...
+- Is this area a problem for users?
+
+A proposals template will be created to help with this.
+
+### Proposal selection
+
+Proposal selection (November 1 to November 30) will be done using a positions process similar to [Interop 2022](https://github.com/web-platform-tests/interop-2022/issues/38) but with some changes. The process is:
+
+- Each participating organization takes one of 3 positions on each proposal: support, neutral, or oppose.
+- A proposal is accepted if there is implementer support coming from at least 2 browser engines, and no opposition.
```
foolip, line 64: This is formalizing what we did for Interop 2022 in https://github.com/web-platform-tests/interop-2022/issues/39#issue-1073181003

> I suggest defining consensus as support from at least two organizations contributing to different browser engines, and a lack of opposition from everyone else.

It's noteworthy that the outcome would have been the same if we only required support from one organization and no opposition. That's because we didn't have any proposals with 1 support and everyone else neutral. I think it's worth a discussion how we want to handle that case if it arises.

foolip:
This is a big RFC, so I'd like to call out some things worth looking at:

- The timeline! We should stick to it, so review it carefully!
- To avoid the proposal/sub-proposal structure, proposals should be "as specific as reasonably possible", and I've added a  grouping/dropping step. Dropping requires consensus, it's not doing the support/neutral/oppose positions over again.
- Investigation efforts aren't mentioned and "investigate" is not one of the possible positions on proposals. Should we have explicit "investigate" proposals, and what should we require for them? I think it's fair to say these got off to a slow start in 2022, can we improve on that?
- User needs are mentioned, since that is what motivated the Web Compat focus area of 2022. How to weigh developer needs, user needs, etc., is left to the participating organizations.

rfcs/interop_2023.md
```diff
@@ -0,0 +1,84 @@
+# RFC 116: Interop 2023
+
+## Summary
+
+Interop 2023 is an effort to improve interoperability in areas believed to be important to web developers and users. This is the next iteration of [Interop 2022](https://github.com/web-platform-tests/interop-2022), an effort which is still ongoing at the time of this RFC.
+
+The focus areas will be selected by a proposals period followed by a proposals review and selection process. The [Interop 2023 team](https://github.com/orgs/web-platform-tests/teams/interop-2023) will manage this process and make decisions based on consensus, see [governance](#governance).
```
gsnedders, line 7: I do vaguely wonder if it's worthwhile having a per-year repo? To me it doesn't seem super worthwhile.

rfcs/interop_2023.md
```diff
@@ -0,0 +1,84 @@
...
+
+If necessary, additional changes to the test lists can be made throughout the year with the consensus of the team, see [updating the tests](#updating-the-tests).
+
+### Governance
+
+The [Interop 2023 team](https://github.com/orgs/web-platform-tests/teams/interop-2023) will manage the overall Interop 2023 effort. The team will be formed with the same initial membership as the [’22 team](https://github.com/orgs/web-platform-tests/teams/interop-2022).
+
+The team will meet regularly, with agenda issues created in the repo at least 48 hours in advance, and meeting minutes posted in the issue.
+
+The team makes decisions based on consensus, which is defined as support from at least two participating organizations and no objections. Additionally, during the [proposal selection](#proposal-selection) process, the positions of implementers in each browser engine are used to determine whether a proposal will be accepted, to ensure that the proposal is on track to interoperability across browser engines.
```
gsnedders, line 39: Can we be consistent about whether we're using "organisations" or "implementers in each browser engine"? These aren't the same. Can we not simply rely on objections for the latter part?

rfcs/interop_2023.md
```diff
@@ -0,0 +1,84 @@
...
+
+The Interop 2023 effort is subject to the [WPT code of conduct](https://github.com/web-platform-tests/wpt/blob/master/CODE_OF_CONDUCT.md).
+
+The [WPT core team](https://github.com/orgs/web-platform-tests/teams/wpt-core-team/) is not expected to be involved beyond this RFC, but could in exceptional cases intervene via the RFC process to amend the governance.
```
gsnedders, line 43: Is this necessary? This is implicit anyway.

rfcs/interop_2023.md
```diff
@@ -0,0 +1,84 @@
...
+  - Complete announcement plans. Prepare dashboard for launch.
+- January 23 to January 27: Launch.
+
+If necessary, additional changes to the test lists can be made throughout the year with the consensus of the team, see [updating the tests](#updating-the-tests).
+
+### Governance
+
+The [Interop 2023 team](https://github.com/orgs/web-platform-tests/teams/interop-2023) will manage the overall Interop 2023 effort. The team will be formed with the same initial membership as the [’22 team](https://github.com/orgs/web-platform-tests/teams/interop-2022).
+
+The team will meet regularly, with agenda issues created in the repo at least 48 hours in advance, and meeting minutes posted in the issue.
```
gsnedders, line 37: Let's also be explicit about the team also doing work within the repo, rather than everything being done in meetings.

gsnedders (commented):
FWIW, I'm not totally sure it makes sense to actually have one of these RFCs annually at this point; I think there's broad agreement across the Core team that we should keep on doing this as long as there's cross-vendor interest.

I'd be more interested in an RFC that acts closer to a charter for the Interop team going forward, and leave it down to that team what happens (potentially with an explicit limitation where it only holds as long as it has a similar coverage of browser vendors to the Core Team).

In a sense, I suggest we focus on the governance section here, and define the scope of what we believe Interop should be.

(Thus not immediately commenting on the rest.)

rfcs/interop_2023.md
foolip: My intention was to limit the special status of browser engines and implementers only to the positions process. Do you think we should use one definition of consensus everywhere, and if so what should it be?

Some specific cases to consider, whether in the positions process or other decisions:
- Google and Microsoft support and everyone else is neutral.
- Bocoup and Igalia support and everyone else is neutral. Consider both the case where there is commitment to implement and not.
- Apple, Google, Microsoft and Mozilla support but another newly joined organization objects.

I hope dearly that we won't have to explore the limits of our process and think that in practice we'll have a conversation and resolve things without falling back to the formal definition of consensus. But in particular for the positions process I think we need to be clear, and that what we did last time worked out pretty well.

> Can we not simply rely on objections for the latter part?

Do you mean that decisions require support&times;2 and no objections, and that's it?

rfcs/interop_2023.md
foolip: We can leave it implicit, I take it that's your preference.

rfcs/interop_2023.md
foolip: Do we want a way to make decisions outside of meetings? How do we establish enough support and no objections?

rfcs/interop_2023.md
foolip: I take it the proposal is to redirect https://github.com/web-platform-tests/interop-2022 to https://github.com/web-platform-tests/interop?

foolip:
> FWIW, I'm not totally sure it makes sense to actually have one of these RFCs annually at this point; I think there's broad agreement across the Core team that we should keep on doing this as long as there's cross-vendor interest.
> 
> I'd be more interested in an RFC that acts closer to a charter for the Interop team going forward, and leave it down to that team what happens (potentially with an explicit limitation where it only holds as long as it has a similar coverage of browser vendors to the Core Team).
> 
> In a sense, I suggest we focus on the governance section here, and define the scope of what we believe Interop should be.

I like that direction if others in the core team and interop-2022 team are on board with it.

Maybe a setup like this?

- Work done in https://github.com/web-platform-tests/interop
- A `charter.md` which is created by the RFC process and can be amended by it
- A folder or markdown file per year, backfilling with [`interop_2022.md`](https://github.com/web-platform-tests/rfcs/blob/master/rfcs/interop_2022.md)

rfcs/interop_2023.md
gsnedders: > My intention was to limit the special status of browser engines and implementers only to the positions process. Do you think we should use one definition of consensus everywhere, and if so what should it be?

Yes, and I'll deal with the second clause below…

> Some specific cases to consider, whether in the positions process or other decisions:
> 
> * Google and Microsoft support and everyone else is neutral.

At first glance, that seems unproblematic to approve in general? Like, okay, there's only positive indications from a single browser engine, so this in a sense depends on whether we believe people will actively object to things? On the whole, I think we generally saw that last year (as I think you mentioned, nothing would've changed had we done this last year), at least after much discussion and it being clear how much support each thing had.

> * Bocoup and Igalia support and everyone else is neutral. Consider both the case where there is commitment to implement and not.

Again, depends on whether we expect people to actually object?

> * Apple, Google, Microsoft and Mozilla support but another newly joined organization objects.

While this might be viewed as a harder question, is this actually a harder question than Bocoup objecting (given AFAIK they do little in the way of implementation work)?

The concern in a sense is anyone acting maliciously, I suppose.

> I hope dearly that we won't have to explore the limits of our process and think that in practice we'll have a conversation and resolve things without falling back to the formal definition of consensus. But in particular for the positions process I think we need to be clear, and that what we did last time worked out pretty well.
> 
> > Can we not simply rely on objections for the latter part?
> 
> Do you mean that decisions require support×2 and no objections, and that's it?

No, I meant just removing the second sentence.

rfcs/interop_2023.md
gsnedders: Good question. One option is to just allow rough consensus to suffice, without necessarily being explicit here. At least for test changes we've practically made the decisions in issues already, with the meetings primarily there for pinging people.

rfcs/interop_2023.md
foolip: As discussed in https://github.com/web-platform-tests/interop-2022/issues/95 this will be a separate RFC, in parallel to this one.

rfcs/interop_2023.md
foolip: I've pushed a change to just use support×2 and no opposition everywhere. That did remove the second sentence referenced here. PTAL?

rfcs/interop_2023.md
foolip: I haven't resolved this here, but I think we can define it in the coming interop team governance RFC?

foolip:
Alright, I've pushed two changes:
- Define investigation effort proposals
- Simplify consensus to use the same definition everywhere

I think that resolves all feedback, except the decision making process which I suggest we do in a coming interop team governance RFC.

Everyone please take another look.

foolip:
I still need to change this to no longer refer to a 2023 repo or team, to instead say that we'll rename the existing ones.

foolip:
Alright, those updates now made as well. I'm not aware of any more changes that need to be made, but I might have missed something. Please review :)

rfcs/interop_2023.md
```diff
@@ -0,0 +1,100 @@
...
+
+The team makes decisions based on consensus, which is defined as support from at least two participating organizations and no opposition.
+
+The Interop 2023 effort is subject to the [WPT code of conduct](https://github.com/web-platform-tests/wpt/blob/master/CODE_OF_CONDUCT.md).
+
+### Proposals
+
+During the proposals period (September 15 to October 15), proposals are submitted as issues in the [Interop repo](https://github.com/web-platform-tests/interop). A proposal should be as specific as reasonably possible, but multiple proposals may be grouped later in the selection process.
+
+There are two kinds of proposals: focus areas where progress is measured using test results, and investigation efforts where progress is measuring against a task list. For both kinds of proposals, finishing them should meaningfully advance the state of interoperability for the web platform.
```
gsnedders, line 49: One thing that came up last year—but we ultimately decided against—was making something "investigation" like for "implement infrastructure for".

Some of the things that have previously been proposed for next year require infrastructure we don't currently have, so it continues to feel like it would be reasonable?

rfcs/interop_2023.md
```diff
@@ -0,0 +1,100 @@
...
+
+#### Focus areas
+
+These proposals are for existing web platform features with good a spec and test suite which can be used to improve interoperability. Progress is measured by test results on wpt.fyi.
```
gsnedders, line 53:
```suggestion
These proposals are for existing web platform features with good a spec and automated test suite which can be used to improve interoperability. Progress is measured by test results on wpt.fyi.
```

rfcs/interop_2023.md
```diff
@@ -0,0 +1,100 @@
...
+
+The following information should be provided in a proposal:
+
+- Is this area a problem for web developers? For example, survey data and browser bug counts can help.
+- Is this area a problem for users?
+- How widely used is this feature? For example, [use counters](https://www.chromestatus.com/metrics/feature/popularity), [HTTP Archive](https://httparchive.org/) and [State of CSS](https://2021.stateofcss.com/en-US/features/) can help.
+- Standard/specification for this feature.
```
gsnedders, line 60:
```suggestion
- Specification for this feature.
```

I don't think we need to allow for both? "Standards" are (practically) a subset.

rfcs/interop_2023.md
```diff
@@ -0,0 +1,100 @@
...
+#### Investigation efforts
+
+If an area doesn't have a good spec and test suite it's not possible to use wpt.fyi test results to judge progress. Instead, some other work is needed to bring the area to state where test results can be used.
+
+The following information should be provided in a proposal:
+
+- Is this area a problem for web developers? For example, survey data and browser bug counts can help.
+- Is this area a problem for users?
+- How widely used is this feature? For example, [use counters](https://www.chromestatus.com/metrics/feature/popularity), [HTTP Archive](https://httparchive.org/) and [State of CSS](https://2021.stateofcss.com/en-US/features/) can help.
+- Standard/specification for this feature, if any.
```
gsnedders, line 72:
```suggestion
- Specification for this feature, if any.
```

rfcs/interop_2023.md
```diff
@@ -0,0 +1,100 @@
...
+- Is this area a problem for users?
+- How widely used is this feature? For example, [use counters](https://www.chromestatus.com/metrics/feature/popularity), [HTTP Archive](https://httparchive.org/) and [State of CSS](https://2021.stateofcss.com/en-US/features/) can help.
+- Standard/specification for this feature.
+- A list of tests on wpt.fyi which will be used to score progress towards interoperability on the feature.
+
+#### Investigation efforts
+
+If an area doesn't have a good spec and test suite it's not possible to use wpt.fyi test results to judge progress. Instead, some other work is needed to bring the area to state where test results can be used.
+
+The following information should be provided in a proposal:
```
gsnedders, line 67: This list is essentially duplicating the above. Do we just want to duplicate this, or move this up to the general "Proposals" section above?

rfcs/interop_2023.md
```diff
@@ -0,0 +1,100 @@
...
+
+The purpose of the second round of grouping and optionally dropping proposals is to avoid sub-proposals as in the Interop 2022 process. Instead, proposals will be as specific as reasonably possible, and can be grouped into themes afterwards. The expectation is that proposals will only be dropped if there is consensus that they don't make sense in light of which other proposals were accepted.
+
+### Updating the tests
+
+No test suite is perfect, and there will likely be a need to add, remove or rewrite tests during the course of the year. To suggest changes, file an issue on the [Interop repo](https://github.com/web-platform-tests/interop). The team will make decisions based on consensus, see [governance](#governance).
+
+### Metrics
+
+Per-area and overall metrics will be defined in a similar fashion to [Interop 2022](./interop_2022.md#metrics). The number of areas and their contribution to the total score will be decided in the proposal selection phase.
```
gsnedders, line 92: I think in hindsight some of this hasn't actually worked out too well in Interop 2022; for example with the Web Compat focus area: https://wpt.fyi/results/?label=experimental&label=master&product=chrome&product=firefox&product=safari&aligned&q=label%3Ainterop-2022-webcompat

This only has seven tests in total, four in /css/css-text-decor (for https://github.com/web-platform-tests/interop-2022/issues/23), one in /css/CSS2/floats (for https://github.com/web-platform-tests/interop-2022/issues/33), one in /dom (for https://github.com/web-platform-tests/interop-2022/issues/26), and one in /url (for https://github.com/web-platform-tests/interop-2022/issues/18).

Given this distribution, we've ended up with one of the interop bugs accounting for 4/7ths of the score.

rfcs/interop_2023.md
```diff
@@ -0,0 +1,100 @@
...
+
+The team will meet regularly, with agenda issues created in the repo at least 48 hours in advance, and meeting minutes posted in the issue.
+
+The team makes decisions based on consensus, which is defined as support from at least two participating organizations and no opposition.
+
+The Interop 2023 effort is subject to the [WPT code of conduct](https://github.com/web-platform-tests/wpt/blob/master/CODE_OF_CONDUCT.md).
+
+### Proposals
+
+During the proposals period (September 15 to October 15), proposals are submitted as issues in the [Interop repo](https://github.com/web-platform-tests/interop). A proposal should be as specific as reasonably possible, but multiple proposals may be grouped later in the selection process.
```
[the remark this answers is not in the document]
jgraham: OK, so this is likely overkill, but if the proposals were PRs instead, this would be pretty clear: the actual PR is the proposal. If you want to work with the other parties to make changes then you have that conversation in the PR issue. If you want to fork the proposal then you create a new PR.

In practice I don't want to have to judge between N slight variations on the ~same proposal, so I definitely don't want everyone to be encouraged to fork rather than collaborate.

rfcs/interop_2023.md
jgraham: I think adding test infrastructure should be allowed as a proposal. "Meaningfully advance the state of interoperability on the web platform" is already quite a vauge requirement. Maybe for the sake of clarifying the spirit of the rule, we should add some examples of what would be acceptable.

rfcs/interop_2023.md
```diff
@@ -0,0 +1,100 @@
...
+
+Interop 2023 is an effort to improve interoperability in areas believed to be important to web developers and users. This is the next iteration of [Interop 2022](./interop_2022.md), an effort which is still ongoing at the time of this RFC.
+
+The focus areas will be selected by a proposals period followed by a proposals review and selection process. The [Interop team](https://github.com/orgs/web-platform-tests/teams/interop-2022) will manage this process and make decisions based on consensus, see [governance](#governance).
+
+There will be a dashboard at https://wpt.fyi/interop-2023.
+
+## Details
+
+The Interop 2022 [repo](https://github.com/web-platform-tests/interop-2022) and [team](https://github.com/orgs/web-platform-tests/teams/interop-2022) will be renamed to omit the year and used for this effort as well.
```
jgraham, line 13: We should figure out how to handle the existing issues / proposals in this framework. Options are:
* Rename labels to include the year
* Add milestones for each year and add the issues to those milestones.

A concrete use case here is that there are non-accepted proposals from 2022 that people want to audit for inclusion in 20223.

gsnedders: > Rename labels to include the year

This seems somewhat annoying further in the future; one could easily imagine having to search for `-label:accepted:2022 -label:accepted-2023` etc. etc.

foolip: How about `proposal 2022` and `proposal 2023` but no other labels (like `accepted`) get dates?

rfcs/interop_2023.md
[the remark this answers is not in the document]
foolip: @jgraham on balance, do you think we should do proposals as PRs or not?

Or we ask that the initial comment is updated to reflect the complete proposal at each point?

Either way it's going to be a bit annoying in trivial cases where there's no actual ambiguity, and useful in case where the discussion has made things murky.

rfcs/interop_2023.md
foolip: I think test infra improvements can be good proposals and can meaningfully advance the state of interop in the same way as clarifying specs or writing tests can.

It is mentioned below, is that enough?

rfcs/interop_2023.md
foolip: I just duplicated it because it was less work, and I think still easy enough to understand this way. I don't feel strongly about it though, feel free to push an edit. (A suggestion would span many lines, I presume.)

rfcs/interop_2023.md
foolip: Do you think the RFC should say something more about the scoring than that we'll decide later?

rfcs/interop_2023.md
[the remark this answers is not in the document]
jgraham: Well I'm interested to see what others think about proposal-as-PR.

Technically, the more I think about it the more advantages it has. We can merge the actual PRs so that the repo has a list of all the accepted proposals and various metadata e.g. the wpt.fyi labels used. We can have year-dated directories to keep a history of the project without having to dig through GH issues. We could in theory have an up-to-date list of tests in each proposal and verify that matches the labels on wpt.fyi (or write tooling that will auto-sync the labels). Although all these things are possible with issues, the fact that you have to go through the GH API to access the data does make things harder.

But it does have some additional overhead (perhaps less now that you can simply use github.dev to edit inline). So if that's a blocker for people I'm happy to withdraw this suggestion.

rfcs/interop_2023.md
jgraham: It's fine as long as we're not going to end up having language-lawyer arguments about what it means for something to "meaningfully advance the state of interoperability for the web platform" :) Another way of looking at that is I'm not sure that this sentence is adding much value given the following subsections.

rfcs/interop_2023.md
```diff
@@ -0,0 +1,100 @@
...
+- How widely used is this feature? For example, [use counters](https://www.chromestatus.com/metrics/feature/popularity), [HTTP Archive](https://httparchive.org/) and [State of CSS](https://2021.stateofcss.com/en-US/features/) can help.
+- Specification for this feature, if any.
+- A list of tasks which should be completed as part of the effort. This could be producing a report for a working group, a list of spec bugs which should be resolved, or tests/infrastructure which need to be created. A completed investigation effort should unblock the feature for inclusion as a regular focus area in the future.
+
+### Proposal selection
+
+Proposal selection (November 1 to November 30) will be done using a positions process similar to [Interop 2022](https://github.com/web-platform-tests/interop-2022/issues/38) but with some changes. The process is:
+
+- Each participating organization takes one of 3 positions on each proposal: support, neutral, or oppose.
+- A proposal is accepted if there is consensus, meaning support from at least two participating organizations and no opposition.
```
jgraham, line 80: On the "are two direct supports required" question that has already been somewhat discussed, I'd suggest that we adopt a somewhat looser policy. In particular I think that "oppose" should be absolute, but after that we should have the ability to come to mutual agreement on which proposals to accept based on the total amount we want to include, and any evidence of priority between the proposals without requiring people to formally change their position. For example, if we had 20 proposals, 5 with 2+ "support", 10 with 1 support and all neutral and 5 with all neutral, we might be unhappy with just taking 5 proposals total. But there are 15 other proposals that didn't get opposition and it would seem quite reasonable to say "we'll take 5 of the 10 with one clear support / rest neutral that people are more exicited about given discussion".

rfcs/interop_2023.md
jgraham: Would it be enough to write "similar (but not necessarily identical)"?

I agree that with something where the focus area is defined from a number of sub-proposals it doesn't necessarily make sense to weigh each sub-proposal equally (I think for webcompat we planned to have some additional tests, which has made the problem worse in that case).

rfcs/interop_2023.md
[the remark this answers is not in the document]
jgraham: OK let's take proposals-as-PR as a serious suggestion (or a proposal-proposal, if you will).

rfcs/interop_2023.md
foolip: I suppose I added this to explain the intent/aspiration, and not just have a list of requirements. Ultimately the participants aren't required to give a reason at all for their position on each proposal, so I don't think we'll get to the word lawyering.

The charter would be a good place for some word lawyering though, some principles that we can all agree to.

rfcs/interop_2023.md
foolip: I'll add in "but not necessarily identical" 👍

foolip:
```suggestion
Per-area and overall metrics will be defined in a similar (but not necessarily identical) fashion to [Interop 2022](./interop_2022.md#metrics). The number of areas and their contribution to the total score will be decided in the proposal selection phase.
```

rfcs/interop_2023.md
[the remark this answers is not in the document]
foolip: I'd like us to have an in-repo record of what Interop 2023 is defined as, and doing that as PRs makes sense. My only concern would be raising the bar too much for proposals by people who don't have all of the details ready. Perhaps we accept issues, but at some point there needs to be a PR that formalizes the proposal?

rfcs/interop_2023.md
foolip: That makes sense to me. The second round allows for some of this, but as written anything that has no support in round 1 is off the table. Do you think we should make explicit that we can revive such proposals if we have newfound support? It seems not very likely to happen in practice, but reasonable to account for the possibility.

rfcs/interop_2023.md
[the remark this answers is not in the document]
jgraham: To some extent you can use whatever method you want to workshop a proposal, but you are correct that it could be worth calling out that we are happy for people to use the repo issue tracker to help scope out ideas for proposals.

rfcs/interop_2023.md
foolip: I'll add this:

> Proposals from different years will be distinguished by labels or milestones.

rfcs/interop_2023.md
foolip: I'll reorder slightly and add this:

> Decisions can be made both in meetings and in the repository, as long as consensus can be established.

rfcs/interop_2023.md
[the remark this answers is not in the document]
foolip: I've added this:

> To make proposal scope unambiguous, accepted proposals should be fully documented in the repository. To this end, and to make proposal review easier, the team may require that proposals are submitted as pull requests. Whether and how discussions, issues and pull requests are used in the proposals process is left to the team to decide before the proposal period begins.

rfcs/interop_2023.md
foolip: I'm just going to drop this bit and leave it to the charter or the call for proposals:

> For both kinds of proposals, finishing them should meaningfully advance the state of interoperability for the web platform.

rfcs/interop_2023.md
foolip: I'll leave them duplicated.

foolip:
I've addressed all the remaining feedback now. @web-platform-tests/wpt-core-team please review, for the final time?

rfcs/interop_2023.md
```diff
@@ -0,0 +1,102 @@
+# RFC 116: Interop 2023
+
+## Summary
+
+Interop 2023 is an effort to improve interoperability in areas believed to be important to web developers and users. This is the next iteration of [Interop 2022](./interop_2022.md), an effort which is still ongoing at the time of this RFC.
+
+The focus areas will be selected by a proposals period followed by a proposals review and selection process. The [Interop team](https://github.com/orgs/web-platform-tests/teams/interop-2022) will manage this process and make decisions based on consensus, see [governance](#governance).
```
sideshowbarker, line 7:
```suggestion
The focus areas will be selected by a proposals period followed by a proposals-review-and-selection process. The [Interop team](https://github.com/orgs/web-platform-tests/teams/interop-2022) will manage this process and make decisions based on consensus; see [governance](#governance).
```

rfcs/interop_2023.md
```diff
@@ -0,0 +1,102 @@
...
+  - Start involving PR teams. Write rough drafts of announcements. Begin creating dashboard.
+
+During 2023:
+
+- January 9 to January 20:
+  - Finalize test selection.
+  - Complete announcement plans. Prepare dashboard for launch.
+- January 23 to January 27: Launch.
+
+If necessary, additional changes to the test lists can be made throughout the year with the consensus of the team, see [updating the tests](#updating-the-tests).
```
sideshowbarker, line 33:
```suggestion
If necessary, additional changes to the test lists can be made throughout the year with the consensus of the team; see [updating the tests](#updating-the-tests).
```

rfcs/interop_2023.md
```diff
@@ -0,0 +1,102 @@
...
+  - Complete announcement plans. Prepare dashboard for launch.
+- January 23 to January 27: Launch.
+
+If necessary, additional changes to the test lists can be made throughout the year with the consensus of the team; see [updating the tests](#updating-the-tests).
+
+### Governance
+
+The [Interop team](https://github.com/orgs/web-platform-tests/teams/interop) will manage the overall Interop 2023 effort. The team will be formed by renaming the [’22 team](https://github.com/orgs/web-platform-tests/teams/interop-2022).
+
+The team makes decisions based on consensus, which is defined as support from at least two participating organizations and no opposition.
```
sideshowbarker, line 39:
```suggestion
The team makes decisions based on consensus, which is defined as support from at least two participating organizations, and no opposition.
```

rfcs/interop_2023.md
```diff
@@ -0,0 +1,102 @@
...
+
+The Interop 2023 effort is subject to the [WPT code of conduct](https://github.com/web-platform-tests/wpt/blob/master/CODE_OF_CONDUCT.md).
+
+### Proposals
+
+During the proposals period (September 15 to October 15), proposals are submitted in the [Interop repo](https://github.com/web-platform-tests/interop). A proposal should be as specific as reasonably possible, but multiple proposals may be grouped later in the selection process.
+
+To make proposal scope unambiguous, accepted proposals should be fully documented in the repository. To this end, and to make proposal review easier, the team may require that proposals are submitted as pull requests. Whether and how discussions, issues and pull requests are used in the proposals process is left to the team to decide before the proposal period begins.
+
+There are two kinds of proposals: focus areas where progress is measured using test results, and investigation efforts where progress is measuring against a task list.
```
sideshowbarker, line 51:
```suggestion
There are two kinds of proposals: focus areas where progress is measured using test results, and investigation efforts where progress is measured against a task list.
```

rfcs/interop_2023.md
```diff
@@ -0,0 +1,102 @@
...
+
+- Is this area a problem for web developers? For example, survey data and browser bug counts can help.
+- Is this area a problem for users?
+- How widely used is this feature? For example, [use counters](https://www.chromestatus.com/metrics/feature/popularity), [HTTP Archive](https://httparchive.org/) and [State of CSS](https://2021.stateofcss.com/en-US/features/) can help.
+- Specification for this feature.
+- A list of tests on wpt.fyi which will be used to score progress towards interoperability on the feature.
+
+#### Investigation efforts
+
+If an area doesn't have a good spec and test suite it's not possible to use wpt.fyi test results to judge progress. Instead, some other work is needed to bring the area to state where test results can be used.
```
sideshowbarker, line 67:
```suggestion
If an area doesn't have a good spec and test suite, then it's not possible to use wpt.fyi test results to judge progress. Instead, some other work is needed to bring the area to state where test results can be used.
```

rfcs/interop_2023.md
```diff
@@ -0,0 +1,102 @@
...
+
+Per-area and overall metrics will be defined in a similar (but not necessarily identical) fashion to [Interop 2022](./interop_2022.md#metrics). The number of areas and their contribution to the total score will be decided in the proposal selection phase.
+
+### Dashboard
+
+A public dashboard tracking the metrics over time will be available at https://wpt.fyi/interop-2023.
+
+## Risks
+
+No set of tests is without flaws, and trying to improve a metric based on test results can lead to bad outcomes such as shallow implementations that pass the tests but still isn't usable in practice. The main mitigation for this is a careful review of the tests for the proposed areas, and to include only tests that are judged to be high quality. Also, the tests are not frozen but can be updated or even removed during the year as issues are uncovered.
```
sideshowbarker, line 102:
```suggestion
No set of tests is without flaws, and trying to improve a metric based on test results can lead to bad outcomes such as shallow implementations that pass the tests but still aren't usable in practice. The main mitigation for this is a careful review of the tests for the proposed areas, and to include only tests that are judged to be high quality. Also, the tests are not frozen but can be updated or even removed during the year as issues are uncovered.
```

sideshowbarker (approved):

foolip:
Thanks for the review and fixes, @sideshowbarker!

foolip:
I've now renamed the repo and the team, as per the RFC.

PR #117, opened 2022-07-06
RFC 117: Remove dated tags for epoch branches

foolip:


Ms2ger (approved):

jgraham (commented):
I don't believe we rely on this for anything, so no objections on that front, but the RFC doesn't seem to set ouf the case for the change, only that there are alternative ways to get the same information.

foolip:
> I still rely on this tags when there is a issue with WebKitGTK on the CI that needs to be investigated.

Can you describe how you use the existing tags? Is it to figure out which runs have been triggered in order to find the CI runs in the first place, or is it when you're already looking at a failed CI run and want to find some previous run? Or something else?

jgraham:
> We could also do likewise with the merge tags.

That would be a ~breaking change for the gecko sync (we use the tags in the first instance to associate a commit on master with a PR). Not impossible to update ofc, but not something we should just blindly change.

jgraham:
> My motivation is to make git fetch and git log less noisy. 

Can you update the RFC with this reasoning?

foolip:
> > We could also do likewise with the merge tags.
> 
> 
> 
> That would be a ~breaking change for the gecko sync (we use the tags in the first instance to associate a commit on master with a PR). Not impossible to update ofc, but not something we should just blindly change.

We also need regular tags to create releases, and release artifacts is how we store the manifests.

I would like to get rid of those tags (and lots of branches) too but that's not for this RFC.

foolip:
> > My motivation is to make git fetch and git log less noisy. 
> 
> 
> 
> Can you update the RFC with this reasoning?

Will do after I've heard from @clopez. It's still possible I've misunderstood what these tags are for.

gsnedders:
> > > We could also do likewise with the merge tags.
> > 
> > 
> > That would be a ~breaking change for the gecko sync (we use the tags in the first instance to associate a commit on master with a PR). Not impossible to update ofc, but not something we should just blindly change.
> 
> We also need regular tags to create releases, and release artifacts is how we store the manifests.
> 
> I would like to get rid of those tags (and lots of branches) too but that's not for this RFC.

ah, that's true; yeah, probably worthwhile as a separate discussion

clopez:
> The other potential option here would be to use a different namespace for the epoch refs, such as pushing them into `refs/epochs/*`.

That would be also a good option, i like it. It keeps the info for those interested on it (via `git show-ref`) meanwhile it hides this info from the default view of `git log` / `git tag`.

PR #118, opened 2022-07-14
RFC-118: Proxy everything to WPT server instead of PAC

noamr:
The PAC implementation reached a roadblock since Mac (Safari+Chrome) does not allow programmatic setting of PAC URLs. Proposing a simpler solution that would work for more tests that require a proxy.

### DIFF
--- a/rfcs/proxy_all.md
+++ b/rfcs/proxy_all.md
@@ -1,0 +1,23 @@
+# RFC 118: Use a "proxy all" directive instead of PAC
+
+## Summary
+
+Proxy all traffic to the WPT server for certain test, when they need a fresh registrable domain.
+
+## Details
+
+See [RFC 112](https://github.com/web-platform-tests/rfcs/pull/112) for original reasoning.
+The issue with using PAC files as previously suggested, is that they don't work on Mac. Mac only
+supports setting the PAC URL in the OS settings, and not programatically. This is true not only for
+Safari.
+
+Most of the tests that require this need a fresh domain, but do not necessarily require the whole set
+of features offered by PAC files.
+
+So the proposal is that instead of relying on PAC as implemented and previously proposed, have a simpler
+meta tag that allows proxying everything to the WPT server , e.g.
+`<meta name="proxy" content="all">`.
+
+## Risks
+
+* As is with pac, more a limitation than a risk - by generating ad-hoc registrable domains we won't be able to use the certificate, which would limit that kind of test to   non-secure only.


noamr:
@jgraham sorry to bug you with more proxy stuff. Thoughts?

jgraham:
It sounds simpler at least. If you have an experimental implementation I think that would help me reason about any tradeoffs or issues that I'm not immediately thinking of.

noamr:
> It sounds simpler at least. If you have an experimental implementation I think that would help me reason about any tradeoffs or issues that I'm not immediately thinking of.

Sure, I'll post a WIP PR. It's based on the PAC implementation so at least that work was not a waste :)

gsnedders:
What's the intention as to how to proxy everything?

noamr:
> What's the intention as to how to proxy everything?

Pass the WPT server's address as an HTTP proxy to the browser prefs. All requests would go to the WPT server, and some of them would have a `Host` header. It should generally "just work".

gsnedders:
> The PAC implementation reached a roadblock since Mac (Safari+Chrome) does not allow programmatic setting of PAC URLs. 

As you've now found with https://github.com/web-platform-tests/wpt/pull/34911, the limitation with Safari is an inability to set proxy configuration as all (this is rdar://36854985), rather than there being any different/special with PACs. (I don't know what the situation is for Chrome on macOS?)

noamr:
> > The PAC implementation reached a roadblock since Mac (Safari+Chrome) does not allow programmatic setting of PAC URLs.
> 
> As you've now found with [web-platform-tests/wpt#34911](https://github.com/web-platform-tests/wpt/pull/34911), the limitation with Safari is an inability to set proxy configuration as all (this is rdar://36854985), rather than there being any different/special with PACs. (I don't know what the situation is for Chrome on macOS?)

Right. Still I think using the proxy-all technique makes everything simpler.

jgraham:
I imagine the usecase it doesn't support is adding latency early in the connection (by busy-looping in the PAC). Is that still an important requirement.

noamr:
Ping?

noamr:
Abandoning this as RFC112 (PAC) covers this sufficiently and the issue in Chrome has been resolved.

jgraham:
Do we have any reason to believe that this might be more likely to be usable in Safari before the PAC approach?

noamr:
> Do we have any reason to believe that this might be more likely to be usable in Safari before the PAC approach?

No.  As @gsnedders had mentioned, the Safari issue is not related to PAC vs. regular proxy, it's only about setting proxy at all in WebDriver.

PR #119, opened 2022-07-26
RFC 119: Allow wptrunner to retry tests with unexpected results

jonathan-j-lee:
[Rendered](https://github.com/jonathan-j-lee/rfcs/blob/no-repeat-expected/rfcs/retry_unexpected.md)


Ms2ger (approved):
Seems sensible.

rfcs/no_repeat_expected.md
```diff
@@ -0,0 +1,57 @@
...
+## Summary
+
+Add a `--no-repeat-expected` flag to wptrunner to skip tests in subsequent
+repeat iterations once they run as expected.
+
+## Details
+
+### Rationale
+
+This flag is used with `--repeat` to implement a retry mechanism.
```
jgraham, line 12: It seems like this flag actually has two effects:
* Retries tests that got an unexpected result up to N times.
* Changes the result handling to prefer the expected result in case of multiple results (i.e for an expected PASS, FAIL, PASS will be treated just like PASS).

So rather than viewing this as a modofoer to `--repeat`, I suggest we just make a new flag like `--retry-unexpected=N` and document that it has the behaviour of taking any expected status as the final result.

rfcs/no_repeat_expected.md
```diff
@@ -0,0 +1,57 @@
...
+  timeouts).
+* Incomplete expectation data checked into the trunk of a browser vendor's
+  source control system.
+  The CI check might not surface statuses that should have been added.
+
+### Proposal
+
+When passed `--no-repeat-expected`, wptrunner will run all loaded tests in the
+initial repeat iteration.
+On subsequent iterations, tests that ran as expected in a previous iteration
```
jgraham, line 25: Do we actually need the `SKIP` status? I think it might be easier to just not generate a status for those tests (since `SKIP` usually implies that the test was actively disabled, not that it was just not selected for the current run).

jgraham (commented):
Generally this looks like a reasonable proposal, just a couple of comments on the details.

rfcs/no_repeat_expected.md
jonathan-j-lee: Thanks for the info. No need for `SKIP`s then.

jonathan-j-lee:
Friendly ping @jgraham

jgraham (approved):
Thanks for the ping, we also discussed this at the infra meeting, and I think it looks good now.

jonathan-j-lee:
Unless there are additional comments, I think this is ready to merge @web-platform-tests/wpt-core-team?

PR #120, opened 2022-08-03
RFC 120: Interop Team Charter

foolip:
[Rendered](https://github.com/web-platform-tests/rfcs/blob/interop_charter/rfcs/interop_charter.md)

PR #122, opened 2022-09-13
RFC 122: Remove browser specific failures graph

jgraham:
[Rendered](https://github.com/web-platform-tests/rfcs/blob/remove_bsf/rfcs/remove_bsf.md).

### DIFF
--- a/rfcs/remove_bsf.md
+++ b/rfcs/remove_bsf.md
@@ -1,0 +1,69 @@
+# RFC : Remove browser specific failures from wpt.fyi
+
+## Summary
+
+wpt.fyi currently shows a chart of "browser specific failures"; a
+score of tests that are failing only in a single browser. This RFC
+proposes entirely removing that graph from wpt.fyi.
+
+## Details
+
+The original motivation for browser-specific-failures was to provide
+browser vendors with insights into tests that might be causing interop
+problems and therefore might be especially valuable to spend
+engineering effort fixing. On the basis of this hypothesis, a graph
+was added to wpt.fyi showing a browser-specific-failures "score" for
+each browser engine, so that vendors could track their progress on
+fixing these issues.
+
+However at this point we have identified a number of issues with
+"browser specific failures" as a metric, for example:
+
+* It's hard to correlate a change in the score to a change in browser
+  behaviour.
+
+* The way the score is computed biases the score toward missing
+  features rather than interop failures in already shipping features
+  (since a missing feature usually causes a large number of failures).
+
+* The metric doesn't provide any way of controlling for the user
+  impact of failures; browsers can get a "bad" score from a large
+  number of failures that in practice don't cause any observed
+  problems for authors.
+
+For these reasons and others, we haven't reached a critical mass of
+adoption for browser specific failures as a metric to improve interop
+on the web platform, and its original function has been largely
+replaced by the Interop-20xx project.
+
+Although browser specific failures isn't providing the initially hoped
+for value, having it on the wpt.fyi does create some work as it
+encourages people to try to understand the current scores or changes
+in the score.
+
+Removing the graph entirely seems like the simplest way to indicate
+that we no longer consider this useful as a metric.
+
+## Alternatives
+
+* Keep the graph but move it to a less prominent page.
+
+  Although this would make the graph less obvious, it would still
+  imply some endorsement of browser specific failures as useful at the
+  level of a metric. Since Interop scores are a metric that browser
+  vendors have explicitly committed to, and which solve many of the
+  problems with browser specific failures, it's better to clearly
+  commit to one public metric.
+
+## Risks
+
+* Browser engineers might be using browser specific failures as a way
+  to identify good tests to fix.
+
+  This doesn't depend on having a metric / graph of browser-specific
+  failures, and could be better solved by making it easier to see an
+  actual list of browser specific failures in a given feature. This is
+  already possible on wpt.fyi, but is quite complex to write as a
+  query. Other frontends on wth wpt.fyi data like
+  https://jgraham.github.io/wptdash/ provide a engineer-focused view
+  of this data.
--- a/rfcs/remove_bsf.md
+++ b/rfcs/remove_bsf.md
@@ -1,10 +1,11 @@
-# RFC : Remove browser specific failures from wpt.fyi
+# RFC 122: Remove browser specific failures from wpt.fyi
 
 ## Summary
 
-wpt.fyi currently shows a chart of "browser specific failures"; a
-score of tests that are failing only in a single browser. This RFC
-proposes entirely removing that graph from wpt.fyi.
+[wpt.fyi](https://wpt.fyi) currently shows a chart of "browser
+specific failures"; a score of tests that are failing only in
+a single browser. This RFC proposes entirely removing that graph
+from wpt.fyi.
 
 ## Details
 


DanielRyanSmith:
I agree that there is likely a better way to leverage these metrics, and it seems like an outright improvement to developer utility if the graph is replaced with links that display queries of all BSFs for a given browser.

foolip:
I think we should remove the graph from the top of /results, but I don't think we should *just* remove it. We have triaged Chrome-only failures to keep our BSF number under 500, and I see it might be time to do that again. And based on PRs from @gsnedders to the metrics code I assume they've looked at it too.

/insights already has "Anomalies" which allows getting to views for browser-specific failures, like this one:
https://wpt.fyi/results/?label=master&label=experimental&product=chrome&product=firefox&product=safari&view=subtest&q=%28chrome%3A%21pass%26chrome%3A%21ok%29%20%28firefox%3Apass%7Cfirefox%3Aok%29%20%28safari%3Apass%7Csafari%3Aok%29

(Although it's buggy, I filed https://github.com/web-platform-tests/wpt.fyi/issues/2964.)

If I can make a wishlist, it would be:
- Put the graph under /insights
- Make it possible to click points in the graphs to get the corresponding list of tests (even if the weighting isn't the same)
- Ensure the "Anomalies" widget gives the test list of the most recent results, if one is comparing Chrome vs. Firefox vs. Safari

foolip:
Maybe a view like this would be the most friendly:
https://wpt.fyi/results/?label=master&label=experimental&product=chrome&product=firefox&product=safari&view=subtest&q=%28chrome%3A%21pass%26chrome%3A%21missing%29%20firefox%3Apass%20safari%3Apass

jgraham:
My view is that if people want to use the concept of browser specific failures as an internal tool for understanding areas of interop difficulty that's good, and I fully support that. But I don't think we have widespread agreement on its use as a metric, and the reasoning in the RFC suggests that the lack of curation makes the numbers difficult to interpret.

If specific vendors want a number to look at I think it's reasonable to make that number an internal metric instead. That has the additional advantage that it allows some customisation e.g. it allows filtering the inputs to exclude tests that aren't considered a priority/problem for whatever reason, or dividing up the score into team-specific metrics rather than just having one top-level number. That isn't something we can do with a purely shared metric.

gsnedders:
While I've certainly looked at the metric, it's far from the only data derived from WPT results that I've looked at. I think I otherwise agree with @jgraham here.

gsnedders:
To be clear, as the RFC says, there are a variety of biases with this metric, and some of these get quite extreme:

Looking at the Safari data, `/html/canvas/offscreen` accounts for 32.48% of Safari's current score, and `/css/css-ui/compute-kind-widget-generated` 7.56%.

I don't personally believe 40.04% of Safari's "incompatibility" or "web developer pain" (or however we want to define the goal of the BSF metric) is down to those two features.

If we look at the graph over the past year with those two directories removed, we see a very different graph:

The browser-specific-failure graph with a slowly decreasing Safari over the first six months, stabilising afterwards

foolip:
@gsnedders thanks, that clearly demonstrates the outsized impact of test suites with lots of individual tests. For comparison/posterity, here's the current BSF graph on wpt.fyi:

image

A few options for improving the metric:
- Simply excluding some directories, like `/referrer-policy` and `/html/canvas/offscreen`
- Giving each directory equal weight
- Equal weight as a starting point, but with manually adjusted weights for some huge directories like `/html`

I disagree with deleting the graphs outright, but would be happy with both moving it to `/insights` and tweaking it.

jgraham:
I think a proposal for a new interop metric, even if based on BSF, would clearly be something for the Interop team to consider.

past:
Improving the BSF metric seems like a worthwhile goal, either through ideas like the ones Sam and Philip propose or through a reimagined Interop metric based on BSF as James suggests. I would encourage the Interop team to explore that path.

However, since we don't have that yet, removing the metric entirely would be a step backwards. In Chromium we _do_ pay attention to the overall score and invest considerably in improving interoperability over time. Hiding that number in favor of team-specific metrics will regress that effort. It will reduce visibility of Chromium interoperability issues at the organizational level and will pass the burden to individual teams with different priorities. 

From my perspective, removing things that are currently in use without a suitable replacement is wrong. But perhaps moving the graph to /insights as an interim step before we have an improved metric would be a reasonable compromise.

karlcow:
Not fully matured idea: If the graph is a kind of barometer on web technologies support **across browsers**, would it make sense to have there things which are only supported (standard positions) uniformly by the 3 browsers represented in the graph?

karlcow:
@past 

>  It will reduce visibility of Chromium interoperability issues at the organizational level and will pass the burden to individual teams with different priorities.

What is(are) the audience(s) for the graph? 

And depending on that what are the useful views for each specific audience?

past:
The audience is senior leaders who are making sure Chromium remains interoperable and competitive with other browser engines over time. The current view of overall browser specific failures is still useful in that task.

jgraham:
Whilst I'm happy that Chrome's leadership are finding the graph useful, that usefulness as a metric is not a consensus position among browser vendors, and therefore it seems more appropriate to host it at a Chromium-specific location.

foolip:
@jgraham how do you see this RFC interacting with https://github.com/web-platform-tests/rfcs/pull/120? Per that RFC the interop team will take ownership of this.

And I now see that RFC should be considered passed, given approvals and several weeks passing.

past:
While still supporting improvements to the graph, I will say that adding up the numbers in your three bullets above seems to reasonably explain the lack of impact of the improvements you made.

gsnedders:
> While still supporting improvements to the graph, I will say that adding up the numbers in your three bullets above seems to reasonably explain the lack of impact of the improvements you made.

The number of subtests do, yes. But that's a complete coincidence, given the "normalisation" to test.

If you look at the [actual directory-level diff](https://docs.google.com/spreadsheets/d/1dqsp6RobCpvGDVaCEdj-7ZvlJNqPq6HkiOzga7RTn2o/edit#gid=1808308357), it becomes very apparent the overwhelming majority of the change is in `/html/canvas/offscreen`. And if you look at [that directory](https://wpt.fyi/results/html/canvas/offscreen?run_id=5701482373447680&run_id=6320955010318336&run_id=5737772443762688&run_id=6276531089833984&run_id=5644647339655168&run_id=6308121446711296), you'll see there's only 1805 tests (1910 subtests), which account for almost the entire lack of overall progression.

Again, the problem is to a large degree weighting all the tests the same.

gsnedders:
> @jgraham how do you see this RFC interacting with #120? Per that RFC the interop team will take ownership of this.
> 
> And I now see that RFC should be considered passed, given approvals and several weeks passing. I'll hold off merging for a bit though.

For anyone confused, I believe we (the WPT Core Team) decided to defer this RFC until the Interop Team had time to consider it.

foolip:
We never resolved (merged) https://github.com/web-platform-tests/rfcs/pull/120 but indeed that seems like the best way to handle this.

gsnedders:
Adding Test262 to WPT has again showed the issues with the metric; Safari Technology Preview's BSF has jumped by 7,148, of which 6,565 is Temporal, to a total of 11,317.

The relative BSF scores continue to be cited as evidence that Safari is massively more buggy than Chrome or Firefox, but this is clearly misinterpreting what the metric’s value actually represents: I don’t think one can reasonably claim that 58% of STP’s bugginess is due to Temporal not being supported.

Years on, it remains the case that the usefulness of BSF isn't a consensus position among browser vendors, and when people continue to misinterpret the metric, the proposal to remove BSF continues to be something I would support.

jgraham:
I continue to agree this should be removed.

captainbrosset:
Since Temporal is the feature which contributes the most to the current Safari spike in the BSF graph, and since we know it's a much needed developer feature, it would be great to know when Safari expects to deliver an implementation to it in the browser.

karlcow:
@captainbrosset the removal of the graph doesn't remove the tests.
They are still there. So everything is still accessible.
https://wpt.fyi/results/third_party/test262/test?label=master&label=experimental&aligned

jgraham:
Management of this was delegated to the Interop team in RFC #120 (https://github.com/web-platform-tests/rfcs/blob/main/rfcs/interop_charter.md). Given that, this issue can be closed in this repo for now, on the basis that the Interop team is working on the problems under discussion.

PR #123, opened 2022-10-04
Document that no RFC is needed for most testdriver.js additions

foolip:


Ms2ger (approved):
Fine by me

jgraham:
I think we decided that this should itself be an RFC.

foolip:
@past agreed to turn it into a RFC. It's pretty much this language, plus stating that there should be tests.

jcscottiii:
@foolip Now that #127 is merged, this can be merged. Are there any remaining changes to this PR?

PR #128, opened 2022-11-18
RFC 128: Consume User Activation

marcoscaceres:
See details in the proposed .md. 

tl;dr: this is a proposal to add...

* change on WPT side: https://github.com/web-platform-tests/wpt/pull/37176

```JS

 /**
  * Consumes the user activation
  *
  * Matches the behavior of the `Consume User Activation of window`
  * as proposed in: https://github.com/w3c/webdriver/pull/1695
  *
  * Which corresponds to these steps in HTML:
  * https://html.spec.whatwg.org/#consume-user-activation
  *
  * @param {WindowProxy} context - Browsing context in which to run the
  *                                call.
  * @returns {Promise<boolean>} fulfilled when user activation is consumed.
  */
 consume_user_activation(context=null) {
     return window.test_driver_internal.consume_user_activation(context);
 },
```

### DIFF
--- a/rfcs/consume_user_activation.md
+++ b/rfcs/consume_user_activation.md
@@ -9,7 +9,7 @@ Some of these means are non-standard in as far as they are not specified to cons
 
 The workarounds being used are as follows, with several drawbacks:
 
-* Using Fullscreen API: non-standard, takes significant time to put an element into fullscreen. 
+* Using Fullscreen API: it takes significant time to put an element into fullscreen. 
 * Using window.open() and window.close() - non-standard, not mobile friendly. 
 * Using Payment Request API - not widely implemented (e.g., not exposed in Gecko). 
 
--- a/rfcs/consume_user_activation.md
+++ b/rfcs/consume_user_activation.md
@@ -1,0 +1,41 @@
+# RFC: Consume user activation 
+
+Editor: Marcos Caceres, Apple Inc.
+
+## Summary 
+
+Tests in WPT are currently relying on various indirect means to "[consume user activation](https://html.spec.whatwg.org/#consume-user-activation)" [HTML].
+Some of these means are non-standard in as far as they are not specified to consume user activation, meaning that the tests are inshrining non-standard behavior for conformance purposes.
+
+The workarounds being used are as follows, with several drawbacks:
+
+* Using Fullscreen API: non-standard, takes significant time to put an element into fullscreen. 
+* Using window.open() and window.close() - non-standard, not mobile friendly. 
+* Using Payment Request API - not widely implemented (e.g., not exposed in Gecko). 
+
+We also looked at a range of other possible APIs that consume user activation, and found [none to be suitable](https://github.com/web-platform-tests/wpt/issues/36727#issuecomment-1296349964). 
+
+## Details 
+
+We would like to propose the addition of an async function `test_driver.consume_user_activation()` method that returns a `Promse<boolean>` (representing if the activation was consumed or not). For example: 
+
+```
+const consume = await test_driver.consume_user_activation();
+```
+
+The `consume_user_activation()` method can take a `Window` <var>context</var> (which defaults to the current Window object). Allowing consumption to happen at a particular window, if required. 
+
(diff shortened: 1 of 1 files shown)


Ms2ger:
Is this covered by  #127?

foolip:
> Is this covered by #127?

After both that and https://github.com/w3c/webdriver/pull/1695 are landed, yes. But I think we can keep this as its own RFC to not depend on any of that.

rfcs/consume_user_activation.md
```diff
@@ -0,0 +1,41 @@
+# RFC: Consume user activation
```
foolip, line 1:
```suggestion
# RFC 128: Consume user activation 
```

rfcs/consume_user_activation.md
```diff
@@ -0,0 +1,41 @@
...
+Editor: Marcos Caceres, Apple Inc.
+
+## Summary 
+
+Tests in WPT are currently relying on various indirect means to "[consume user activation](https://html.spec.whatwg.org/#consume-user-activation)" [HTML].
+Some of these means are non-standard in as far as they are not specified to consume user activation, meaning that the tests are inshrining non-standard behavior for conformance purposes.
+
+The workarounds being used are as follows, with several drawbacks:
+
+* Using Fullscreen API: non-standard, takes significant time to put an element into fullscreen.
```
foolip, line 12: Can you link to https://github.com/whatwg/fullscreen/pull/153 so this doesn't seem non-standard to someone reading the RFC in the future?

rfcs/consume_user_activation.md
```diff
@@ -0,0 +1,41 @@
...
+The `consume_user_activation()` method can take a `Window` <var>context</var> (which defaults to the current Window object). Allowing consumption to happen at a particular window, if required. 
+
+The `consume_user_activation()` method would be implemented via the proposed addition of “[Consume user activation of Window](https://github.com/w3c/webdriver/pull/1695)” to the Web Driver specification.
+
+This prposal has several advantages:
+ * it's fast - no opening windows or waiting for fullscreen to enter/exit.  
+ * it's lightweight - it doesn't create new browsing contexts.
+ * it's mobile friendly - as above.  
+ * it's build for purpose - no more using other APIs to indirectly achieve the desired outcome.
+ * It's really simple - it compliments `.bless()` and other user activaton functionality already available.
```
foolip, line 35:
```suggestion
 * It's really simple - it complements `.bless()` and other user activation functionality already available.  
```

foolip (approved):

rfcs/consume_user_activation.md
```diff
@@ -0,0 +1,41 @@
...
+Editor: Marcos Caceres, Apple Inc.
+
+## Summary 
+
+Tests in WPT are currently relying on various indirect means to "[consume user activation](https://html.spec.whatwg.org/#consume-user-activation)" [HTML].
+Some of these means are non-standard in as far as they are not specified to consume user activation, meaning that the tests are inshrining non-standard behavior for conformance purposes.
+
+The workarounds being used are as follows, with several drawbacks:
+
+* Using Fullscreen API: non-standard, takes significant time to put an element into fullscreen.
```
marcoscaceres, line 12:
```suggestion
* Using Fullscreen API: it takes significant time to put an element into fullscreen. 
```

marcoscaceres: As https://github.com/whatwg/fullscreen/pull/153 is about to land, let’s treat it as standard.

mustaqahmed (approved):

foolip:
Per https://github.com/web-platform-tests/rfcs#the-rfc-process we can merge this if there's no further feedback in 1 week.

jgraham:
I don't think it makes sense to land this without the WebDriver PR landing. Without that PR, the details of this API are underspecified and it doesn't meet our bar for inclusion.

marcoscaceres:
That's ok @jgraham (and maybe we can get that landed also within a week? 🙏).

As we have multiple implementers interested, who do we need to bug over on the Web Driver side to get things moving?

jcscottiii:
This RFC is good to go now that the PR is merged right?
@jgraham @foolip

jgraham:
Per policy we no longer require an RFC for testdriver additions that correspond directly to WebDriver endpoints. So we can either merge this or not, but in any case it doesn't block adding the feature.

PR #134, opened 2023-03-09
Update minimum supported Python to 3.7

jgraham:
This ought to be possible now that Gecko CI is all using 3.7+.

### DIFF
--- a/rfcs/python_37.md
+++ b/rfcs/python_37.md
@@ -1,0 +1,43 @@
+# RFC 134: Bump minimum supported Python to 3.7
+
+## Summary
+
+Increase the minimum supported Python version to 3.7.
+
+## Details
+
+Currently we support Python 3.6+. However 3.6 was EOL in December
+2021, and the Python ecosystem has generally moved away from offering
+3.6 support. Practically this means that we have been unable to update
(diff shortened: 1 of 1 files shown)


gsnedders (approved):

rfcs/python_37.md
```diff
@@ -0,0 +1,43 @@
...
+
+Previously the blocker for updates was the downstream integration with
+Gecko's CI which still used Python 3.6. This has recently been updated
+to use [3.7](https://bugzilla.mozilla.org/show_bug.cgi?id=1734402),
+which unblocks updating to that version.
+
+Practically this requires several manual changes:
+
+ * Change the CI jobs running under Python 3.6 to instead use 3.7.
+ * Update any outdated vendored dependencies that are being held back
```
foolip, line 22: We should do this *after* the 3.7 upgrade, right?

foolip (approved):

rfcs/python_37.md
gsnedders: No, that sounds very much like an operational concern.

jgraham:
No objections here for more than a week, so merging this.

PR #135, opened 2023-03-10
Rename private to local in address space overrides

johnathan79717:
In https://github.com/WICG/local-network-access/issues/91, we decided to rename Private Network Access to Local Network Access. The spec has already been renamed in
https://github.com/WICG/local-network-access/pull/97. The latest spec: https://wicg.github.io/local-network-access/

Chromium has also renamed private to local in
https://source.chromium.org/chromium/chromium/src/+/d5b320062c4982d2e215386e068826f8bc97c512 so the ip-address-overrides switch value needs to be updated too.

This PR makes the corresponding changes in the RFC.

cc @letitz

### DIFF
--- a/rfcs/address_space_overrides.md
+++ b/rfcs/address_space_overrides.md
@@ -3,15 +3,15 @@
 ## Summary
 
 The goal of this RFC is to enable web platform tests to cover the
-[Private Network Access](https://wicg.github.io/private-network-access)
+[Local Network Access](https://wicg.github.io/local-network-access)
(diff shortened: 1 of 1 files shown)


johnathan79717:
Thanks for the approval.  Can someone with write access merge this PR?

jgraham:
No, per the process we need at least a week to allow people to read the RFC and bring forward any concerns.

jgraham:
Although we don't usually do in-place edits of RFCs like this.

PR #136, opened 2023-03-14
RFC 136: Remove the CSS build system

foolip:
[Rendered](https://github.com/web-platform-tests/rfcs/blob/css_build_system/rfcs/css_build_system.md)

### DIFF
--- a/rfcs/css_build_system.md
+++ b/rfcs/css_build_system.md
@@ -17,6 +17,8 @@ These lint rules will be removed:
 
 The MISSING-LINK ("Testcase file must have a link to a spec") rule which also applies only to `css/` will _not_ be removed, as it isn't only used by the build system, but is also a matter of style and preference. This may be revisited later.
 
+This RFC will be implemented by https://github.com/web-platform-tests/wpt/pull/38976.
+
 ## Risks
 
 Reviving http://test.csswg.org/suites/ becomes harder with this removal, so if that system has important functionality, it will require more work to meet those needs again. However, the web-platform-tests project remains open to contributions to support missing use cases, should the need arise.
--- a/rfcs/css_build_system.md
+++ b/rfcs/css_build_system.md
@@ -1,0 +1,22 @@
+# RFC 136: Remove the CSS build system
+
+## Summary
+
+Remove CSS build system in `css/tools/` and the path lint rules for `css/` that support that build system.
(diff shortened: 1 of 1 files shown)


jgraham (approved):
This needs to be fleshed out more, clearly, with specific details about what will be removed, but in general I"m strongly in favour of this and don't expect to object to any of those details.

rfcs/css_build_system.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Summary
+
+Remove css/tools/ and the special lint rules for css/.
+
+## Details
+
+http://test.csswg.org/suites/ hasn't been updated since January.
+
+TODO: does the CSSWG want to keep the spec link rule?
```
jgraham, line 11: I would proceed for now on the assumption the answer is "yes", but it's worth asking the question. If not we can always remove the requirement in a second step.

rfcs/css_build_system.md
```diff
@@ -0,0 +1,15 @@
+# RFC 136: Remove the CSS build system
+
+## Summary
+
+Remove css/tools/ and the special lint rules for css/.
+
+## Details
+
+http://test.csswg.org/suites/ hasn't been updated since January.
```
gsnedders, line 9: https://github.com/w3c/csswg-drafts/issues/6896 makes it sound like there have been serious problems there for much longer. (And indeed I get the error given in https://github.com/w3c/csswg-drafts/issues/6896#issuecomment-996047519 when I try to build it locally.)

rfcs/css_build_system.md
foolip: I've done it in a separate commit. Let's just say it's not part of this RFC to reduce risk.

gsnedders:
CSS WG [resolved](https://github.com/w3c/csswg-drafts/issues/6896#issuecomment-1499457107), so we can proceed with this.

Ms2ger (approved):

jgraham:
@foolip can you update this to include the detailed plan, and remove the TODOs?

foolip:
@jgraham it is done!

jgraham (approved):

foolip:
I've marked https://github.com/web-platform-tests/wpt/pull/38976 as ready for review now.

PR #138, opened 2023-03-15
Introduce 'subsuites' functionality to wptrunner

jgraham:
This is intended to allow running multiple groups of tests, corresponding to different browser configurations in the same wptrunner process.

### DIFF
--- a/rfcs/subsuites.md
+++ b/rfcs/subsuites.md
@@ -1,0 +1,348 @@
+# RFC 137: support subsuites in wptrunner
+
+## Summary
+
+Support running tests in multiple different browser configurations in
+a single wptrunner invocation.
+
+## Details
+
+It is common for browser vendors to want to run web platform tests
+with particular configurations of the browser. For example when
+developing a major new feature, or refactoring existing code, they may
+want to run with non-default configuration to enable the new
+code paths. Often, for cost and efficiency reasons, these test runs
+will initially be limited to just the tests most likely to be affected
+by the changes e.g. a change to session history is highly likely to
+affect tests involving navigation, but highly unlikely to affect CSS
+layout.
+
+Typically vendors wanting to enable multiple configurations had to
+define those configurations externally, and invoke wptrunner multiple
+times, once per configuration. This has some advantages; in particular
+it keeps the vendor-specific issue of browser configurations cleanly
+separated from the actual business of running tests. However there are
+also a couple of disadvantages:
+
+* When configurations are handled externally, they are either written
+  into a CI configuration, which is then difficult to run locally, or
+  the vendor is responsible for writing a substantial frontend on top
+  of wptrunner which handles invoking it for each of the different
+  configurations.
+
+* Even with a vendor frontend it's difficult to schedule the tests
+  optimally across multiple cores. This is because you essentially
+  have two choices on how to handle the different
+  configurations. Option one is to run each configuration in serial,
+  with each wptrunner process using all available cores. This can be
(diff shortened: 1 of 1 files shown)


rfcs/subsuites.md
```diff
@@ -0,0 +1,348 @@
...
+1. Put configuration information inside the test id, so that from the
+   point of view of consumers, the same test run in different
+   configurations look like different tests. This is how the Chromium
+   CI system handles this problem with the concept of "virtual"
+   testsuites. It has the advantage that it requires minimal changes
+   to data structures; tests get an id like `/<prefix>/<configuration
+   name>/<actual test id>`, so from the point of view of consumers
+   these are different tests. However this is also a disadvantage;
+   every consumer that wants to process the results (e.g. to compare
+   results between configurations) has to know about the details of
```
WeizhongX, line 106: We don't compare results between configurations in a large scale. Developers will compare results between virtual/non-virtual tests by compare the output of the two, but this is done manually.

I actually have a concern that all expectations from all subsuite to a single file could make that file too complicated. But I will have to wait to see what are the dev's opinions.

jgraham: Comparing results between configurations is definitely something we are interested in.

Note that we don't have any proposal so far that would naturally allow writing multiple log files. But with this proposal it would be relatively easy to write a log handler that reads the log messages and demuxes them to multiple files, if that's what you want.

rfcs/subsuites.md
```diff
@@ -0,0 +1,348 @@
...
+
+The automatic metadata update works by collecting test results from
+multiple files, grouped by their `run_info` data. This can naturally
+be extended to support subsuites, just by extending the processing to
+create a `run_info` entry per subsuite rather than assuming the
+`run_info` is the same for all tests in a file.
+
+The main difficulty here is that the metadata expression language
+doesn't have a concept of `null`. So we need to represent the default
+subsuite name as the empty string in `run_info`, or add a `null`
```
WeizhongX, line 334: I think we would prefer to add a `null` concept. We already such issue for adding flag specific config in run_info. And we are using the empty string to represent non flag specific suite.

rfcs/subsuites.md
WeizhongX: Thanks. For now we will keep that in a single file and see what is devs' opinion.

WeizhongX:
Hi James, I have no additional comments and I will be ok if we review and approve this offline.

@foolip @past

Ms2ger (approved):

WeizhongX:
@foolip @past Can either of you approve this? I heard we need two approvals for RFCs.

jcscottiii (approved):

foolip (approved):

foolip:
@WeizhongX helped me find https://phabricator.services.mozilla.com/D171999 which (together with a mozlog release) I guess is a prerequisite for any WPT-side changes.

rfcs/subsuites.md
```diff
@@ -0,0 +1,348 @@
...
+
+* `config` (required) - Product-specific configuration options for the
+  subsuite. This is used to e.g. set the command line arguments used
+  when launching the browser, or the prefs set in the browser profile.
+
+* `run_info` (optional) - Any updates to the `run_info` associated with the
+  subsuite, other than its name (the `run_info["subsuite"] = name` key
+  is set implicitly).
+
+* `include` (optional) - A list of test id prefixes to include when
```
WeizhongX, line 227: jgraham@, I just see this is optional, which means the subsuite includes all tests?

PR #149, opened 2023-06-01
RFC 149: Support reftest variants

dlrobertson:
Add support for variants when writing reftests.

[Rendered](https://github.com/dlrobertson/wpt-rfcs/blob/reftest-variants/rfcs/reftest_variants.md)

Initial implementation: https://github.com/web-platform-tests/wpt/pull/40352


dlrobertson:
CC @jgraham

dlrobertson:
😆 should have double checked which RFC number this would be

Ms2ger:
> > I think the examples in this RFC are simplified too much to be convincing, but agree that the overall idea makes sense.
> 
> Good point. My intent was to stick to a well known example. Do you think it is worthwhile to create a more complex example with more parameters?

I think it's probably fine

jcscottiii:
As discussed in the last infra [meeting](https://github.com/web-platform-tests/wpt-notes/blob/master/minutes/2023-06-06.md#rfc-149-support-reftest-variants-link), we will merge this.

PR #158, opened 2023-07-11
RFC 158# Do not allow zero length variant names

WeizhongX:


WeizhongX:
@jgraham @gsnedders @past can you guys pls take a look? thanks!

WeizhongX:
> How many tests currently use zero-length variants?

There are about 200 tests (js or html) use zero-length variants, and they are mostly in websockets and navigation-api.

WeizhongX:
Hey @jcscottiii, can you also pls take a look? Thanks!

jgraham (approved):
I think this is OK.

For me the main argument against using `?` or `#` to indicate a zero length variant is how many libraries don't distinguish between "this has a query/fragment, but it's empty" and "this doesn't have a query/fragment". The URL spec does make the distinction, so I don't think there's a principled reason we couldn't depend on it, but it might be surprising and hard to implement. Meanwhile making people name each variant does seem like it adds a little extra documentation.

jcscottiii (approved):

WeizhongX:
@jcscottiii , I do not have write access, so can you help merge this one (if it is ready)? thanks!

PR #163, opened 2023-08-22
RFC #163: Add web_feature metadata file to web-platform-tests

jcscottiii:
Replacement of https://github.com/web-platform-tests/rfcs/pull/157 that uses the feedback from the last WPT [meeting](https://github.com/web-platform-tests/wpt-notes/blob/master/minutes/2023-08-01.md#rfc-157---add-feature-meta-tag-to-web-platform-tests-meta-data).

[Rendered](https://github.com/web-platform-tests/rfcs/blob/Add-feature_set-metadata-file-to-web-platform-tests/rfcs/feature-set-metadata-file-to-web-platform-tests.md)

### DIFF
--- a/rfcs/feature-set-metadata-file-to-web-platform-tests.md
+++ b/rfcs/feature-set-metadata-file-to-web-platform-tests.md
@@ -1,0 +1,221 @@
+# RFC #: Add feature_set metadata file to web-platform-tests
+
+Author: @jcscottiii
+
+# Introduction
+
+The web-platform-tests (WPT) project is a valuable resource for testing web
+platform features and contains valuable metadata such as feature specs.
+Recently, the [WebDX Community Group](https://www.w3.org/community/webdx/)
+began creating the [feature-set](https://github.com/web-platform-dx/feature-set)
+repository. That repository serves as a basic shared catalog of feature
+definitions of the web platform. Feature-set itself doesn't intend to produce
+data but rather link to existing data that will inform audiences which features
+are part of [Baseline](https://web.dev/baseline/). However, there's a great
+opportunity to connect the WPT ecosystem to the feature-set catalog. By doing
+this, it would enable users of wpt.fyi to filter by feature-set grouping, which
+is similar to
+[the ability to filter by spec](https://github.com/web-platform-tests/wpt.fyi/issues/1489).
+The RFC proposes the addition of a FEATURE_SET.yml metadata file which would
+enable the linkage between WPT and feature-set.
+
+# Proposed change
+
+The proposed change includes:
+
+1. Introduce a new metadata file, FEATURE_SET.yml
+2. Adjust the wpt-pr-bot to handle reviews of the changes
+3. Create a script to generate a manifest
+
+Below are the details for each step.
+
+## Step 1. Introduce a new metadata file, FEATURE_SET.yml
+
+Originally, there was a RFC that proposed adding these details to the existing
+META.yml. During the August 01, 2023 Monthly WPT
+[Meeting](https://github.com/web-platform-tests/wpt-notes/blob/master/minutes/2023-08-01.md#rfc-157---add-feature-meta-tag-to-web-platform-tests-meta-data),
+it was discussed that it should be a separate file. This section describes the structure of the file.
+
+This file is expected to be in the same places developers would expect a META.yml.
+
+### File examples
+
+```
+parser_version: v1
+apply_mode: RECURSIVE
+feature_set: subgrid
+overrides:
+- file_name: name.txt
+  feature_set: feature2
+```
+
+### Schema
+
+```json
+{
+    "$schema": "http://json-schema.org/draft-06/schema#",
+    "$ref": "#/definitions/Main",
+    "definitions": {
+        "Main": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "parser_version": {
+                    "type": "string",
+                    "description": "version of the parser"
+                },
+                "feature_set": {
+                    "type": "string",
+                    "description": "The feature set key"
+                },
+                "apply_mode": {
+                    "type": "string",
+                    "anyOf": [
+                    	{
+                            "const": "DEFAULT",
+                            "description": "Applies recursively until the presence of another FEATURE_SET.yml in a subdirectory."
+                        },
+                    	{
+                            "const": "FORCE_RECURSIVE",
+                            "description": "Applies recursively throughout the all sub-directories. Useful if feature_set hierarchy exactly matches wpt directory structure."
+                        }
+                    ],
+                    "default": "DEFAULT"
+                },
+                "overrides": {
+                    "type": "array",
+                    "items": {
+                        "$ref": "#/definitions/Override"
+                    }
+                }
+            },
+            "required": [
+                "feature_set",
+                "parser_version",
+                "apply_mode"
+            ],
+            "title": "Main"
+        },
+        "Override": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "file_name": {
+                    "type": "string",
+                    "description": " The file name of the test in the same directory as this file."
+                },
+                "feature_set": {
+                    "type": "string",
+                    "description": "The feature set key"
+                },
+                "override_mode": {
+                    "type": "string",
+                    "anyOf": [
+                    	{
+                            "const": "APPEND",
+                            "description": "Append to the current feature_set key(s) for the given file"
+                        },
+                    	{
+                            "const": "REPLACE",
+                            "description": "Replace the current feature_set key(s) for the given file"
+                        },
+                    ],
+                    "default": "APPEND"
+                },
+            },
+            "required": [
+                "feature_set",
+                "file_name"
+            ],
+            "title": "Override"
+        }
+    }
+}
+```
+
+The next section describes the manifest generation process section.
+
+## Step 2. Create a script to generate a manifest
+
+A command will be created in the `tools/manifest` folder to generate the
+FEATURE_SET_MANIFEST.json file. While it is in that folder, its logic will
+be independent of the existing manifest logic. (As discussed in the August 01,
+2023 Monthly WPT
+[Meeting](https://github.com/web-platform-tests/wpt-notes/blob/master/minutes/2023-08-01.md#rfc-157---add-feature-meta-tag-to-web-platform-tests-meta-data),
+it should not use the same logic.) The only thing it will it
+do is generate data in the same structure.
+
+By using the same format, wpt.fyi can use it to allow filtering by feature set.
+
+The rules of the generation are described in the file schema above by the
+directory level `apply_mode` flag and the per file `override_mode`.
+
+## Step 3. Adjust the wpt-pr-bot to handle reviews of the changes
+
+### Current State of wpt-pr-bot
+
+Currently, the wpt-pr-bot builds a list of PR reviewers by:
+1. Retrieving the paths for all of the files
+2. Finding the nearest META.yml file for each path
+3. Add the list of suggested_reviewers from a given META.yml to a set of
+   reviewers
+
+### Proposed changes
+
+Have the wpt-pr-bot filter the FEATURE_SET.yml file changes to only request
+reviews from feature-set contributors.
+
+This will reduce the amount of unneeded reviews from non feature-set contributors.
+
+---
+
+# Risks
+
+As a result of putting all the information in FEATURE_SET.yml, the overrides
+which are linked to the file can go out of sync as test files are moved,
+renamed, or deleted.
+
+## Short Term Mitigation
+
+While desired, there is no anticipated percentage of tests that will use the
+override functionality. As a result, an outside process could alert when there
+is some issue with the data. The feature set team will be responsible for that.
+
+If it becomes a problem to manage, one of the following future mitigation options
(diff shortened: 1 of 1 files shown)


rfcs/feature-set-metadata-file-to-web-platform-tests.md
```diff
@@ -0,0 +1,221 @@
...
+The RFC proposes the addition of a FEATURE_SET.yml metadata file which would
+enable the linkage between WPT and feature-set.
+
+# Proposed change
+
+The proposed change includes:
+
+1. Introduce a new metadata file, FEATURE_SET.yml
+2. Adjust the wpt-pr-bot to handle reviews of the changes
+3. Create a script to generate a manifest
```
Ms2ger, line 28: Match the order of your steps below, please

rfcs/feature-set-metadata-file-to-web-platform-tests.md
```diff
@@ -0,0 +1,221 @@
...
+[Meeting](https://github.com/web-platform-tests/wpt-notes/blob/master/minutes/2023-08-01.md#rfc-157---add-feature-meta-tag-to-web-platform-tests-meta-data),
+it was discussed that it should be a separate file. This section describes the structure of the file.
+
+This file is expected to be in the same places developers would expect a META.yml.
+
+### File examples
+
+```
+parser_version: v1
+apply_mode: RECURSIVE
```
Ms2ger, line 45: This does not seem to match the schema below

rfcs/feature-set-metadata-file-to-web-platform-tests.md
```diff
@@ -0,0 +1,221 @@
...
+META.yml. During the August 01, 2023 Monthly WPT
+[Meeting](https://github.com/web-platform-tests/wpt-notes/blob/master/minutes/2023-08-01.md#rfc-157---add-feature-meta-tag-to-web-platform-tests-meta-data),
+it was discussed that it should be a separate file. This section describes the structure of the file.
+
+This file is expected to be in the same places developers would expect a META.yml.
+
+### File examples
+
+```
+parser_version: v1
```
Ms2ger, line 44: Do we need this?

rfcs/feature-set-metadata-file-to-web-platform-tests.md
```diff
@@ -0,0 +1,221 @@
...
+opportunity to connect the WPT ecosystem to the feature-set catalog. By doing
+this, it would enable users of wpt.fyi to filter by feature-set grouping, which
+is similar to
+[the ability to filter by spec](https://github.com/web-platform-tests/wpt.fyi/issues/1489).
+The RFC proposes the addition of a FEATURE_SET.yml metadata file which would
+enable the linkage between WPT and feature-set.
+
+# Proposed change
+
+The proposed change includes:
```
Ms2ger, line 24: Presumably there's also a step where this data is integrated in wpt.fyi

Ms2ger (changes requested):
This seems reasonable

rfcs/feature-set-metadata-file-to-web-platform-tests.md
```diff
@@ -0,0 +1,221 @@
+# RFC #163: Add feature_set metadata file to web-platform-tests
+
+Author: @jcscottiii
+
+# Introduction
+
+The web-platform-tests (WPT) project is a valuable resource for testing web
+platform features and contains valuable metadata such as feature specs.
```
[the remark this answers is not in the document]
jcscottiii: You are correct that this is vague. I have clarified that it is spec [links](https://github.com/web-platform-tests/wpt/pull/12024) specifically.

rfcs/feature-set-metadata-file-to-web-platform-tests.md
jcscottiii: Technically, no. But I'd rather be safe than sorry. I'll remove this and we can always add it back if needed.

foolip:
https://github.com/web-platform-dx/web-features has been renamed to match the name of the NPM package, so I suggest renaming this to `web_features` or `features`.

rfcs/feature-set-metadata-file-to-web-platform-tests.md
jcscottiii: This will come as a separate RFC. Now that this RFC is moving to separate things (as opposed to #157), there might be a different way to integrate this into wpt.fyi instead of just modifying the current [metadata ingestion path](https://github.com/web-platform-tests/wpt.fyi/blob/9f00b3c20554d7926bbc4ca7877e5b17a59e406e/shared/metadata_util.go#L84).

jcscottiii:
Thanks for pointing that out @foolip ! I have renamed things.

foolip:
@jcscottiii did you forget to push changes? There's a lot of "feature-set" in the text still.

jcscottiii:
@foolip that's quite embarrassing of me  😅. Thanks for catching that.

I fixed that.

jcscottiii:
@jgraham I added the lint suggestion that you mentioned during the https://github.com/web-platform-tests/wpt-notes/blob/master/minutes/2023-09-05.md

foolip (approved):
Thanks @jcscottiii!

rfcs/feature-set-metadata-file-to-web-platform-tests.md
```diff
@@ -0,0 +1,220 @@
...
+is similar to
+[the ability to filter by spec links](https://github.com/web-platform-tests/wpt.fyi/issues/1489).
+The RFC proposes the addition of a WEB_FEATURE.yml metadata file which would
+enable the linkage between WPT and web-features.
+
+# Proposed change
+
+The proposed change includes:
+
+1. Introduce a new metadata file, WEB_FEATURE.yml
```
jgraham, line 26: I think maybe there was a miscommunication here (or I just changed my mind), but using the existing META.yml files is fine in my opinion. I'm just not keen for this to end up affecting the existing `MANIFEST.json` code. _But_ maybe a new file works better for the use case of not creating reviews for changes. So this is also fine.

rfcs/feature-set-metadata-file-to-web-platform-tests.md
```diff
@@ -0,0 +1,220 @@
...
+Typical example:
+
+```
+web_feature: subgrid
+```
+
+An example using all of the fields:
+
+```
+apply_mode: FORCE_RECURSIVE
```
jgraham, line 53: So generally these `apply_mode` / `override_mode` settings feel a bit over-engineered / action at a distance to me.

I'm also slightly concerned that this is different to the way that metadata ini files work and different to the way that the wpt-metadata repository works; it seems like each time we come across a similar problem we invent a slightly different solution.

Would it be enough to just have something like:

```
features:
   name: feature1
   files: *
```
For something that applies to all files (recursively) and then
```
features:
  name: feature2
  files:
    - test.html
    - subdir
 ```
 For features that only apply to specific files or directories, and then have a rule that we only inherit in cases that there's no more-specific rule in a subdirectory? And if we need multiple features per test (which I'm not sure if it's a requirement or not) we turn the `features` key into a list, and just say that if you want an inherited feature and a new one you need to repeat the inherited rule? Or inherit and add by default but have a magic feature name like `@reset` that blocks inherited features from applying (that's basically how this works in the expectation metadata).

**Review by jgraham — COMMENTED** [review]
So generally this seems OK, but I'd at least like to understand the requirements that drove the inheritance design.

Longer term, I'm worried about the duplication between this and wpt-metadata. wpt.fyi seems like the obvious UI to add feature metadata, even though (like interop labels) they don't correspond to test results. Indeed the lowest-risk way of prototyping this work and exploring requirements would be just adding labels to tests like `feature-subgrid`.

File: rfcs/feature-set-metadata-file-to-web-platform-tests.md (hunk 6 above)
Comment (jcscottiii) [reply rfcs/feature-set-metadata-file-to-web-platform-tests.md:26]: It was the only way I could think of to get around the unnecessary review pings issue. ([comment](https://github.com/web-platform-tests/rfcs/pull/157#discussion_r1268017940)) (aside from having wpt-pr-bot read the file directly before assigning reviewers.

File: rfcs/feature-set-metadata-file-to-web-platform-tests.md (hunk 7 above)
Comment (jcscottiii) [reply rfcs/feature-set-metadata-file-to-web-platform-tests.md:53]: I like the idea of switching `features` to a list
I can see getting rid of `override_mode`. But I think the `apply_mode` field makes it clearer.

Default case:

```yaml
features:
   - name: feature1
     files: **
```

List out files case:

```yaml
features:
- name: test-2
  files:
  - test-2-01.html
  - test-2-02.html
- name: test-1
  files:
  - test-1-01.html
```

Ignore parent features (or `@reset`)

```yaml
apply_mode: IGNORE_PARENT
features:
- name: reset-test-1
  files: "**"
```

Ignore parent features (or `@reset`) while listing out explicit files

```yaml
apply_mode: IGNORE_PARENT
features:
- name: reset-test-2
  files:
  - reset-test-2-01.html
- name: reset-test-1
  files:
  - reset-test-2-01.html
```

The jsonschema would look like: 

```json
{
    "$schema": "http://json-schema.org/draft-06/schema#",
    "$ref": "#/definitions/Main",
    "definitions": {
        "Main": {
            "type": "object",
            "additionalProperties": false,
            "properties": {
                "features": {
                    "type": "array",
                    "description": "List of features",
                    "items": {
                        "$ref": "#/definitions/FeatureEntry"
                    }
                },
                "apply_mode": {
                    "type": "string",
                    "anyOf": [
                    	{
                            "const": "IGNORE_PARENT",
                            "description": "Ignores features from previous parent directories."
                        }
                    ]
                }
            },
            "required": [
                "features"
            ],
            "title": "Main"
        },
        "FeatureEntry": {
            "type": "object",
            "additionalProperties": false,
            "properties": {
                "name": {
                    "type": "string",
                    "description": "The web feature key"
                },
                "files": {
                    "oneOf": [
                        {
                            "const": "**",
                            "description": "All files recursively",
                            "title": "Files"
                        },
                        {
                            "type": "array",
                            "items": {
                                "type": "string",
                                "description": "A specific file within the current directory"
                            }
                        }

                    ]
                }
            },
            "required": [
                "name",
                "files"
            ],
            "title": "FeatureEntry"
        }
    }
}
```

Thoughts @jgraham ? If this looks fine, I will readjust the RFC.

Also, I did some experimentation on my fork ([script](https://github.com/jcscottiii/rfcs/blob/experiment/test.py)) and [testdata directory](https://github.com/jcscottiii/rfcs/tree/experiment/testdata).

<details>
  <summary>Result</summary>
  
  ```
FeatureResult(_feature_tests_map_: OrderedDict([('root-feature', {'testdata/feature-policy/resources/feature-policy-picture-in-picture.html', 'testdata/css/grid/subgrid/subgrid1.html', 'testdata/html/html-test.html', 'testdata/picture-in-picture/pip1.html', 'testdata/css/grid/grid1.html', 'testdata/css/grid/grid2.html'}), ('picture-in-picture', {'testdata/picture-in-picture/pip1.html'}), ('grid', {'testdata/css/grid/subgrid/subgrid1.html', 'testdata/css/grid/grid1.html', 'testdata/css/grid/grid2.html'}), ('subgrid', {'testdata/css/grid/subgrid/subgrid1.html'}), ('reset-test-1', {'testdata/html/reset-test-1/reset-test-1-01.html', 'reset-test-2-01.html', 'testdata/html/reset-test-1/reset-test-1-02.html'}), ('reset-test-2', {'reset-test-2-01.html'})]))
  ```
</details>

foolip: A `features` list makes sense to me for cases where one directory has tests for multiple features, which is the case for https://wpt.fyi/results/css/selectors for example. In some of those cases it would be nice to just group them into directories, but I doubt we can *always* do it and thus avoid the problem entirely.

FWIW, I don't think it's an important use case to have one test be associated with multiple features, but if it's incidentally supported that's fine.

Regarding inheritance, I think we ought to support marking up all of https://wpt.fyi/results/css/css-grid *except* the masonry/ and subgrid/ subdirectories as being part of the grid feature.

jcscottiii: As discussed in today's wpt infra [meeting](https://github.com/web-platform-tests/wpt-notes/blob/master/minutes/2023-10-03.md), I added the new jsonschema from above into the RFC.

foolip:
@jcscottiii now that the file can list multiple features, can you rename it from `WEB_FEATURE.yml` to plural `WEB_FEATURES.yml`? This aligns with the package name `web-features` too :)

jcscottiii:
@foolip Thanks for the suggestion. Done!

foolip (approved):
Thanks @jcscottiii! I've looked this over once more. There are lots of ways that one could do the "exclude features from parent directories" part, and this approach seems pragmatic to me. (You could also imagine a system of includes/excludes, but then you need to both exclude tests from a parent feature and include them into a child feature, which isn't a big deal, but also not an improvement IMHO.)

jcscottiii:
@jgraham any thoughts on the revisions?

jgraham (approved):
I think this looks good enough to be worth prototyping.

foolip:
A first `WEB_FEATURES.yml` file was added in https://github.com/web-platform-tests/wpt/pull/42709. Thanks @jcscottiii for driving this RFC!

PR #165, opened 2023-09-23
RFC #165 expected-fail-message in test metadata

WeizhongX:
Chromium developers want to have a way to track the expected failure message, since a change in how a test fails can also be unintended and suggestive of a problem.

Made a proposal for discussion.

Bug: 42034

### DIFF
--- a/rfcs/expected_fail_message_tag_for_subtests.md
+++ b/rfcs/expected_fail_message_tag_for_subtests.md
@@ -61,7 +61,7 @@ maintanance headache, and are rejected consequently.
 ## Risks
 
 "expected-fail-message" will only take effect when it presents in the metadata
-file. when not presented, the test will behave exactly the same way as before.
+file. When not presented, the test will behave exactly the same way as before.
 It would be at developers' own discretion to use this feature or not. There is a
 chance this could be abused, but as the metadata file sits in each vendor's
 repository, the impact should be limited.
--- a/rfcs/expected_fail_message_tag_for_subtests.md
+++ b/rfcs/expected_fail_message_tag_for_subtests.md
@@ -1,4 +1,4 @@
-# RFC 158: Support expected-fail-message in test metadata for subtests
+# RFC 165: Support expected-fail-message in test metadata for subtests
 
 ## Summary
 
--- a/rfcs/expected_fail_message_tag_for_subtests.md
+++ b/rfcs/expected_fail_message_tag_for_subtests.md
@@ -46,7 +46,7 @@ additional member function to return "expected-fail-message",
 expected FAIL or PRECONDITION_FAILED to unexpected failures per the rule
 discribed above.
 
-A lint rule is also needed to prevent dangling "expected-fail-message", i.e. a
+A lint rule might be needed to prevent dangling "expected-fail-message", i.e. a
 subtest has expected-fail-message but FAIL or PRECONDITION_FAILED is not an
 expected result.
 
--- a/rfcs/expected_fail_message_tag_for_subtests.md
+++ b/rfcs/expected_fail_message_tag_for_subtests.md
@@ -1,0 +1,70 @@
+# RFC 158: Support expected-fail-message in test metadata for subtests
+
+## Summary
+
+Add "expected-fail-message" in test metadata so that we can track how
+a subtest fails.
+
+## Details
+
+In Chromium developers are using baselines to track expected output message
+for testharness tests. While the output message for a pass subtest is
+generally not interesting, the output message for a failed subtest tracks
+how a subtest fails, and a change in how a test fails could be unintended
+and suggestive of a problem. Chromium developers want to keep such capability
+if we want to use Wptrunner to run WPTs in Chromium.
+
+The suggestion is to support "expected-fail-message" tag for subtests. When
+present, it takes effect on expected FAIL or expected PRECONDITION_FAILED
+results. The test runner should check if the actual output message matches
+with the "expected-fail-message". When a mismatch is found, the test runner
+should turn the results to unexpected FAIL or unexpected PRECONDITION_FAILED
+respectively, otherwise the results continue to be the expected one.
+
+Wdspec tests can also have subtests. "expected-fail-message" should work the
+same way for an expected FAIL Wdspec subtest. Other than that, "expected-fail-message"
+should not have impact on any other scenarios.
+
+One example metadata file will look like below:
+```
+[event.html]
+  expected: OK
+  [Changing selection from handler: input event]
+    expected: FAIL
+
+  [Command createLink, value "": input event]
+    expected: [FAIL, PRECONDITION_FAILED, PASS]
+    expected-fail-message: "number of input events fired expectei 1 but got 0"
+```
+
+### Implementations
+
+The current implementation for test metadata parser is flexible enough to
+understand any addtional tag. The changes needed is 1) wpttest.Test to have one
+additional member function to return "expected-fail-message",
+2)testrunner.TestRunnerManager.test_ended to have additional logic to turn
+expected FAIL or PRECONDITION_FAILED to unexpected failures.
+
+A lint rule is also needed to prevent dangling "expected-fail-message", i.e. a
+subtest has expected-fail-message but FAIL or PRECONDITION_FAILED is not an
+expected result.
+
+
+### Alternatives Considered
+
+Various alternatives have been considered, e.g. writing a new test to assert the
+behavior a baseline wants to assert, or duplicate the test in question as a
+legacy web test so that baseline will continue work. Those all would be a
+maintanance headache, and are rejected consequently.
+
+## Risks
+
+"expected-fail-message" will only take effect when it presents in the metadata
+file. when not presented, the test will behave exactly the same way as before.
+It would be at developers' own discretion to use this feature or not. There is a
+chance this could be abused, but as the metadata file sits in each vendor's
+repository, the impact should be limited.
+
+This feature will not have any impact to the results presented at wpt.fyi, as
+this only changes if a FAIL is expected or not, but not the actual result
+itself.
--- a/rfcs/expected_fail_message_tag_for_subtests.md
+++ b/rfcs/expected_fail_message_tag_for_subtests.md
@@ -40,7 +40,7 @@ One example metadata file will look like below:
 ### Implementations
 
 The current implementation for test metadata parser is flexible enough to
-understand any addtional tag. The changes needed is 1) wpttest.Test to have one
+understand any additional tag. The changes needed is 1) wpttest.Test to have one
 additional member function to return "expected-fail-message",
 2)testrunner.TestRunnerManager.test_ended to have additional logic to turn
 expected FAIL or PRECONDITION_FAILED to unexpected failures per the rule
--- a/rfcs/expected_fail_message_tag_for_subtests.md
+++ b/rfcs/expected_fail_message_tag_for_subtests.md
@@ -43,7 +43,8 @@ The current implementation for test metadata parser is flexible enough to
 understand any addtional tag. The changes needed is 1) wpttest.Test to have one
 additional member function to return "expected-fail-message",
 2)testrunner.TestRunnerManager.test_ended to have additional logic to turn
-expected FAIL or PRECONDITION_FAILED to unexpected failures.
+expected FAIL or PRECONDITION_FAILED to unexpected failures per the rule
+discribed above.
 
 A lint rule is also needed to prevent dangling "expected-fail-message", i.e. a
 subtest has expected-fail-message but FAIL or PRECONDITION_FAILED is not an
--- a/rfcs/expected_fail_message_tag_for_subtests.md
+++ b/rfcs/expected_fail_message_tag_for_subtests.md
@@ -7,7 +7,7 @@ a subtest fails.
 
 ## Details
 
-In Chromium developers are using baselines to track expected output message
+In Chromium developers are using [baselines](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/external/wpt/editing/event-expected.txt) to track expected output message
 for testharness tests. While the output message for a pass subtest is
 generally not interesting, the output message for a failed subtest tracks
 how a subtest fails, and a change in how a test fails could be unintended
@@ -46,10 +46,13 @@ additional member function to return "expected-fail-message",
 expected FAIL or PRECONDITION_FAILED to unexpected failures per the rule
 discribed above.
 
+How to add "expected-fail-message" to test metadata is out of scope of this RFC.
+Usually this tag can be added manually by developers after carefully review the
+test results and implementations.
+
 A lint rule might be needed to prevent dangling "expected-fail-message", i.e. a
 subtest has expected-fail-message but FAIL or PRECONDITION_FAILED is not an
 expected result.
-
 
 ### Alternatives Considered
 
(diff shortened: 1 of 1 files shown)


rfcs/expected_fail_message_tag_for_subtests.md
```diff
@@ -0,0 +1,71 @@
...
+
+  [Command createLink, value "": input event]
+    expected: [FAIL, PRECONDITION_FAILED, PASS]
+    expected-fail-message: "number of input events fired expectei 1 but got 0"
+```
+
+### Implementations
+
+The current implementation for test metadata parser is flexible enough to
+understand any addtional tag. The changes needed is 1) wpttest.Test to have one
```
Ms2ger, line 43:
```suggestion
understand any additional tag. The changes needed is 1) wpttest.Test to have one
```

Ms2ger (approved):
As long as there's clear documentation that this places no burden on other developers - in particular that this will not be used to argue against changes that affect the error messages, I don't think I have particular objections

rfcs/expected_fail_message_tag_for_subtests.md
```diff
@@ -0,0 +1,71 @@
+# RFC 165: Support expected-fail-message in test metadata for subtests
+
+## Summary
+
+Add "expected-fail-message" in test metadata so that we can track how
+a subtest fails.
```
foolip, line 6: Do we also need to track harness error messages?

rfcs/expected_fail_message_tag_for_subtests.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Details
+
+In Chromium developers are using baselines to track expected output message
```
foolip, line 10: Can you link an example with a permalink? I expect not everyone knows that a baseline is a -expected.txt file?

rfcs/expected_fail_message_tag_for_subtests.md
```diff
@@ -0,0 +1,71 @@
...
+for testharness tests. While the output message for a pass subtest is
```
foolip, line 11: Is there even a message anywhere when passing?

rfcs/expected_fail_message_tag_for_subtests.md
```diff
@@ -0,0 +1,71 @@
...
+### Implementations
+
+The current implementation for test metadata parser is flexible enough to
+understand any additional tag. The changes needed is 1) wpttest.Test to have one
+additional member function to return "expected-fail-message",
+2)testrunner.TestRunnerManager.test_ended to have additional logic to turn
+expected FAIL or PRECONDITION_FAILED to unexpected failures per the rule
+discribed above.
+
+A lint rule is also needed to prevent dangling "expected-fail-message", i.e. a
```
foolip, line 49: Do we lint ini files at all now? Or is this a lint that will have to be in Chromium?

rfcs/expected_fail_message_tag_for_subtests.md
```diff
@@ -0,0 +1,71 @@
...
+
+Various alternatives have been considered, e.g. writing a new test to assert the
+behavior a baseline wants to assert, or duplicate the test in question as a
+legacy web test so that baseline will continue work. Those all would be a
+maintanance headache, and are rejected consequently.
+
+## Risks
+
+"expected-fail-message" will only take effect when it presents in the metadata
+file. when not presented, the test will behave exactly the same way as before.
```
foolip, line 64:
```suggestion
file. When not presented, the test will behave exactly the same way as before.
```

rfcs/expected_fail_message_tag_for_subtests.md
WeizhongX: I included this for completeness. We do lint ini files in Chromium but not at upstream now. As we are going to use test expectations/baselines in Chromium, we won't lint this either in Chromium. Reworded this to "A lint rule might be needed ...".

https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/tools/blinkpy/tool/commands/lint_wpt.py;drc=8185de7bf7ac655feddf94e7db7d544d6abb3e5c;l=67

rfcs/expected_fail_message_tag_for_subtests.md
WeizhongX: Asked this in crbug.com/1481730.

foolip: I think the answer should probably be yes. When there's a harness error, the exception message is often enough to figure out the problem, so having the importer write that to the repo by default would be convenient, and is the current behavior with -expected.txt.

rfcs/expected_fail_message_tag_for_subtests.md
WeizhongX: We do not include the message in -expected.txt for a pass subtest, but the field is always there from Wptrunner's point of view. Any suggestion for rewording if needed?

rfcs/expected_fail_message_tag_for_subtests.md
WeizhongX: Include the harness error message in actual output would be convenient for devs to know what is wrong, but do we also want to use that to prevent regression? If that is the case, we would need another key "expected-harness-error-message"? Is there a way we could consolidate the two into one?

If we do not need to track regression, we can still include the error message in actual output, but convert harness error into a test level expected error when converting baselines to ini files.

rfcs/expected_fail_message_tag_for_subtests.md
```diff
@@ -0,0 +1,71 @@
...
+  expected: OK
+  [Changing selection from handler: input event]
+    expected: FAIL
+
+  [Command createLink, value "": input event]
+    expected: [FAIL, PRECONDITION_FAILED, PASS]
+    expected-fail-message: "number of input events fired expectei 1 but got 0"
+```
+
+### Implementations
```
jgraham, line 40: Will you also update the manifest update code to allow updating the `expected-fail-message` metadata? If so this needs to be an optional flag so that only people who are using this code get the additional keys.

jgraham (commented):
I am somewhat concerned that this puts additional constraints on test authors. For example how does this work with a message like `expected 250 got 312` where the latter can vary? Does the existing baseline setup allow some kind of wildcards? But once we have wildcards we can't auto-update the metadata (or at least can only update to literals, which seems insufficient).

foolip:
> I am somewhat concerned that this puts additional constraints on test authors. For example how does this work with a message like `expected 250 got 312` where the latter can vary? Does the existing baseline setup allow some kind of wildcards?

Currently in Chromium that would lead to a flaky test which needs to be disabled in `TestExpectations`, or if the difference is just between platforms, we have platform-specific `-expected.txt` baselines. (There are no wildcards.) However, this all seems surprisingly rare. By keeping the same basic scheme, it won't add new constraints on Chromium developers or other test authors.

jgraham:
> However, this all seems surprisingly rare

That is _very_ surprising to me, because I can think of multiple features with a lot of tests with failure messages like that. And certainly the audio tests that dump floating point numbers into the test names (despite the fact that this is a violation of the testsuite semantics) cause no end of problems for us.

Anyway, my point of view is that this is broadly up to you in the sense that if Chromium want this feature as proposed then it's fine to go ahead and implement it. My only concern is that it shouldn't later be used as a reason to make test failure messages less useful / more consistent (e.g. by removing variable data from the output). If you do find yourselves wanting that you should instead figure out a way to relax these conditions (e.g. having a way to not update the annotations for tests where the output is not consistent, or a way to specify regex to match for the expected output.

Another concern is that people start treating the test failure message as an actual part of the API so instead of actually writing subtests they just dump a bunch of textual data in the failure message and assume that's good enough to distinguish which cases passed and which failed. I hope the risk of that is small since it hasn't happened so far, but if it does happen I think we'd need to consider how to prevent it.

WeizhongX:
> My only concern is that it shouldn't later be used as a reason to make test failure messages less useful / more consistent (e.g. by removing variable data from the output)

"expected-fail-message" is an optional field, so when not used there is not any requirement on the fail messages.

> Another concern is that people start treating the test failure message as an actual part of the API so instead of actually writing subtests they just dump a bunch of textual data in the failure message

There is such risk, but failure message is implementation specific, so tests written this way would not work on a different browser thus will be surfaced pretty soon?

As I said in the bug(42034), we are already running WPTs this way in Chromium, and the error message is mandatory, so it is unlikely to deteriorate the status.

jgraham:
> "expected-fail-message" is an optional field, so when not used there is not any requirement on the fail messages.

The concern is that people intentionally sacrifice usefulness for consistency e.g replacing `assert_greater_than(x, 200)` with assert_true(x > 200)` because the latter produces a message that's independent of the value of `x` whereas the former embeds useful — but possibly variable — data into the message.

This is more of a concern when the failure message looks like an official part of the wpt API vs when it's clearly only part of vendor embeddings.

WeizhongX:
> > "expected-fail-message" is an optional field, so when not used there is not any requirement on the fail messages.
> 
> The concern is that people intentionally sacrifice usefulness for consistency e.g replacing `assert_greater_than(x, 200)` with `assert_true(x > 200)` because the latter produces a message that's independent of the value of `x` whereas the former embeds useful — but possibly variable — data into the message.
> 
> This is more of a concern when the failure message looks like an official part of the wpt API vs when it's clearly only part of vendor embeddings.

Our automation tool will not make change to the test. It will simply mark the test as flaky. Will developers favor `assert_true` over `assert_greater_than`? I think it is up to developer's decision. But as I said we are already using baselines in chromium, so it might be that developers are already using that technology to deflake a test. But as `expected-fail-message` is optional, we now give them a second option by continue to use `assert_greater_than` without the `expected-fail-message` in the ini file. If the developers think the error message is not interesting, they should simply not using `expected-fail-message`, instead of make a change to the test?

The intention of `expected-fail-message` is to have a way to assert behavior. In that sense `assert_greater_than` has more information and would be prefered if they want to use `expected-fail-message`? This matches with the case I mentioned in bug/42034

rfcs/expected_fail_message_tag_for_subtests.md
WeizhongX: Done

rfcs/expected_fail_message_tag_for_subtests.md
WeizhongX: Added
```
How to add "expected-fail-message" to test metadata is out of scope of this RFC.
Usually this tag can be added manually by developers after carefully review the
test results and implementations.
```

I don't want to make this process automatically, in case it could be abused.

jgraham: OK, but I don't understand how that will work with your sync. If you import a bunch of tests which have changed their failure messages will you have to manually update any `expected-fail-message` data?

WeizhongX: We will continue use baselines, and when running test, recreate test metadata from the baselines.

So for us, we will use `expected-fail-message` if there is a fail message from the baseline.

mfreed7:
> I am somewhat concerned that this puts additional constraints on test authors. For example how does this work with a message like `expected 250 got 312` where the latter can vary? Does the existing baseline setup allow some kind of wildcards? But once we have wildcards we can't auto-update the metadata (or at least can only update to literals, which seems insufficient).

This is precisely the thing we'd like to be able to test for. If the test "usually" says `expected 250 got 312` but as the result of some code change starts saying `expected 250 got 313` then we'd like to catch that and triage it appropriately. I don't think it places additional constraints on test authors - they should keep doing this! It's helpful. Note that in both cases this will still be a failure, just a failure *change* that can be detected.

mfreed7:
> Anyway, my point of view is that this is broadly up to you in the sense that if Chromium want this feature as proposed then it's fine to go ahead and implement it. My only concern is that it shouldn't later be used as a reason to make test failure messages less useful / more consistent (e.g. by removing variable data from the output). If you do find yourselves wanting that you should instead figure out a way to relax these conditions (e.g. having a way to not update the annotations for tests where the output is not consistent, or a way to specify regex to match for the expected output.

Totally agree with this - adding this capability shouldn't be used as a reason to change test output text.

rfcs/expected_fail_message_tag_for_subtests.md
mfreed7: I'm not against adding a way to test against "expected" harness errors, but this is certainly a much less common case. Typically/ideally these are truly harness errors, not something failing in a way we'd like to continue testing for. I could be proven wrong with examples. But if I'm right, this can be lower priority.

rfcs/expected_fail_message_tag_for_subtests.md
```diff
@@ -0,0 +1,74 @@
...
+additional member function to return "expected-fail-message",
+2)testrunner.TestRunnerManager.test_ended to have additional logic to turn
+expected FAIL or PRECONDITION_FAILED to unexpected failures per the rule
+discribed above.
+
+How to add "expected-fail-message" to test metadata is out of scope of this RFC.
+Usually this tag can be added manually by developers after carefully review the
+test results and implementations.
+
+A lint rule might be needed to prevent dangling "expected-fail-message", i.e. a
```
jgraham, line 53: Note that we currently don't have any lints at all for metadata. If you want to write such lints I'm of course happy for you to do so, but perhaps it should be a separate discussion.

rfcs/expected_fail_message_tag_for_subtests.md
```diff
@@ -0,0 +1,74 @@
...
+### Alternatives Considered
+
+Various alternatives have been considered, e.g. writing a new test to assert the
+behavior a baseline wants to assert, or duplicate the test in question as a
+legacy web test so that baseline will continue work. Those all would be a
+maintanance headache, and are rejected consequently.
+
+## Risks
+
+"expected-fail-message" will only take effect when it presents in the metadata
```
jgraham, line 66: I would ask that the documentation for this feature captures the discussion in this issue around not writing tests in a way that's designed to produce consistent output, but instead it being up to specific consumers to only use this property in cases where they have reason to believe the output will be consistent anyway.

jgraham:
> If the test "usually" says expected 250 got 312 but as the result of some code change starts saying expected 250 got 313 then we'd like to catch that and triage it appropriately. 

The cases I was thinking of are things like timing tests where the output is inherently unstable (e.g. if some timing property is supposed to be 0 but due to a bug you set it to some actual time interval, the actual interval is going to be highly dependent on the specific environmental conditions and so vary from run to run). With e.g. layout tests it seems likely that things are more stable, but even for fuzzy annotations we allow a range of values.

As I've said, I'm not confident that this doesn't create pressure to elide such information from tests if the implementation is both doing the failure message comparison, and happens to fail the test at the point of creation. But if you're already doing this anyway, and are going to continue to do so, it doesn't seem obviously worse for it to be in wptrunner itself.

rfcs/expected_fail_message_tag_for_subtests.md
WeizhongX: I discussed with foolip@ offline, and he agreed this can be lower priority, so resolve it now.

rfcs/expected_fail_message_tag_for_subtests.md
WeizhongX: Thanks, let me remove this line, and we can have a separate RFC for this.

We do have some lints for metadata in Chromium now, but not exactly sure how much that can be applied to upstream.

jgraham (approved):

PR #169, opened 2023-10-18
RFC 169: WebRTC Echo server

fippo:
described in
  https://lists.w3.org/Archives/Public/public-webrtc/2021Feb/0043.html
  https://www.youtube.com/watch?v=nPkyvxsKed8 (meeting)

Draft implementation:
  https://github.com/web-platform-tests/wpt/pull/36487

### DIFF
--- a/rfcs/webrtc_test_server.md
+++ b/rfcs/webrtc_test_server.md
@@ -1,4 +1,5 @@
-# RFC xx: WebRTC Test Server
+# RFC 169: WebRTC Test Server
+
 
 ## Summary
 
--- a/rfcs/webrtc_test_server.md
+++ b/rfcs/webrtc_test_server.md
@@ -1,0 +1,39 @@
+# RFC xx: WebRTC Test Server
+
+## Summary
+
+Add a [WebRTC](https://datatracker.ietf.org/doc/html/rfc8834) server to wpt. The server is used to test the [WebRTC Web APIs](https://w3c.github.io/webrtc/). The server requires Python 3.
+
+The server uses a WebSocket connection to negotiate the WebRTC connection.
+It terminates STUN, DTLS and SRTP and forwards unencrypted RTP or RTCP packets to the browser.
+This allows the tests to receive and inspect any packets received.
+
+Future versions may use WebTransport instead of WebSockets for signaling and transport of packets to the client and allow the client to send packets over the WebSocket to be forwarded
+via WebRTC.
+
+## Details
+
+### Implementation
+
+The WebRTC test server is built on top of `aiortc`, similar to how the WebTransport/H3 test server is based on `aioquic`.
+
+The server will be implemented under `tools/webrtc` directory.
+
+### `wptserve` integration
+
+When the WebRTC test server runs as a part of `wptserve`, `$ROOT` will be the same as `$WPT_ROOT`.
+
+`wptserve` runs the WebRTC test server in a way similar to [pywebsocket](https://github.com/web-platform-tests/wpt/blob/246a32576020cb9c4241b7cfbc296f92d944ff6b/tools/serve/serve.py#L713):
+`wptserve` launches the server as a daemon and passes relevant configurations (port number, root directory etc) to the server. The daemon is monitored by `wptserve` like other daemons.
+
+### Dependencies
+
+As of writing this RFC, the only dependency is `aiortc`.
+
+## Risks
+
+Risks are similar to [RFC #42](https://github.com/web-platform-tests/rfcs/blob/master/rfcs/quic.md#risks).
+
+The WebRTC test server itself is standalone and the maintenance cost of the code will be low from the `wptserve`'s perspective.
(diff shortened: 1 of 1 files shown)


fippo:
(obviously the RFC is borrowed from the h3 server as is the implementation ;-))

rfcs/webrtc_test_server.md
```diff
@@ -0,0 +1,39 @@
+# RFC xx: WebRTC Test Server
```
gsnedders, line 1:
```suggestion
# RFC 169: WebRTC Test Server

```

rfcs/webrtc_test_server.md
```diff
@@ -0,0 +1,40 @@
...
+### Dependencies
+
+As of writing this RFC, the only dependency is `aiortc`.
+
+## Risks
+
+Risks are similar to [RFC #42](https://github.com/web-platform-tests/rfcs/blob/master/rfcs/quic.md#risks).
+
+The WebRTC test server itself is standalone and the maintenance cost of the code will be low from the `wptserve`'s perspective.
+The WebRTC team at Google and Microsoft will be responsible for maintaining the integration point and the server.
```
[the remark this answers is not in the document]
foolip: Thanks for confirming!

foolip (approved):

rfcs/webrtc_test_server.md
```diff
@@ -0,0 +1,40 @@
...
+`wptserve` runs the WebRTC test server in a way similar to [pywebsocket](https://github.com/web-platform-tests/wpt/blob/246a32576020cb9c4241b7cfbc296f92d944ff6b/tools/serve/serve.py#L713):
+`wptserve` launches the server as a daemon and passes relevant configurations (port number, root directory etc) to the server. The daemon is monitored by `wptserve` like other daemons.
+
+### Dependencies
+
+As of writing this RFC, the only dependency is `aiortc`.
+
+## Risks
+
+Risks are similar to [RFC #42](https://github.com/web-platform-tests/rfcs/blob/master/rfcs/quic.md#risks).
```
jgraham, line 36: Does this involve new binary dependencies? I assume it does, but it would help to be specific here since those require special handling, especially in vedor repos where we can't depend on pypi. They also add some maintenance risk (as we've seen with aioquic not updating its releases for newer Python versions)

fippo: [Python bindings for libsrtp](https://github.com/aiortc/pylibsrtp) and google-crc32c should be the only new dependencies I think. @jlaine might know more.

gsnedders: https://github.com/aiortc/aiortc/blob/0ecc5379762883fbf6454d4633b948c2ad50dec8/pyproject.toml#L28-L38 shows quite a lot of (direct) Python dependencies:

```
dependencies = [
    "aioice>=0.9.0,<1.0.0",
    "av>=9.0.0,<12.0.0",
    "cffi>=1.0.0",
    "cryptography>=2.2",
    'dataclasses; python_version < "3.7"',
    "google-crc32c>=1.1",
    "pyee>=9.0.0",
    "pylibsrtp>=0.5.6",
    "pyopenssl>=23.1.0",
]
```

I think `av` is the dependency that worries me most here—especially insofar as it has a pretty major dependency (`ffmpeg`—which is a dependency likely to make legal departments nervous given the patent status of many codecs it supports—and is also just a pretty sizeable project which is more likely to cause issue on new platforms).

(There's a few other notable new dependencies: `aiortc` itself depends on opus and libvpx, and `pylibstrp` depends on libsrtp, but these are probably less problematic)

alvestrand: > Are we actually planning on doing any testing that actually requires media codecs?

Yes.

The initial proposals (passing the RTP packets back to the test for analysis) would not need media codecs on the server, but once we have experience with those, running media codecs on the server is a definite option.
In order to do the tests, we have to negotiate use of media codecs; I have no idea whether it's possible to build aiortc in such a way that it can claim to have media codecs but actually not have them.

The argument for doing these tests in wpt is that we would get wpt's infrastructure for tracking and sharing tests with a relatively small investment in engineering on the webrtc support side (by using aiortc). If we start asking to reengineer aiortc in order to integrate it, the value of this approach starts dropping.

fippo: Now forking to remove "av" as a dependency is possible: https://github.com/aiortc/aiortc/compare/main...fippo:aiortc:no-av?expand=1
but that gets to a point where the maintenance cost will add up.

Just running this in Chromium's CI as the tests will be testing low-level libWebRTC behavior which is going to be largely identical in consuming browsers like Safari.

fippo:
What is the next step here?

jgraham:
I think the next step is for @gsnedders to give some feedback on what can work from WebKit's point of view.

alvestrand:
Ping - @gsnedders was asked to comment.

jcscottiii:
We talked about this today during the core meeting. @gsnedders will take a look again.

PR #173, opened 2023-11-24
RFC 173 - Use @page for print refests

jgraham:


jwatt (approved):
Yes, please.

foolip (approved):
I've confirmed with @mstensho that this makes sense for Chrome.

foolip:
@past FYI, I approved this.

mstensho:
Now is a good time to ping on https://github.com/web-platform-tests/wpt/issues/40788 - we have https://web-platform-tests.org/writing-tests/print-reftests.html which says nothing about the 0.5 inch margins. Can this documentation be updated, or removed? This is the documentation I got to when searching, and I wasn't even aware of https://github.com/web-platform-tests/rfcs/blob/master/rfcs/print_test.md .

Ruben-VandeVelde (approved):

PR #174, opened 2023-11-29
RFC 174: Add testdriver.js support for the Get Window Rect command

javifernandez:
Add testdriver.js support for the [WebDdriver 'Get Window Rect' command](https://www.w3.org/TR/webdriver/#get-window-rect).

### DIFF
--- a/rfcs/get_window_rect.md
+++ b/rfcs/get_window_rect.md
@@ -1,0 +1,16 @@
+# RFC 174: Add testdriver.js support for the Get Window Rect command
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


foolip:
We have an exemption in the [RFC process](https://github.com/web-platform-tests/rfcs#when-to-use-the-rfc-process) for things like this:

> Extending testdriver.js with a method that closely matches a [WebDriver](https://w3c.github.io/webdriver/) endpoint. To notify maintainers of testdriver.js vendor integration, label the pull request `testdriver.js`. (This exemption was introduced by [RFC 127](https://github.com/web-platform-tests/rfcs/blob/master/rfcs/rfc_exemption_testdriver_method.md).)

So, no paperwork needed!

PR #177, opened 2023-12-18
RFC 177 - fetch_json and other ShadowRealm facilities

ptomato:
Request for feedback on the APIs proposed in
https://github.com/web-platform-tests/wpt/pull/43639 adding some facilities for executing existing tests inside ShadowRealm global contexts.

### DIFF
--- a/rfcs/fetch_json_et_al.md
+++ b/rfcs/fetch_json_et_al.md
@@ -1,0 +1,122 @@
+# RFC _NNN_: `fetch_json` and other `ShadowRealm` facilities
+
+## Summary
+
+This follows up [RFC 107][rfc-107].
+[`ShadowRealm`s][proposal] are a new sandboxing primitive, currently in
+the process of being integrated into relevant Web specifications.
+`ShadowRealm` environments resemble vanilla JS shell environments, but
+do have some Web interfaces available, such as `URL`.
+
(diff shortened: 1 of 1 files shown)


rfcs/fetch_json_et_al.md
```diff
@@ -0,0 +1,122 @@
...
+
+The most notable change is a `fetch_json` function to allow importing
+test data from JSON files across the sandboxing boundary.
+
+## Details
+
+The [pull request][pr] comprises four commits, which I'll explain in
+detail in this section.
+
+### Document self.GLOBAL.isShadowRealm()
```
foolip, line 25: If you want to land this fix right away and drop it from this RFC that would be fine.

foolip (approved):

rfcs/fetch_json_et_al.md
ptomato: Thanks, I may split it out if I get a chance before the RFC finishes.

ptomato:
Just checking in about the process — can we consider this accepted or is it wise to wait another week to allow folks to come back from winter holidays?

ptomato:
@foolip At this point, can we land this?

foolip:
@ptomato sure thing, per our process enough time has passed, this RFC has passed.

PR #179, opened 2024-01-25
RFC 179: Remove support for Internet Explorer and Microsoft Edge Legacy

mwjacksonmsft:
Remove support for running WPT tests in Internet Explorer and Microsoft Edge Legacy.

### DIFF
--- a/rfcs/remove_ie_edge_legacy.md
+++ b/rfcs/remove_ie_edge_legacy.md
@@ -1,0 +1,32 @@
+# RFC 179: Remove support for Internet Explorer and Microsoft Edge Legacy
+
+### Summary
+
+Remove support for running WPT tests in Internet Explorer and Microsoft Edge
+Legacy.
+
+### Details
+
+Support for the Microsoft Edge Legacy desktop application ended on March 9,
+2021[1], and support for the Internet Explorer desktop application ended on
+June 15, 2022[2].
+
+The proposal is to remove the `edge`, `edge_webdriver`, and `ie` options from
+the product parameter to `wpt run`, along with all related code.
+
+After removal, the remaining supported products would be:
+
+```
+[--product {android_weblayer,android_webview,chrome,chrome_android,chrome_ios,
+chromium,content_shell,edgechromium,firefox,firefox_android,safari,sauce,servo,
+servodriver,opera,webkit,webkitgtk_minibrowser,wktr,epiphany,ladybird}]
+```
+
+[1] https://blogs.windows.com/msedgedev/2021/03/09/microsoft-edge-legacy-end-of-support/
+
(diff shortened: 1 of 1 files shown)


gsnedders:
cc @jimevans 

IIRC, IEDriver is still used for IE Mode in Edge, which is still supported. We probably want some way to still test IEDriver?

Edge Legacy is probably a smaller deal here, and probably has code-rotted and is probably of little interest now.

rfcs/remove_ie_edge_legacy.md
```diff
@@ -0,0 +1,32 @@
...
+June 15, 2022[2].
+
+The proposal is to remove the `edge`, `edge_webdriver`, and `ie` options from
+the product parameter to `wpt run`, along with all related code.
+
+After removal, the remaining supported products would be:
+
+```
+[--product {android_weblayer,android_webview,chrome,chrome_android,chrome_ios,
+chromium,content_shell,edgechromium,firefox,firefox_android,safari,sauce,servo,
```
foolip, line 21: This doesn't need to be decided now, would make sense to wait for a few months after this RFC and then rename `edgechromium` to `edge` to simplify matters.

foolip (approved):

Ms2ger (commented):
Is the scope of this RFC purely `wpt run`? That seems reasonable, but I'd be more hesitant if it was read as actively making it harder to run tests manually in historical browsers.

A section on the benefits of this change would be useful as well. Is it just removing cruft or is this support blocking other work?

mwjacksonmsft:
@Ms2ger - The proposal is to just clean up and remove unsupported Microsoft browsers. I've put together a draft PR (https://github.com/web-platform-tests/wpt/pull/44233) that shows the remove of the Microsoft Edge Legacy browser.

mwjacksonmsft:
@gsnedders I don't believe IEDriver uses anything from WPT tests to set up and use. I also don't believe that you can currently run WPT tests in IEMode via `wpt run`. If we wanted to add support, then I'd suggest we'd want the product name to be 'edge-iemode' (or similar) to avoid confusion with the IE-app.

foolip:
One week has passed and there's no substantive disagreement, so I'll go ahead and merge this.

If it turns out that some of this code is still used by someone, then of course I think we should revisit.

foolip (approved):

foolip:
The removals have happened in https://github.com/web-platform-tests/wpt/pull/44233 + https://github.com/web-platform-tests/wpt/pull/44344.

Thanks @mwjacksonmsft for cleaning this up! As I said, I hope that after some months we can rename the current Edge to just "edge" so it's more intuitive for users of `./wpt run`.

Issue #172, opened 2023-10-27
Remove tests that rely on undefined testing APIs


gsnedders:
WPT contains a variety of tests that rely on MojoJS to run in Chromium-based browsers.

[Chromium's own WPT docs](https://chromium.googlesource.com/chromium/src/+/main/docs/testing/web_platform_tests.md#mojojs) say:

> Some specs may define testing APIs (e.g. [WebUSB](https://wicg.github.io/webusb/test/)), which may be polyfilled with internal API like [MojoJS](https://chromium.googlesource.com/chromium/src/+/main/mojo/public/js/README.md). MojoJS is only allowed in WPT for this purpose. Please reach out to [blink-dev@chromium.org](mailto:blink-dev@chromium.org) before following the process below for adding a new test-only API: [.…]

However, we have many cases where things are "polyfilled" via MojoJS where the spec doesn't define the testing API, and nor has there been any attempt within WPT to document what the testing API the tests expect is. (Furthermore, https://groups.google.com/a/chromium.org/g/blink-dev/search?q=mojojs makes it look like basically nobody has _ever_ followed those steps, as nobody has done the "reach out" step.)

The tests are, thus, practically Chromium-specific, short of someone reverse-engineering what the testing API is.

For example, the Shape Detection tests that were added to WPT in https://github.com/web-platform-tests/wpt/commit/dfc0080d132ac25c4cf824a25735203b3a2b6853 rely on [/shape-detection/resources/shapedetection-helpers.js](https://github.com/web-platform-tests/wpt/blob/e83deaa539b2c3d5d02386b8b7ef0ef10079fbe5/shape-detection/resources/shapedetection-helpers.js) to load testing APIs, except nowhere in https://wicg.github.io/shape-detection-api/ (or any other document) are `FaceDetectionTest`, `BarcodeDetectionTest`, or `TextDetectionTest` actually defined.

Notwithstanding questions of whether any such testing API would ship in the browser, this means wpt.fyi will forever show Safari as failing many of the shape detection tests, even if the feature flag were enabled by default, unless we invest in reverse-engineering Chromium's testing API. Part of the very goal of standards and WPT is to avoid vendors having to invest in reverse-engineering one-another.

In extreme, [/serial/README.md](https://github.com/web-platform-tests/wpt/blob/e83deaa539b2c3d5d02386b8b7ef0ef10079fbe5/serial/README.md) even notes:

> there is no separate specification of the API other than the tests themselves and the Chromium implementation

[...]

PR #180, opened 2024-02-02
RFC 180: Tentative automation APIs with testdriver-tentative.js

foolip:


gsnedders (changes requested):
This feels like, practically, another way to just add browser-specific tests into WPT. (See also: #172.)

As I wrote in #172:

> WPT cannot be a dumping ground for inherently vendor-specific tests, written with little-to-no regard for other vendors using them, as this undermines the very premise of it being a shared test suite. It is unreasonable and unhelpful for Chromium engineers to export tests which other vendors must reverse-engineer the testing API to gain any benefit from. Chromium can maintain these tests in their own, non-exported tree, and do so until such point as they define what the testing APIs actually are.
>
> As such, we should remove all tests that rely on undefined testing APIs from WPT.

The history here tends to suggest that developers would implement some "tentative" automation solution (regardless of whether it's MojoJS or a tentative testdriver API), but would never put in any work in ensuring those tests are usable cross-browser, practically putting all the work on any other interested implementer (who either has to reverse-engineer the API, or just write an entirely new test suite—either way, little value has been derived from having the test suite in WPT).

jgraham:
I think usually with `tentative` tests, we expect there to be a standards document, and the test is `tentative` because the pass condition presumes one outcome of an ongoing technical discussion.

Allowing a class of tests that depend on something that's entirely nonstandard — moreso even than single-vendor WICG proposals — does seem like a step change compared to what's accepted today.

I agree with the underlying (initial) goal of making the accessibility tree available to automation. But I don't see how it benefits us to start with tests that depend on the current Chromium behaviour, rather than starting with some agreement over the general shape the API should take (and FWIW I strongly suspect that standardising "an API that returns the current a11y tree" is not the hard part here; it's making a11y trees from different implementations similar enough — or even standard — so that you can write meaningful tests covering multiple browsers).

PR #181, opened 2024-02-16
Add RFC XXX - URLManifestItem.url shouldn't contain non-URL-unit characters

gsnedders:
[Rendered](https://github.com/gsnedders/rfcs/blob/valid-item-urls/rfcs/valid-item-urls.md)


gsnedders:
(There's also the variant of this RFC where all we do is `url.replace(" ", "%20")` which avoids the "space separated URLs" problem. But it still carries the same risks.)

rfcs/valid-item-urls.md
```diff
@@ -0,0 +1,40 @@
...
+./css/CSS2/syntax/support/'green block.png
+./html/semantics/document-metadata/the-meta-element/pragma-directives/attr-meta-http-equiv-refresh/support/url foo
+./html/canvas/offscreen/path-objects/2d.path.roundrect.1.radius.dompoint.single argument.worker.js
+./html/canvas/offscreen/path-objects/2d.path.roundrect.1.radius.dompoint.single argument.html
+./html/canvas/element/path-objects/2d.path.roundrect.1.radius.dompoint.single argument.html
+```
+
+Only the last three of these are test files, which generate three test items.
+
+One question here is whether we should disallow test file paths that allow spaces, or whether we should just escape these in the URL.
```
Ms2ger, line 32:
```suggestion
One question here is whether we should disallow test file paths that contain spaces, or whether we should just escape these in the URL.
```

foolip:
> One question here is whether we should disallow test file paths that contain spaces, or whether we should just escape these in the URL.

Given how few such files there are, let's just disallow them.

jgraham (commented):
Storing URLs (and therefore test ids) in url-encoded form does make sense.

Migrating expectation data is a challenge. I wonder if for the wpt expectation files (which handle spaces fine) we could just unencode the test id before using the manifest? That makes it more likely that the manifest will match whatever people write in the `<meta name=varaint>`. Alternatively if we're going to change to using the URL encoded form everywhere, I think we should have a lint to ensure we use it consistently (maybe we should have a lint either way, so that we don't get `%20` in some places and ` ` in another).

I don't think banning spaces in test names is worthwhile.

PR #193, opened 2024-04-30
Navigate to about:blank after loading a crashtest

jgraham:
This helps catch a class of delayed crashes that are currently difficult to attribute to a specific test and so difficult to provide correct metadata for.

### DIFF
--- a/rfcs/crash_test_navigate.md
+++ b/rfcs/crash_test_navigate.md
@@ -1,0 +1,41 @@
+# RFC 193: Navigate after Crash Tests
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


Ms2ger (approved):

gsnedders:
There's some wptrunner PR for this, right?

gsnedders:
I still can't see any implementation of this in wptrunner. We should probably at least have an issue tracking "implement the RFC"?

jgraham:
I had a patch locally which was going to land in m-c but I forgot about. Now posted as https://github.com/web-platform-tests/wpt/pull/49026

PR #198, opened 2024-05-28
RFC 198: Move the container image from Docker Hub to GitHub Registry

OrKoN:
This RFC proposes moving the container image from Docker Hub to GitHub Registry.

Issue https://github.com/web-platform-tests/wpt/issues/28903

[Preview](https://github.com/OrKoN/rfcs/blob/orkon/docker/rfcs/docker_github_registry.md)

### DIFF
--- a/rfcs/docker_github_registry.md
+++ b/rfcs/docker_github_registry.md
@@ -15,8 +15,6 @@ used in taskcluster being outdated as documented in [this PR](https://github.com
 The following are the benefits of using the GitHub registry for hosting:
 
 - It is [hosted on GitHub](https://github.com/web-platform-tests/wpt/packages) together with the wpt repository.
-- Publishing is possible by any project member who has the write permissions to the
-  repository.
 - Publishing is possible [via GitHub Actions](https://github.com/web-platform-tests/wpt/blob/master/.github/workflows/docker.yml):
 
     - The workflow can be manually invoked,
@@ -31,6 +29,24 @@ tools/docker/frontend.py.
 
 This RFC also proposes to change the version numbers to be integers starting
 with `1`.
+
+### Permissions
+
+The following permissions shall apply to publishing of the image and Dockerfile changes:
+
+* Registry permissions (**this part needs to be configured by the project admin on [this page](https://github.com/orgs/web-platform-tests/packages/container/wpt/settings)**):
+  * Default inheritance of permissions from the repository should be disabled and the access needs to be provided to:
+    * The [admin](https://github.com/orgs/web-platform-tests/teams/admins) team
+    * The [core](https://github.com/orgs/web-platform-tests/teams/wpt-core-team) team
+    * the github workflows ([docs](https://docs.github.com/en/packages/learn-github-packages/about-permissions-for-github-packages#maintaining-access-to-packages-in-github-actions-workflows))
+
+* Dockerfile changes
+  * Use CODEOWNERS to require reviews from the admins and core teams.
+
+### Download time
+
+We measured the download time for Docker Hub and Github from a TaskCluster job
+and didn't see any regression (both jobs took approx ~48 seconds).
 
 ## Risks
 
--- a/rfcs/docker_github_registry.md
+++ b/rfcs/docker_github_registry.md
@@ -40,8 +40,8 @@ The following permissions shall apply to publishing of the image and Dockerfile 
     * The [core](https://github.com/orgs/web-platform-tests/teams/wpt-core-team) team
     * the github workflows ([docs](https://docs.github.com/en/packages/learn-github-packages/about-permissions-for-github-packages#maintaining-access-to-packages-in-github-actions-workflows))
 
-* Dockerfile changes
-  * Use CODEOWNERS to require reviews from the admins and core teams.
+* `Dockerfile` and `.taskcluster.yml` changes
+  * Use CODEOWNERS to require reviews for changes in these files from the admins and core teams.
 
 ### Download time
 
--- a/rfcs/docker_github_registry.md
+++ b/rfcs/docker_github_registry.md
@@ -22,7 +22,7 @@ The following are the benefits of using the GitHub registry for hosting:
     - The workflow can be manually invoked,
     - or it can be further automated to deploy any changes that land.
 
-- We can use the chane to change the versioning scheme to use integer numbers.
+- We can take the opportunity to change the versioning scheme to use integer numbers.
 
 The following PR implements changes to use the image from GitHub registry
 instead of the Docker Hub one
--- a/rfcs/docker_github_registry.md
+++ b/rfcs/docker_github_registry.md
@@ -1,0 +1,41 @@
+# RFC 198: Move the container image from Docker Hub to GitHub Registry
+
+## Summary
+
+This RFC proposes moving the container image from Docker Hub to GitHub Registry.
+
+## Details
+
+This RFC addresses the issue
+[#28903](https://github.com/web-platform-tests/wpt/issues/28903). Docker Hub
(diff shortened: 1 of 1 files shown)


rfcs/docker_github_registry.md
```diff
@@ -0,0 +1,41 @@
...
+
+- It is [hosted on GitHub](https://github.com/web-platform-tests/wpt/packages) together with the wpt repository.
+- Publishing is possible by any project member who has the write permissions to the
+  repository.
+- Publishing is possible [via GitHub Actions](https://github.com/web-platform-tests/wpt/blob/master/.github/workflows/docker.yml):
+
+    - The workflow can be manually invoked,
+    - or it can be further automated to deploy any changes that land.
+
+- We can use the chane to change the versioning scheme to use integer numbers.
```
foolip, line 25: Fixing the "chane" typo by rewording:

```suggestion
- We can take the opportunity to change the versioning scheme to use integer numbers.
```

rfcs/docker_github_registry.md
```diff
@@ -0,0 +1,41 @@
...
+
+The following PR implements changes to use the image from GitHub registry
+instead of the Docker Hub one
+https://github.com/web-platform-tests/wpt/pull/46279 including modifications to
+tools/docker/frontend.py.
+
+This RFC also proposes to change the version numbers to be integers starting
+with `1`.
+
+## Risks
```
foolip, line 35: I would guess the biggest risk is that Taskcluster pulling the image gets errors more often than with Docker Hub, or perhaps that some caching works differently so that all jobs take longer to start.

The reliability is hard to judge ahead of time, but the time it takes can be measured once at least to ensure no big difference.

foolip (approved):

rfcs/docker_github_registry.md
```diff
@@ -0,0 +1,41 @@
...
+[#28903](https://github.com/web-platform-tests/wpt/issues/28903). Docker Hub
+organization allows for a limited number of members and only a few members can
+publish a new Docker image that sometimes results in the Docker image that is
+used in taskcluster being outdated as documented in [this PR](https://github.com/web-platform-tests/wpt/pull/44576#issue-2133724788).
+
+The following are the benefits of using the GitHub registry for hosting:
+
+- It is [hosted on GitHub](https://github.com/web-platform-tests/wpt/packages) together with the wpt repository.
+- Publishing is possible by any project member who has the write permissions to the
+  repository.
```
Ms2ger, line 19: I don't know what this image is used for, but that's a lot of people who we haven't (and probably don't want to) vet for more than "is willing to review tests"

rfcs/docker_github_registry.md
OrKoN: I think I can try to get the numbers from the test runs. I will add the information and update the RFC.

rfcs/docker_github_registry.md
OrKoN: Screenshot 2024-05-29 at 10 19 34

Members with the Write access can make changes to the Dockerfile so it seems appropriate to also allow publishing the changes. Alternatively, on the package settings page there is an option to not inherit the permissions or disallow access from the workflows so the admin can change that and only allow selected members if that is desired.

jgraham: We could use a different repo for the images if we wanted different permissions. 

We can also use `CODEOWNERS` to ensure that any changes to the docker image get admin review before being merged which would prevent updating the CI to use a new version of the image, but wouldn't prevent  publishing images in the first place.

We can also change the permissions on specific packages (notably the `wpt` package) to not inherit from the wpt base permissions. I think it would still (and maybe always was?) possible for people to publish new packages which would by default inherit permissions from the repo.

jcscottiii (commented):
Hi @OrKoN 

We discussed this today in the infra [meeting](https://github.com/web-platform-tests/wpt-notes/blob/master/minutes/2024-06-04.md).

Overall this RFC looks good. Can you add a Permissions subsection to the Details section? In particular:

- Docker image pushing:
  - Disable default inheritance from the repository and provide access to:
    - The [admin](https://github.com/orgs/web-platform-tests/teams/admins) team
    - The [core](https://github.com/orgs/web-platform-tests/teams/wpt-core-team) team
    - the github workflows ([docs](https://docs.github.com/en/packages/learn-github-packages/about-permissions-for-github-packages#maintaining-access-to-packages-in-github-actions-workflows))
  - By default, a lot of individuals would have access to publish this package through the inherited permissions. This will limit the blast radius in case a certain account is compromised and pushes a package.
 - Use CODEOWNERS to require reviews from the admins and core teams for Dockerfile changes.

rfcs/docker_github_registry.md
OrKoN: I have took two random TC jobs using the old and the new image and suprisingly the download time is ~48 seconds for both (I didn't see any indication of caching in the logs):

dockerhub registry: https://community-tc.services.mozilla.com/tasks/ZNnHobGNTiS4qBiv0l7Z-A/runs/0/logs/live/public/logs/live.log

github registry: https://community-tc.services.mozilla.com/tasks/F3k9fZ6TSdC6Cd9kIwnX0A/runs/1/logs/public/logs/live.log

OrKoN:
@jcscottiii thanks for the feedback, I have included this into RFC. Note that someone with an admin permissions should configure the desired permissions (I have no access) and I can only add CODEOWNERS to my PR.

OrKoN:
Screenshot 2024-06-04 at 20 03 58

The admins team does not seem to be suitable for CODEOWNERS. Perhaps it does not have write access?

rfcs/docker_github_registry.md
```diff
@@ -0,0 +1,57 @@
...
+The following permissions shall apply to publishing of the image and Dockerfile changes:
+
+* Registry permissions (**this part needs to be configured by the project admin on [this page](https://github.com/orgs/web-platform-tests/packages/container/wpt/settings)**):
+  * Default inheritance of permissions from the repository should be disabled and the access needs to be provided to:
+    * The [admin](https://github.com/orgs/web-platform-tests/teams/admins) team
+    * The [core](https://github.com/orgs/web-platform-tests/teams/wpt-core-team) team
+    * the github workflows ([docs](https://docs.github.com/en/packages/learn-github-packages/about-permissions-for-github-packages#maintaining-access-to-packages-in-github-actions-workflows))
+
+* Dockerfile changes
+  * Use CODEOWNERS to require reviews from the admins and core teams.
```
jgraham, line 44: Can you be specific about which changes will require review? I think it should be changes to the Dockerfile and changes to the `.taskcluster.yml` file.

jgraham (approved):
Approved, but note the comment.

rfcs/docker_github_registry.md
OrKoN: Updated

foolip:
> Screenshot 2024-06-04 at 20 03 58
> The admins team does not seem to be suitable for CODEOWNERS. Perhaps it does not have write access?

Yeah, everyone in the admin team is made an org owner, so we never give the team itself write access to anything. I've given the team write access to the wpt repo now, can you try again?

OrKoN:
@foolip I have updated my PR, and it works now. The admins can be specified in CODEOWNERS.

OrKoN:
Shall we land this RFC? (I have no permission to merge, so if anyone could hit the merge button it would be great)

jgraham:
I think technically we need to wait another day.

OrKoN:
@foolip @jgraham @jcscottiii it looks like a bit more than 2 weeks already passed and there are not unresolved concerns. Therefore, if someone could merge the PR, it would be great.

PR #206, opened 2024-08-06
RFC: Web Platform Feature Known Issues Management Repository

jcscottiii:


rfcs/web_features_known_issues.md
```diff
@@ -0,0 +1,58 @@
+## Summary
+
+This RFC proposes the creation of a new repository within the web-platform-tests organization to manage information about known issues affecting web platform tests. The repository will provide a transparent and collaborative space for documenting these issues. The data in this repository aims to be machine readable source of truth for sites or tools that make use of the web-features grouping of WPT such as webstatus.dev.
+
+## Details
+
+* **Repository Name:** web-platform-tests-feature-issues
```
foolip, line 7: Any repo name will be OK, but in terms of tab completion, I wonder if web-features-issues would be acceptable?

rfcs/web_features_known_issues.md
```diff
@@ -0,0 +1,58 @@
...
+* **Process:**
+    1. Submit a Pull Request (PR) to add/modify feature files.
+    2. Community review and discussion.
+    3. Approval and merging of the PR.
+* **Governance:**
+    * Open to contributions from all browser vendors and the wider community.
+    * Clear guidelines for approvals and time limits will be established.
+* **Enumerations:**
+    * To standardize reasons for hiding scores, the following enumerations will be used in feature YAML files:
+        * `incomplete`: Less than 50% of the feature is tested by any reasonable coverage metric..
```
foolip, line 21:
```suggestion
        * `incomplete`: Less than 50% of the feature is tested by any reasonable coverage metric.
```

jgraham (changes requested):
Just to make sure the discussion we've had in the meeting is recorded here:

This RFC needs to explain what problem it's trying to solve. The way it's phrased here it seems like it's about implementing a feature on `webstatus.dev`. But given there are existing concerns about the incentive structure generated by that project being detrimental to the overall health of the web platform tests project, anything that's motivated by that project needs to explain how it's going to address those concerns.

Also, if this is something that requires governance, the governance section needs to be much more explicit, rather than just saying "will be established".

Ms2ger (changes requested):
I also don't understand what this is for

PR #210, opened 2024-09-16
RFC 199: Add Bazel configuration files

anonrig:
Following up for the changes on https://github.com/web-platform-tests/wpt/pull/48084, I'd like to propose adding 2 Bazel files to web-platform tests to make integrations as seamless as possible.

cc @jasnell

Note: This RFC does not propose running WPT on Bazel. It proposes adding Bazel files to the git repository so that the implementors such as Cloudflare Workers can easily run WPT without any modifications.

### DIFF
--- a/rfcs/bazel_files.md
+++ b/rfcs/bazel_files.md
@@ -1,0 +1,55 @@
+# RFC 199: Add Bazel configuration files
+
+## Summary
+
+This RFC proposes adding a BUILD.bazel file and a tests.bzl file to the Web Platform Tests
+(WPT) GitHub repository.
+
+## Details
+
+To support build systems like Bazel, implementors such as Cloudflare are currently required
+to maintain patches on top of the Web Platform Tests GitHub repository. These patches expose
+file groups representing each subfolder of the repository.
+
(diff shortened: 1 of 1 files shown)


anonrig:
hey @jgraham, what are the next steps?

jgraham (changes requested):
My high-level concern here is that this seems to embed vendor-specific configuration decisions into the shared repository. We generally try to avoid that e.g. we don't have `moz.build` files from mozilla or ninja files from Chromium. It seems potentially confusing for users to see a Bazel file and assume that there might be some requirement to use Bazel in order to run the tests. Even for other vendors who are using Bazel it's not clear to me to what extent the proposal here is generic to any use case / setup vs encoding cloudflare-specific decisions (for example, maybe other vendors don't want the same division of tests by directories? I'm not sure).

It seems like https://github.com/web-platform-tests/wpt/pull/48084#issuecomment-2372593126 has some suggestions for alternate approaches that might not require landing this configuration in the upstream repo; I'd suggest starting with that and only proceeding with this approach if that can't work, or if there is clear multi-vendor demand for the same Bazel setup.

PR #211, opened 2024-10-08
RFC 211: Support testdriver.js in other test types

jonathan-j-lee:
[Rendered](https://github.com/jonathan-j-lee/rfcs/blob/testdriver-in-other-types/rfcs/testdriver_in_other_types.md)

Tracking issue: web-platform-tests/wpt#13183
Prototype implementation: web-platform-tests/wpt#48486
CC: @mfreed7 @WeizhongX

### DIFF
--- a/rfcs/testdriver_in_other_types.md
+++ b/rfcs/testdriver_in_other_types.md
@@ -1,0 +1,132 @@
+# RFC 209: Support testdriver.js in other test types
+
+## Summary
+
+Allow reftests, print-reftests, and crashtests to use `testdriver.js`, and add
+the necessary support to `wpt {run,manifest,lint}`.
+
+## Motivation
+
+[`testdriver.js`][0] provides APIs that allow tests to send WebDriver-like
+commands from client-side JavaScript.
+Tests often use testdriver.js to access powerful capabilities that would
+normally not be automatable because the web platform does not expose them.
+
+Currently, only [testharness tests] may use testdriver.js.
+However, test authors [have expressed interest] in extending testdriver.js
+support to [reftests], WPT's main source of rendering coverage.
+The features they wish to test with respect to rendering are often gated on
+user activation or require pointer input, which can't be obtained without
+testdriver.js or manual interaction:
+
+| Tests | testdriver.js usage | Purpose |
+| --- | --- | --- |
+| [`fullscreen/`][1] | `bless()` | Provide [required transient activation][2] |
+| [`customizable-select/`][3] | `bless()`, `click()` | Provide transient activation [required to show `<select>` pickers][4] |
+| Various [`css/`][5] | `Actions().pointerMove()` | Exercise `:hover` styles |
+
+* The `customizable-select/` tests are confined to [Chromium's
+  `wpt_internal/`][6] because only Chromium's `run_web_tests.py` and
+  [`testdriver-vendor.js`][7] can run them.
+  These tests are effectively not portable to other vendors.
+* The other tests have already landed in web-platform-tests/wpt, but
+  [time out in all browsers][1] when run with [wptrunner], WPT's canonical test
+  harness.
+  These tests don't provide reliable coverage.
+
+[have expressed interest]: https://github.com/web-platform-tests/wpt/issues/13183
+[wptrunner]: https://web-platform-tests.org/tools/wptrunner/README.html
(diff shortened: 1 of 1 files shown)


WeizhongX (approved):
LGTM

past:
Here are the [relevant meeting notes](https://github.com/web-platform-tests/wpt-notes/blob/master/minutes/2024-11-05.md#211-httpsgithubcomweb-platform-testsrfcspull211---support-testdriverjs-in-other-test-types) from this week's discussion.

jgraham:
So, in principle I support the idea of this. In practice it's really hard for Gecko to implement at the moment, because we have a specific internal endpoint for running reftests, and although it's not impossible to imagine a mechanism to allow it to be interrupted, it feels likely to be challenging. In theory we could fall back to the more standard reftest implementation only for these reftests, allowing an implementation much more similar to the WebDriver one, although that might represent a surprising amount of behaviour change.

Also in the meeting @gsnedders brought up some concerns that even with this support some of the use cases listed might be challenging to support cross-browser. In particular, testing `<select>` might end up in poorly-defined territory; whether the popup window should be accessible to WebDriver seems unclear, although maybe the customizable select work itself makes that clearer.

WeizhongX:
> So, in principle I support the idea of this. In practice it's really hard for Gecko to implement at the moment, because we have a specific internal endpoint for running reftests, 

I thought with this change, reftests do not use testdriver should continue to work as usual?

Jonathan has a draft PR here: https://github.com/web-platform-tests/wpt/pull/48486.

> Also in the meeting @gsnedders brought up some concerns that even with this support some of the use cases listed might be challenging to support cross-browser. In particular, testing `<select>` might end up in poorly-defined territory; whether the popup window should be accessible to WebDriver seems unclear, although maybe the customizable select work itself makes that clearer.

Maybe in this RFC our goal is to enable testdriver for reftests, crash tests etc? Other issues should be addressed separately. Popup window is an issue even for now, right?

mfreed7:
> Also in the meeting @gsnedders brought up some concerns that even with this support some of the use cases listed might be challenging to support cross-browser. In particular, testing `<select>` might end up in poorly-defined territory; whether the popup window should be accessible to WebDriver seems unclear, although maybe the customizable select work itself makes that clearer.

Just to clarify, there are two "modes" that `<select>` can use for popups. One is the old/existing one, where (in all mainstream browsers) the popup window is actually a separate OS window. I agree that making this interoperably-testable will be tough. However, the other mode is the new "customizable-`<select>`" mode, where `select, ::picker(select) {appearance: base-select}` is used. In that case, the picker is defined to be a popover, so it renders right in the same renderer. There are no separate popup windows. That mode is the one for which we badly need this capability. And hopefully it doesn't present any issues, right?

jonathan-j-lee:
Edited the "Risks" section a bit to capture the practical implementation challenges discussed. PTAL @jgraham @gsnedders

gsnedders (approved):
I think it's true that the only real risk here is with the internal Marionette runner; otherwise this has broadly seemed like a good idea for a while.

rfcs/testdriver_in_other_types.md
```diff
@@ -0,0 +1,150 @@
...
+
+Because the [default completion behavior][completion-behavior] is not
+well-ordered with respect to arbitrary testdriver calls, testdriver in
+non-testharness tests will only be supported with `class=(ref)test-wait`.
+
+[completion-behavior]: https://web-platform-tests.org/writing-tests/reftests.html#controlling-when-comparison-occurs
+
+### Manifest Format Changes
+
+Similar to testharness manifest items, items for the other types may set a
```
jgraham, line 83: This shouldn't be "may", but "will" or "must".

jgraham (approved):
I think this is OK; a Gecko implementation that falls back to the WebDriver-style external runner isn't ideal, but it's likely better than not having the feature at all.

rfcs/testdriver_in_other_types.md
jonathan-j-lee: Reworded a bit

past:
Thanks!

mfreed7:
I see that this landed 🎉 . But it's "just" the RFC. I assume (but please correct me if I'm wrong) that this doesn't yet mean WPTs can use testdriver.js for reference tests? Or am I now free to start committing such tests?

WeizhongX:
Not yet, the actual change to Wptrunner is under review now.

PR #212, opened 2024-10-17
RFC 212: Introduce `require_webdriver_bidi` metadata tag

sadym-chromium:
[Preview](https://github.com/sadym-chromium/wpt_rfcs/blob/sadym/require_webdriver_bidi/rfcs/require_webdriver_bidi.md)


rfcs/require_webdriver_bidi.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Background
+
+[RFC 185: Add WebDriver BiDi support to testdriver.js](https://github.com/web-platform-tests/rfcs/blob/master/rfcs/testdriver_bidi.md) discusses incorporating BiDi functionality into `testdriver.js`. This entails migrating from [WebDriverProtocol](https://github.com/web-platform-tests/wpt/blob/9c76757f5332678f9952f6ccb3824f62d30eca1f/tools/wptrunner/wptrunner/executors/executorwebdriver.py#L644) to [WebDriverBidiProtocol](https://github.com/web-platform-tests/wpt/blob/9c76757f5332678f9952f6ccb3824f62d30eca1f/tools/wptrunner/wptrunner/executors/executorwebdriver.py#L736). However, this switch may cause regression in certain tests that don't utilize the testdriver BiDi API due to unintended side effects resulting from the transport change. Nonetheless, all these regressions need to be resolved eventually. Implementing a flag will enable immediate provision of the testdriver BiDi API to tests, eliminating the need for test authors to wait for the migration of all previous tests to `WebDriverBidiProtocol`.
```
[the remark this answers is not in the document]
sadym-chromium: > This feels like it presupposes migrating everything over to `WebDriverBiDiProtocol`, instead of continuing with the status quo of some being in `WebDriverProtocol` and others BiDi-exclusive being in `WebDriverBiDiProtocol`?

I guess migrating everything over to `WebDriverBiDiProtocol` can happen at certain point, but not in the nearest future. As long as we don't have consensus on that yet, let's consider we are going to support both.

> That wasn't in the previous RFC as far as I can see, and that is a much larger (and more breaking change) than we've previously discussed. It is highly likely that there will be a number of vendors with no BiDi implementation running wptrunner for a long time into the future, thus migrating things over to requiring BiDi when they can be implemented via Classic seems problematic.

This document does not restrict usage of `WebDriverProtocol` and such a migration (BiDi by default) should be discussed and planned separately.

rfcs/require_webdriver_bidi.md
```diff
@@ -0,0 +1,114 @@
+# RFC 212: Introduce `require_webdriver_bidi` metadata tag
+
+## Summary
+
+To facilitate the gradual enablement of WebDriver BiDi transports, this RFC proposes the introduction of a new metadata tag, `require_webdriver_bidi` in html and js files in WPT tests. This metadata would allow for a gradual transition from [WebDriverProtocol](https://github.com/web-platform-tests/wpt/blob/9c76757f5332678f9952f6ccb3824f62d30eca1f/tools/wptrunner/wptrunner/executors/executorwebdriver.py#L644) to [WebDriverBidiProtocol](https://github.com/web-platform-tests/wpt/blob/9c76757f5332678f9952f6ccb3824f62d30eca1f/tools/wptrunner/wptrunner/executors/executorwebdriver.py#L736), ensuring that implementations can opt-in into WebDriver BiDi gradually in case there are implementation blockers like in Chrome's case. [Prototype](https://github.com/web-platform-tests/wpt/pull/48622).
```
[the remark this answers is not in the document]
sadym-chromium: I believe this paragraph is poorly worded. What I meant is that the flag indicates if the test requires BiDi protocol. Currently if some specific testdriver API is not implemented in the implementation protocol and test tries to use it, a error **"{method name} is not implemented by testdriver-vendor.js"** is thrown. This part is kept untouched.

> Because one could imagine a test authored by people working on WebKit that works fine with BiDi there, but will fail in Chrome due to some BiDi implementation blocker. From the WebKit side, putting the require_webdriver_bidi metadata in the test seems fine, but from the Chrome side it doesn't, thus the test is the wrong place to store this metadata.

In this example the test will fail with the **"{method name} is not implemented by testdriver-vendor.js"** exception in Chrome.

@gsnedders please let me know if you think this approach is reasonable, and will update the RFC without mentioning "gradual enablement".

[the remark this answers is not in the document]
sadym-chromium: It will be something like "`require_webdriver_bidi` is used to signal to the Browser/Executor that the test uses [testdriver.bidi](https://github.com/web-platform-tests/wpt/blob/2ab788df7f0d2ee35d67511348ce5eb4162c12d5/resources/testdriver.js#L55)".

rfcs/require_webdriver_bidi.md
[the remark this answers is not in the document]
sadym-chromium: The wording is fixed. We stick to the #185 proposal, and do not propose to retire `WebDriverProtocol`.

rfcs/require_webdriver_bidi.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Risks
+
+### Having metadata in the Test is not sufficient
+
+`TestExecutor` needs information about whether to activate BiDi or not at `connect`, which happens before the runner receives the test metadata. This can be addressed by passing the data to implementation-specific `TestharnessExecutor` via the implementation-specific `BrowserSettings` and  like in the [example](https://github.com/web-platform-tests/wpt/pull/48618/files#diff-0df24b5b583c460182e687f7dc7a6a79dd2cd3389bc4a96f48483f60fceb51f7R269).
+
+### Potential false-negative due to missing `require_webdriver_bidi` tag
+
+Some implementations can enable BiDi by default, while others would still need enabling it via `require_webdriver_bidi`. This could lead to false-negative test passes by the tests requiring BiDi but not having the `require_webdriver_bidi` tag. This risk can be mitigated by adding documentation describing the logic behind.
```
jgraham, line 111: The BiDi testdriver APIs could enforce a check that they are only being called by a test with this flag set, perhaps? At least when they're called from a top-level test. I don't think documentation alone will be enough here.

rfcs/require_webdriver_bidi.md
```diff
@@ -0,0 +1,121 @@
...
+
+#### Lint rules
+
+Some lint rules has to be added to verify is used correctly:
+* It is unique per test file.
+* It has the correct value (either `true` or `false`).
+
+### Deprecating of the `require_webdriver_bidi` metadata tag
+
+Expectations are that eventually all the implementations would support BiDi either via WebDriverBiDi, or via implementation-specific bindings. At that point, the `require_webdriver_bidi` metadata tag can be deprecated. This can be done by simply removing the tag.
```
jgraham, line 101: s/deprecated/removed/

rfcs/require_webdriver_bidi.md
```diff
@@ -0,0 +1,121 @@
...
+
+Some implementations can enable BiDi by default, while others would still need enabling it via `require_webdriver_bidi`. This could lead to false-negative test passes by the tests requiring BiDi but not having the `require_webdriver_bidi` tag. This risk can be mitigated by adding documentation describing the logic behind.
+
+## Alternatives considered
+
+### Add metadata to `.ini` files
+
+In this alternative approach, the `require_webdriver_bidi` metadata would be added to the corresponding `.ini` file. However, this approach was ultimately declined because the `.ini` file metadata is primarily concerned with the expectations of the test run rather than the specific API requirements. On the other hand, the `require_webdriver_bidi` metadata is directly related to the test API itself.
+
+### Extract testdriver BiDi API into a separate file
```
jgraham, line 119: A different alternative here would be to make `testdriver.js` have a `features` query parameter that it could check before enabling the features, and which could be supported in the manifest to extract the required features. So the author would need to write `<script src="/resources/testdriver.js?features=bidi"></script>`, and that would be used both by the js code to decide whether to enable the feature, and also the manifest code to set the flag on the test. The problem would be cases where you include `testdriver` in a support file but _not_ in the top-level test.

jgraham (commented):
In principle I think this seems like a reasonable approach to the problem of opting in to BiDi and has the short-term benefit that implementations without BiDi support can simply skip the tests that require it.

rfcs/require_webdriver_bidi.md
sadym-chromium: To clarify: do you think it worse mentioning as analternative, or you insist we should implement this way instead?

jgraham: I thought of it when reviewing, and I think it's worth discussing here :)

The fact that it doesn't really work for tests that are using testdriver outside the test document seems like a notable downside because you'd still need some mechanism e.g. the meta tag in the test document. But it does naturally fail closed in the sense that if you don't specify the feature you can't use it, which is a notable advantage.

sadym-chromium: I don't think think this approach would allow adding or not adding `bidi` in the `test_driver` object. So in fact in both cases of `testdriver.js?features=bidi` and `testdriver.js`, the `bidi` will be in the `test_driver` object. However, it's up to the backend to restrict the bidi calls if "bidi" parameter is not set. Please confirm I understood your proposal correctly, and if so, I think it makes sense and I will re-write the proposal accordingly.

sadym-chromium: UPD: actually, `testdriver` can use `document.currentScript.src` to get it's query parameters, and enable/disable BiDi based on this.

sadym-chromium: https://github.com/sadym-chromium/wpt_rfcs/blob/sadym/require_webdriver_bidi/rfcs/require_webdriver_bidi.md#add-feature-query-parameter-to-testdriverjs

mathiasbynens: > UPD: actually, `testdriver` can use `document.currentScript.src` to get it's query parameters, and enable/disable BiDi based on this.

Note that this doesn’t work in `<script type=module>`. Is that a use case testdriver needs/wants to support?

sadym-chromium: I started drafting this alternative in the https://github.com/web-platform-tests/rfcs/pull/214

OrKoN: > Note that this doesn’t work in <script type=module>. Is that a use case testdriver needs/wants to support?

I think testdriver.is currently not a module but `import.meta.url` could be used if it becomes a module.

past:
Superseded by [RFC 214](https://github.com/web-platform-tests/rfcs/blob/master/rfcs/testdriver-features.md).

PR #213, opened 2024-10-31
RFC 213: Wait for `document.fonts.ready` before running `checkLayout()` tests

jonathan-j-lee:
[Rendered](https://github.com/jonathan-j-lee/rfcs/blob/check-layout/wait-for-fonts/rfcs/check_layout_wait_for_fonts.md)

Tracking issue: https://crbug.com/375344326

### DIFF
--- a/rfcs/check_layout_wait_for_fonts.md
+++ b/rfcs/check_layout_wait_for_fonts.md
@@ -1,0 +1,73 @@
+# RFC 213: Wait for `document.fonts.ready` before running `checkLayout()` tests
+
+## Summary
+
+Change [`checkLayout()`][check-layout-th] to wait for
+[`document.fonts.ready`][fonts-ready] to fulfill before running tests.
+
+[check-layout-th]: https://github.com/web-platform-tests/wpt/blob/9952f6fa233ad9e067b2ff28ab32832cc6d8fd66/resources/check-layout-th.js#L211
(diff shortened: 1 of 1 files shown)


gsnedders:
I wonder, without thinking about this much, whether we should use the same logic as reftests (and crashtests, incidentally) use: https://github.com/web-platform-tests/wpt/blob/6c7c4003dba847b63e16083a0b7b8f473d9ad7e1/tools/wptrunner/wptrunner/executors/test-wait.js

Notably, this waits for the load event (and in some browsers this might wait for font loading, c.f. https://github.com/w3c/csswg-drafts/issues/1088), which maybe relevant for rendering when images for example are involved, and then waits for `document.fonts.ready` and then two further rAFs (which we probably don't need).

jgraham (approved):

jonathan-j-lee:
Reviewing [the meeting minutes], @jgraham said:

> Removing the installation of fonts should require a RFC.

Just to be clear, this RFC is not proposing the removal of `wpt run --install-fonts`, which some users might still find useful. This RFC only drops that option from WPT CI.

Assuming we're on the same page about that, I think this RFC is ready to be accepted?

[the meeting minutes]: https://github.com/web-platform-tests/wpt-notes/blob/master/minutes/2024-11-05.md#213-httpsgithubcomweb-platform-testsrfcspull213---wait-for-documentfontsready-before-running-checklayout-tests

PR #214, opened 2024-11-11
RFC XXX: Add `?feature=` query parameter to `testdriver.js`

sadym-chromium:
Alternative to [RFC 212](https://github.com/web-platform-tests/rfcs/pull/212).

### DIFF
--- a/rfcs/testdriver-features.md
+++ b/rfcs/testdriver-features.md
@@ -1,0 +1,104 @@
+# RFC XXX: Add `?feature` query parameter to `testdriver.js`
+
+## Summary
+
+This RFC is an alternative to [RFC 212](https://github.com/web-platform-tests/rfcs/pull/212). It proposes adding a query parameter `?features` to `testdriver.js`. At the moment the only supported feature will be `bidi`, which will be used to enable the [testdriver BiDi API](https://github.com/web-platform-tests/rfcs/blob/master/rfcs/testdriver_bidi.md) for specific WPT tests, The supported features can be extended by other RFCs if required. The feature `bidi`, applicable to HTML and JS files, would let test runners choose between [**WebDriverProtocol**](https://github.com/web-platform-tests/wpt/blob/9c76757f5332678f9952f6ccb3824f62d30eca1f/tools/wptrunner/wptrunner/executors/executorwebdriver.py#L644) and [**WebDriverBidiProtocol**](https://github.com/web-platform-tests/wpt/blob/9c76757f5332678f9952f6ccb3824f62d30eca1f/tools/wptrunner/wptrunner/executors/executorwebdriver.py#L736). This approach does not require all the browsers to implement WebDriver BiDi.
+
+[Prototype](https://github.com/web-platform-tests/wpt/pull/48622).
+
+## Background
+
+[RFC 185](https://github.com/web-platform-tests/rfcs/blob/master/rfcs/testdriver_bidi.md) proposes adding BiDi support to  `testdriver.js`, which involves using [WebDriverBidiProtocol](https://github.com/web-platform-tests/wpt/blob/9c76757f5332678f9952f6ccb3824f62d30eca1f/tools/wptrunner/wptrunner/executors/executorwebdriver.py#L736) instead of [WebDriverProtocol](https://github.com/web-platform-tests/wpt/blob/9c76757f5332678f9952f6ccb3824f62d30eca1f/tools/wptrunner/wptrunner/executors/executorwebdriver.py#L644) for test runners that use the WebDriver protocol. However, this change could cause regressions in tests that don't use the testdriver BiDi API due to unintended side effects of the transport change. To address this, the query parameter `?features=bidi` in `testdriver.js` will inform the test runner whether the given test requires the testdriver BiDi API, allowing the runner to choose the appropriate transport, if the default one does not support BiDi protocol parts.
+
+If at a certain point all the browsers implement testdriver BiDi and enable it by default, this flag can be deprecated.
+
+## Details
+
+### Examples
+
+[html test](https://github.com/web-platform-tests/wpt/blob/9395d384f5c69a9a3a7fc4de04249f77500b2d3f/infrastructure/webdriver/bidi/subscription.html#L3)
+```javascript
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Test console log are present</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="/resources/testdriver.js?features=bidi"></script>
+<script src="/resources/testdriver-vendor.js"></script>
+<script>
+    promise_test(async () => {
+        const some_message = "SOME MESSAGE";
+        // Subscribe to `log.entryAdded` BiDi events. This will not add a listener to the page.
+        await test_driver.bidi.log.entry_added.subscribe();
+        // Add a listener for the log.entryAdded event. This will not subscribe to the event, so the subscription is
+        // required before. The cleanup is done automatically after the test is finished.
+        const log_entry_promise = test_driver.bidi.log.entry_added.once();
+        // Emit a console.log message.
+        // Note: Lint rule is disabled in `lint.ignore` file.
+        console.log(some_message);
+        // Wait for the log.entryAdded event to be received.
+        const event = await log_entry_promise;
+        // Assert the log.entryAdded event has the expected message.
+        assert_equals(event.args.length, 1);
+        const event_message = event.args[0];
+        assert_equals(event_message.value, some_message);
+    }, "Assert testdriver can subscribe and receive events");
+```
+
+```javascript
+// META: script=/resources/testdriver.js?features=bidi
+
+'use strict';
+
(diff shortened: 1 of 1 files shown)


rfcs/testdriver-features.md
```diff
@@ -0,0 +1,104 @@
...
+    const event_message = event.args[0];
+    assert_equals(event_message.value, some_message);
+}, "Assert testdriver can subscribe and receive events");
+```
+
+### Required changes
+
+#### Features
+
+The `testrdriver_features` is a list of values of query parameters with the name `feature`, used when the `testdriver.js` is imported. It has to be passed from `SourceFile` through `TestharnessTest` to `Test`, where it can be consumed by the specific test runners.
```
OrKoN, line 76: Is testrdriver_features a new concept? or does it already exist?

sadym-chromium: it's new

rfcs/testdriver-features.md
```diff
@@ -0,0 +1,104 @@
...
+
+Even though the `TestExecutor` needs information about whether to activate BiDi or not at `connect`, which happens before the runner receives the test to run, the configuration can be done at the point [the browser is configured](https://github.com/web-platform-tests/wpt/pull/48618/files#diff-0df24b5b583c460182e687f7dc7a6a79dd2cd3389bc4a96f48483f60fceb51f7R272). This data can be used later by the specific executor implementation to [decide which protocol to use](https://github.com/web-platform-tests/wpt/pull/48618/files#diff-c2f15328bc1ddfa8fb93c09ae41651e2bcfdad4f257932263a3e92c3f8deffceR248).
+
+#### Lint rules
+
+Some lint rules has to be added to verify is used correctly:
+* It is unique per test file.
+* It has the correct value (either `true` or `false`).
+
+### Deprecating of the `require_webdriver_bidi` metadata tag
```
OrKoN, line 92: this does not seem applicable to this RFC?

OrKoN (approved):

sadym-chromium:
@jgraham @gsnedders @jcscottiii

jcscottiii:
I added this to the upcoming infra meeting: https://pad.0b101010.services/wpt-infra-meeting?edit

rfcs/testdriver-features.md
```diff
@@ -0,0 +1,58 @@
...
+...
+// META: script=/resources/testdriver.js?feature=bidi
+...
+```
+
+### Required changes
+
+#### Configure testdriver.js based on the `?feature` flag
+
+The `testdriver.js` checks if the `?feature=bidi` query parameter was included using [document.currentScript.src](https://developer.mozilla.org/en-US/docs/Web/API/Document/currentScript) like in the [prototype](https://github.com/web-platform-tests/wpt/pull/49122/files#diff-1fe2b624679a3150e5c86f84682c5901b715dad750096a524e8cb23939e5590fR16). If `bidi` was not enabled, a [descriptive error will be thrown](https://github.com/web-platform-tests/wpt/pull/49122/files#diff-1fe2b624679a3150e5c86f84682c5901b715dad750096a524e8cb23939e5590fR22).
```
jgraham, line 36: We need to document in the RFC what the actual syntax of `feature` is. For example is it allowed to supply it multiple times, or is it a (e.g. comma) separated list?

rfcs/testdriver-features.md
```diff
@@ -0,0 +1,58 @@
...
+
+#### (implementation-specific) Set up WebDriver BiDi in the test executor if needed
+
+Even though the `TestExecutor` needs information about whether to activate BiDi or not at `connect`, which happens before the runner receives the test to run, the configuration can be done at the point [the browser is configured](https://github.com/web-platform-tests/wpt/pull/49122/files#diff-0df24b5b583c460182e687f7dc7a6a79dd2cd3389bc4a96f48483f60fceb51f7R275). This data can be used later by the specific executor implementation to [decide which protocol to use](https://github.com/web-platform-tests/wpt/pull/49122/files#diff-0df24b5b583c460182e687f7dc7a6a79dd2cd3389bc4a96f48483f60fceb51f7R275).
+
+## Risks
+
+### Having `testrdriver_features` in the `Test` is not sufficient
```
jgraham, line 48:
```suggestion
### Having `testdriver_features` in the `Test` is not sufficient
```

rfcs/testdriver-features.md
```diff
@@ -0,0 +1,58 @@
...
+
+#### Add `testrdriver_features` to TestharnessTest
+
+This RFC proposes adding `testrdriver_features` to [`TestharnessTest`](https://github.com/web-platform-tests/wpt/pull/49122/files#diff-8f280b64b0700ab9b4b343adabc7ff4d5ea4b1f45cb6c4bd7f50a19b21ebdb8cR169) so that executors can get the list of features declared by a test. The list is populated by parsing the `feature` query param from the `testdriver.js` imports used by a test. Currently, only the `bidi` feature is supported.
+
+#### (implementation-specific) Set up WebDriver BiDi in the test executor if needed
+
+Even though the `TestExecutor` needs information about whether to activate BiDi or not at `connect`, which happens before the runner receives the test to run, the configuration can be done at the point [the browser is configured](https://github.com/web-platform-tests/wpt/pull/49122/files#diff-0df24b5b583c460182e687f7dc7a6a79dd2cd3389bc4a96f48483f60fceb51f7R275). This data can be used later by the specific executor implementation to [decide which protocol to use](https://github.com/web-platform-tests/wpt/pull/49122/files#diff-0df24b5b583c460182e687f7dc7a6a79dd2cd3389bc4a96f48483f60fceb51f7R275).
+
+## Risks
```
jgraham, line 46: A risk that isn't documented here is that testdriver may be used in a subresource, in which case both the top-level resource and the subresource both need to specify the same set of features. This is hard to lint (since we don't have a total graph of subresource usage). The worst case scenario would be if the subresource allowed use of the bidi APIs without the top-level test opting in (e.g. in tests authored against implementations where the flag is a no-op in wptrunner). I think we can avoid this for now, since all testdriver commands go through the top level script, but that will hopefully change in the future (ironically only once we can use BiDi for wptrunner itself).

sadym-chromium: Good point, added it.

rfcs/testdriver-features.md
jgraham: I think we need to specify, and implement, the extension to multiple query parameters now. Otherwise if someone does use this for something internal they might pick a different approach and then we'll be stuck.

I don't think you need to answer any of the other questions. Just specify that for forwards compatibility multiple feature parameters are permitted in the query string and matches are based on exact string match.

rfcs/testdriver-features.md
```diff
@@ -0,0 +1,64 @@
...
+
+The query parameter `feature` currently supports only the `bidi` feature. Future implementations may allow for multiple `feature` parameters, such as `testdriver.js?feature=bidi&feature=SOME_OTHER_FEATURE`.
```
jgraham, line 7:
```suggestion
The query parameter `feature` takes a single value, which is a string representing a feature name. Feature names are compared by case-sensitive string matching. Multiple features are represented as multiple `feature` parameters for example `testdriver.js?feature=bidi&feature=SOME_OTHER_FEATURE`. This RFC only defines the `bidi` feature; future RFCs may add additional features.
```

jgraham: I would also suggest that we explicitly state that any features added by vendors for their own tests must contain a `:` in the name; then we can write a lint that only permitted features are being used and so catch `feature=BiDi` or similar.

sadym-chromium: > I would also suggest that we explicitly state that any features added by vendors for their own tests must contain a `:` in the name; then we can write a lint that only permitted features are being used and so catch `feature=BiDi` or similar.

I believe it makes sense not only to allow for extension features, but for extension query parameters, which is implemented in the prototype lint rules for [query parameters](https://github.com/web-platform-tests/wpt/pull/49122/files#diff-b3de46e07408c6a68229472f80ec0f23d8a7ff634bc01b25f51ff054478dcc60R554) and [features](https://github.com/web-platform-tests/wpt/pull/49122/files#diff-b3de46e07408c6a68229472f80ec0f23d8a7ff634bc01b25f51ff054478dcc60R562).

sadym-chromium: Added this description to the RFC.

sadym-chromium: > I believe it makes sense not only to allow for extension features, but for extension query parameters, which is implemented in the prototype lint rules for [query parameters](https://github.com/web-platform-tests/wpt/pull/49122/files#diff-b3de46e07408c6a68229472f80ec0f23d8a7ff634bc01b25f51ff054478dcc60R554) and [features](https://github.com/web-platform-tests/wpt/pull/49122/files#diff-b3de46e07408c6a68229472f80ec0f23d8a7ff634bc01b25f51ff054478dcc60R562).

After some further consideration, I don't think we should allow for any vendor-specific query parameters at least for now. If required, this can be changed in another RFC.

jgraham (approved):

past:
Merging this since 1 week passed after the last review and no issues were brought up in our last infra meeting.

Issue #30, opened 2019-09-17
Create an RFC for what material is and is not allowed in WPT


boazsender:
We lack a clear policy of when to accept/reject material.

Some further action is needed after the discussion at TPAC regarding "tests that depend on test-only APIs implemented using Mojo for Chromium, e.g. [WebXR](https://github.com/immersive-web/webxr-test-api/blob/master/explainer.md), [WebUSB](https://wicg.github.io/webusb/test/)".

It seems like there was latent consensus that tests which can only be run in one browser engine and will never be able to run in multiple engines, should not be included in WPT. There was even linting recently written to address a case of this: https://github.com/web-platform-tests/wpt/pull/18509.

Hexcles:
Tangentially, we also discussed the possibility of "tentative" directories and correspondingly a guidance like e.g. "a directory for a new API with only an explainer without a draft spec should be marked as tentative". I'm not sure if we should have a different RFC for "tentative" directories though cc @jgraham .

foolip:
@boazsender can I assign this to you? I don't think @web-platform-tests/wpt-core-team feel we are on the hook to resolve issues in this repo, just PRs.

Hexcles:
Also in regards to the specifics of MojoJS: https://github.com/web-platform-tests/wpt/issues/16455#issuecomment-532040838

boazsender:
@foolip: sure thing!

foolip:
@boazsender I'm going to go ahead and close this issue now, but if you do still want to send an RFC for it, certainly that would be welcome!

PR #215, opened 2024-12-06
Add a quick and dirty draft of what might be the de-facto scope of WPT

gsnedders:


rfcs/scope-of-wpt.md
```diff
@@ -0,0 +1,60 @@
...
+ * Web browser behavior currently being explored via an
+   [explainer](https://tag.w3.org/explainers/),
+   but without a specification yet written.
+   (XXX: require prototyping to have started?)
+
+ * Historic but unspecified features in web browsers,
+   especially where major browsers are interoperable,
+   where there exists a long-term intention to specify them.
+   (Note: this excludes features which have been deliberately removed from specifications.
+    These are explicitly out of scope.)
```
gsnedders, line 44: This arguably doesn't actually match up with what [the documentation currently says](https://web-platform-tests.org/writing-tests/file-names.html#:~:text=.tentative,-%3A%20):

> Indicates that a test makes assertions not yet required by any specification, or in contradiction to some specification. This is useful when implementation experience is needed to inform the specification. It should be apparent in context why the test is tentative and what needs to be resolved to make it non-tentative.

The "in contradiction to some specification" is kinda troubling — this implies that if some spec change is made, with no expectation of implementations changing behaviour any time soon, you could change the existing tests to be tentative and add new non-tentative tests for the new behaviour. IMO, that should be against policy — anything that is tentative should be on a path to becoming non-tentative. We should probably be explicit in the scope of what can be tentative that everything which _is_ tentative should be on a path to becoming non-tentative.

I don't think the two things listed here are a complete list of cases; we also have cases:

 * where there are WG resolutions not yet applied to the spec making tests tentative (e.g., [`/css/css-overflow/line-clamp/line-clamp-005.tentative.html`](https://github.com/web-platform-tests/wpt/blob/73fd679bba753839932a62a8b33988e79494f088/css/css-overflow/line-clamp/line-clamp-005.tentative.html) is an example of this, with https://github.com/w3c/csswg-drafts/issues/7708#issuecomment-1758081281),
 * where there is a proposal to change the spec to some other behaviour (e.g., [`svg-aam/role/role-img.tentative.html`](https://github.com/web-platform-tests/wpt/blob/73fd679bba753839932a62a8b33988e79494f088/svg-aam/role/role-img.tentative.html#L17-L19), with https://github.com/w3c/svg-aam/issues/32),
 * where there is a proposed addition to the spec (e.g., [`/html/dom/elements/global-attributes/the-anchor-attribute-001.tentative.html`](https://github.com/web-platform-tests/wpt/blob/73fd679bba753839932a62a8b33988e79494f088/html/dom/elements/global-attributes/the-anchor-attribute-001.tentative.html), with https://github.com/whatwg/html/pull/9144),
 * where the spec is unclear (e.g, [`webdriver/tests/bidi/input/release_actions/sequence_tentative.py`](https://github.com/web-platform-tests/wpt/blob/73fd679bba753839932a62a8b33988e79494f088/webdriver/tests/bidi/input/release_actions/sequence_tentative.py), with https://github.com/w3c/webdriver/issues/1859).

The other thing that stands out is:

```
% rg --glob '*tentative*' 'https?://((bugs\.webkit\.org|bugs\.chromium\.org|crbug\.com|issues\.chromium\.org)/|github\.com/.*/(issue|pull))' --files-without-match | wc -l
    2078
```

Which shows it's very often _not_ obvious "why the test is tentative and what needs to be resolved to make it non-tentative", though that is really a separate issue. (And looking at a few random tests, the commit history often doesn't make it obvious either.)

gsnedders:
See also https://github.com/web-platform-tests/rfcs/issues/30 for an earlier issue about this, which got closed without anything ever happening.

rfcs/scope-of-wpt.md
```diff
@@ -0,0 +1,60 @@
...
+and make it easier for others to understand what is and isn't reasonable to add.
+
+The primary aim of web-platform-tests is to test
+[web specifications](https://github.com/w3c/browser-specs/blob/11a71b738f5e41f9239fdcd2074153388c8c6b8b/README.md#spec-selection-criteria)
+[intended to be implemented by browsers](https://github.com/w3c/browser-specs/blob/11a71b738f5e41f9239fdcd2074153388c8c6b8b/README.md#categories),
+both RFC 2119 "must" and RFC 2119 "should"
+implementation requirements.
+
+A specification does not need to have cross-vendor support for its tests to be included in web-platform-tests.
+(XXX: Should we set a bar of one-vendor support or are we okay with zero-vendor support?)
```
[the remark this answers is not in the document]
past: Then again, the discussion around tentative tests below indicates that this wouldn't be a new problem.

rfcs/scope-of-wpt.md
past: I appreciate this analysis and I wonder if we should separate adding more clarity on tentative tests from the rest of this RFC. I.e. let's not attempt to boil the ocean :-)

rfcs/scope-of-wpt.md
```diff
@@ -0,0 +1,60 @@
...
+[optional tests](https://web-platform-tests.org/writing-tests/file-names.html#:~:text=.optional,-%3A%20)
+or
+[optional subtests](https://web-platform-tests.org/writing-tests/testharness-api.html#optional-features).
+
+
+### Risks
+
+We over-constrain what is allowed in web-platform-tests,
+potentially raising the bar to change what is allowed to "submit a new RFC",
+leading browser vendors to reduce what they submit to the project.
```
past, line 60: I think that there is no bar to submitting an RFC, any and all change suggestions should be welcome. I suspect you meant that there is a risk of raising the bar to adding more tests _without_ sending an RFC first, because adding said tests would require us to relax some of these rules?

past (approved):
This LGTM as a clarification of current practice.

PR #218, opened 2025-01-07
RFC #218: Python vendoring

gsnedders:
@Ms2ger suggested I file an RFC for this.

[Rendered](https://github.com/gsnedders/rfcs/blob/vendoring/rfcs/vendoring.md).

Implementation: https://github.com/web-platform-tests/wpt/pull/49752.

### DIFF
--- a/rfcs/vendoring.md
+++ b/rfcs/vendoring.md
@@ -1,0 +1,142 @@
+# RFC XXX: Migrate Python vendoring to vendoring
+
+## Summary
+
+We should use [vendoring](https://pypi.org/project/vendoring/) to re-do all of our Python vendoring,
+directly into `third_party` (crudely: `pip install -t tools/third_party ...`),
+and then move to a generated `requirements_vendor.txt`.
+
+[Implementation](https://github.com/web-platform-tests/wpt/pull/49752).
+
+
+## Background
+
+We currently have a directory in WPT, called `tools/third_party`,
+which contains various bits of vendored code,
+which we copy into our tree primarily to
+[avoid Mozilla having to update their partial PyPI mirror](https://github.com/web-platform-tests/rfcs/issues/82#issuecomment-2462913061).
+
+This currently contains copies of a number of Python libraries as well as pdf.js,
+currently mostly imported via `git-subtree`.
+
+It's difficult to know what exact versions we have vendored,
+and it is hard to update them,
+and especially hard to know when we no longer need libraries we vendor
+(especially those we vendor only as dependencies of other libraries).
+
+The Python code is added to `sys.path` via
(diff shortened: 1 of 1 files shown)


jcscottiii:
@WeizhongX I don't think this will impact Chromium CI but I thought I would double check with you.

WeizhongX:
> @WeizhongX I don't think this will impact Chromium CI but I thought I would double check with you.

We might need to create vpython wheel for the vendor packages, but should be fine for us.

jgraham:
This generally seems fine to me, and looks like a significant improvement over the status quo. I have the following comments:
* I think the risk of this affecting Mozilla's prioritization of improving our story around non-vendored deps is minimal. Most of the pressure there really comes from deps with binary components, which aren't changed either way by this proposal.
* I think there is some risk that `vendoring` changes in the future in a way that breaks our use. It's specifically documented as being unstable and only designed to work with pip. I think it's a managable risk, but a real one.
* There is also the risk that this breaks unshared tests using wpt, if they are doing custom path manipulations. That isn't any kind of blocker, but people should be aware, since it may affect syncing.
* The implementation should be tested with full runs of wpt using the trigger branches, not just relying on CI jobs, since there's a risk of any test that does something custom with the Python path breaking.

jcscottiii (approved):

rfcs/vendoring.md
```diff
@@ -0,0 +1,142 @@
+# RFC XXX: Migrate Python vendoring to vendoring
```
Ms2ger, line 1:
```suggestion
# RFC 218: Migrate Python vendoring to vendoring
```

gsnedders:
> * I think there is some risk that `vendoring` changes in the future in a way that breaks our use. It's specifically documented as being unstable and only designed to work with pip. I think it's a managable risk, but a real one.

This is fair, but it's essentially "we now use a tool to automate most of the work that may break in future" (and really it isn't a complex tool) versus "we don't really document how to vendor new things and don't really give people guidance on how to do it".

Like, the ultimate fallback is just directly doing [what vendoring does](https://github.com/pradyunsg/vendoring/blob/b7436be6e28116b7cef74127baf1118c319bb30f/src/vendoring/tasks/vendor.py#L17-L30):

```python
    command = [
        "pip",
        "install",
        "--platform",
        "any",
        "-t",
        str(destination),
        "-r",
        str(requirements),
        "--no-compile",
        # We use --no-deps because we want to ensure that dependencies are provided.
        # This includes all dependencies recursively up the chain.
        "--no-deps",
    ]
```

Though this means we then miss things like "check we get all the licenses", and automatically applying patches, and potentially rewriting module names.

> * The implementation should be tested with full runs of wpt using the trigger branches, not just relying on CI jobs, since there's a risk of any test that does something custom with the Python path breaking.

Per https://github.com/web-platform-tests/wpt/pull/49752#issuecomment-2637966394 this all looks like pre-existing flakiness to me.

jgraham (approved):

PR #219, opened 2025-02-19
Add support for testing the WebExtensions API in WPT

kiaraarose:


rfcs/web_extensions.md
```diff
@@ -0,0 +1,57 @@
+# RFC #219: Add support for WebExtensions in WPT
+
+## Summary
+
+The WebExtensions API extends the capabilities of the browser. Adding support
+for WebExtensions in the web platform tests will increase interoperability
+and will help drive the standardization of this API.
+
+This RFC proposes adding a new testharness test type `.extension.js` to handle|
```
elijahsawyers, line 9:
```suggestion
This RFC proposes adding a new `testharness.js` test type, `.extension.js`, to handle
```

rfcs/web_extensions.md
```diff
@@ -0,0 +1,57 @@
...
+testing this API, in addition to using `test_driver` to load and unload extensions.
```
elijahsawyers, line 10:
```suggestion
testing this API, in addition to using `testdriver.js` to load and unload extensions.
```

Ditto throughout the file.

rfcs/web_extensions.md
```diff
@@ -0,0 +1,57 @@
...
+  
+Using test_driver, we add support for testing the WebExtensions API by loading
+a web extension designed to test the functionality of a specific API.
+The extension will be loaded after the tests begins, and unloaded before the
+test is finished.
+
+Most of the test execution is handled within the extension, via the 
+[browser.test](https://chromium.googlesource.com/chromium/src.git/+/master/extensions/docs/testing_api.md)
+API. We’ve elected to use these APIs since all participating browser vendors use
+`browser.test` to test the web extensions internally and they can easily port over
```
elijahsawyers, line 22:
```suggestion
`browser.test` internally, and they can easily port over
```

rfcs/web_extensions.md
```diff
@@ -0,0 +1,57 @@
...
+existing tests to the web platform tests.
+
+Because these tests won’t leverage testharness.js directly, we’ve introduced a new
```
elijahsawyers, line 25:
```suggestion
Because these tests won’t leverage `testharness.js` directly, we’ve introduced a new
```

rfcs/web_extensions.md
```diff
@@ -0,0 +1,57 @@
...
+testharness test type `.extension.js`  that will create the necessary boilerplate to
```
elijahsawyers, line 26:
```suggestion
`testharness.js` test type, `.extension.js`,  that will create the necessary boilerplate to
```

rfcs/web_extensions.md
```diff
@@ -0,0 +1,57 @@
...
+convert the `browser.test` assertions into the corresponding assertions in the test
+harness.
+   
+A proposed patch is available at [insert WPT link here]
```
elijahsawyers, line 30:
```suggestion
A proposed patch is available [here](https://github.com/web-platform-tests/wpt/pull/50648).
```

rfcs/web_extensions.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Alternatives considered
+
+We considered loading the extension statically before each testharness test is run,
```
elijahsawyers, line 34:
```suggestion
We considered loading the extension statically before each `testharness.js` test is run,
```

rfcs/web_extensions.md
```diff
@@ -0,0 +1,57 @@
...
+but we decided against that since test_driver will allow the tests to drive the
```
elijahsawyers, line 35:
```suggestion
but we decided against that since `testdriver.js` will allow the tests to drive the
```

rfcs/web_extensions.md
```diff
@@ -0,0 +1,57 @@
...
+browser, and in the future, test functionality such as opening popups and clicking
+menu items. 
+
+We considered adding a new test type other than using JavaScript tests, but we decided
```
elijahsawyers, line 39: We haven't been referring to the tests has "JavaScript tests". We should stick to using `testharness.js`.
```suggestion
We considered adding a new test type rather than using `testharness.js`, but we decided
```

rfcs/web_extensions.md
```diff
@@ -0,0 +1,57 @@
...
+against it because it was a lot more work compared to using testharness.js.
```
elijahsawyers, line 40:
```suggestion
against it because it was a lot more work compared to using `testharness.js`. 
```

rfcs/web_extensions.md
```diff
@@ -0,0 +1,56 @@
+# RFC #219: Add support for WebExtensions in WPT
+
+## Summary
+
+The WebExtensions API extends the capabilities of the browser. Adding support
+for WebExtensions in the web platform tests will increase interoperability
+and will help drive the standardization of this API.
+
+This RFC proposes adding a new `testharness.js` test type, `.extension.js`, to handle|
```
elijahsawyers, line 9:
```suggestion
This RFC proposes adding a new `testharness.js` test type, `.extension.js`, to handle
```

rfcs/web_extensions.md
```diff
@@ -0,0 +1,56 @@
...
+Using `testdriver.js`, we added support for testing the WebExtensions API by loading
+a web extension designed to test the functionality of a specific API.
+The extension will be loaded after the tests begins, and unloaded before the
+test is finished.
+
+Most of the test execution is handled within the extension, via the 
+[browser.test](https://chromium.googlesource.com/chromium/src.git/+/master/extensions/docs/testing_api.md)
+API. We’ve elected to use these APIs since all participating browser vendors use
+`browser.test`  internally and they can easily port over existing tests to
+the web platform tests.
```
gsnedders, line 23:
```suggestion
web-platform-tests.
```

rfcs/web_extensions.md
```diff
@@ -0,0 +1,56 @@
...
+
+Because these tests won’t leverage `testharness.js` directly, we’ve introduced a new
+`testharness.js`, `.extension.js`, that will create the necessary boilerplate to
+convert the `browser.test` assertions into the corresponding assertions in the test
+harness.
```
gsnedders, line 28: We should probably care to some extent about how easy/difficult it is to write tests _without_ using `extension.js`.

Rather than putting all the code necessary to map from `browser.test` to `testharness.js` in the server-generated wrapper, it would probably be better to put it into a separate file — which then gets included (along with testharness and testdriver) by the server-generated wrapper?

OrKoN: I have actually tried to implement the proposal for `executorsafari.py` in https://github.com/web-platform-tests/wpt/pull/50648#discussion_r1970468555 and I think it would be easier to integrate the tests without adding `extension.js` and using regular testharness.js tests. I think something like this would work:

```
// META: script=/resources/testdriver.js?feature=bidi
// META: script=/resources/testdriver-vendor.js
// META: script=/web-extension/resources/webextension-test.js

webextension_test(async () => {
    // Call for illustrative purposes, in practice, webextension_test
    // defined in webextension-test.js can implement it. 
    await test_driver.bidi.webExtension.install();
}, test_desc);
```

A problem with the current approach is that the infrastructure does not detect the `?feature=bidi` inside the server-generated wrapper.

OrKoN: here is a runnable implementation of this https://github.com/web-platform-tests/wpt/pull/51090

rfcs/web_extensions.md
```diff
@@ -0,0 +1,56 @@
...
+menu items. 
+
+We considered adding a new test type rather than using JavaScript tests, but we decided
+against it because it was a lot more work compared to using `testharness.js`.
+   
+## Risks
+
+There are two potential concerns with this implementation:
+
+1. We have no precedent for tests run via a Classic command in some user agents
```
OrKoN, line 46: I have concerns that we will end up with two methods for dealing with extensions exposed to test authors following different (and separately evolving) specs that are not fully implemented by all vendors. I think it would be preferable to start with adding methods to testdriver.js that match WebDriver BiDi and allow synchronous handling of these methods for non-WebDriver-BiDi implementations (prototype: https://github.com/web-platform-tests/wpt/pull/51090).

rfcs/web_extensions.md
```diff
@@ -0,0 +1,56 @@
+# RFC #219: Add support for WebExtensions in WPT
+
+## Summary
+
+The WebExtensions API extends the capabilities of the browser. Adding support
+for WebExtensions in the web platform tests will increase interoperability
+and will help drive the standardization of this API.
+
+This RFC proposes adding a new `testharness.js` test type, `.extension.js`, to handle
+testing this API, in addition to using `testdriver.js` to load and unload extensions.
```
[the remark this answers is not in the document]
kiaraarose: will change this to say "wrapper type"

rfcs/web_extensions.md
kiaraarose: We ended up discussing this in the Browser Testing and Tools WG last Tuesday, and it was decided that using the classic approach for `testdriver.js` was ok, and Chrome can continue to use the BiDi implementation in the backend

rfcs/web_extensions.md
kiaraarose: > Rather than putting all the code necessary to map from browser.test to testharness.js in the server-generated wrapper, it would probably be better to put it into a separate file — which then gets included (along with testharness and testdriver) by the server-generated wrapper?

@gsnedders what would be the benefits of that approach?

gsnedders: > I think it would be easier to integrate the tests without adding extension.js and using regular testharness.js tests. I think something like this would work:

This would effectively be going from:

```javascript
runTestsWithWebExtension("/resources/runtime/")
```

To:

```javascript
// META: script=/resources/testdriver.js?feature=bidi
// META: script=/resources/testdriver-vendor.js
// META: script=/web-extensions/resources/web-extension-test.js

webextension_test("/resources/runtime/")
```

And that doesn't seem particularly unreasonable.

The real benefit of `.extension.js` would be for cases where the background script is the only file in the extension (and I think that is true for most extension tests that vendors have today?), if _that_ script was an `.extension.js` file, and then generate a manifest and a testharness wrapper for that. It might be worth leaving that as future work, beyond the initial RFC and implementation?

> A problem with the current approach is that the infrastructure does not detect the ?feature=bidi inside the server-generated wrapper.

Given both the manifest generation and the wrapper are versioned together within `/tools`, that doesn't seem like a problem — we just pass the appropriate `testdriver_features` to the `TestharnessTest` in `sourcefile.py`. There's potential concerns about future features we might add and use within the tests, and how we'd want to flag them.

gsnedders: > > Rather than putting all the code necessary to map from browser.test to testharness.js in the server-generated wrapper, it would probably be better to put it into a separate file — which then gets included (along with testharness and testdriver) by the server-generated wrapper?
> 
> @gsnedders what would be the benefits of that approach?

We spoke about this offline, but to summarise here: just being able to have full control of the HTML, if/when needed.

The only to-do for that is just moving the large inline script in https://github.com/web-platform-tests/wpt/blob/18384b4af964a99e2f2138ab3bb7bbee1b6eb4b3/tools/serve/serve.py#L328 into a file of its own.

gsnedders (commented):
I think the RFC should probably also address the question of "do we want to add a new [testdriver feature](rfcs/testdriver-features.md)", given that's important to Chromium where running these tests will inherently depend on BiDi, even though that's not true in other implementations.

Just using the "bidi" flag is potentially problematic if we want to closely tie that to BiDi  usage generally, given other implementations will run these tests without BiDi.

rfcs/web_extensions.md
```diff
@@ -0,0 +1,119 @@
...
+a `browser.test.assert...()` method, it is up to the user agent to send the message "assert"
+or "assert-equality". 
+
+Because these tests don’t leverage `testharness.js` directly, we’ve introduced a new
+wrapper type, `.extension.js`, that will create the necessary boilerplate to
+convert these messages into the corresponding assertions in the `testharness.js`.
+   
+A proposed patch is available at https://github.com/web-platform-tests/wpt/pull/50648.
+
+##  `browser.test` API
```
OrKoN, line 34: how does the browser know if the `browser.test` globals need to be exposed?

rfcs/web_extensions.md
OrKoN: I believe it was the `wpt RFCs and infrastructure` and not the Browser Testing and Tools WG meeting. Could you please add details to the RFC on how using the proposed testdriver.js API would be possible for implementations relying on WebDriver BiDi?

rfcs/web_extensions.md
```diff
@@ -0,0 +1,119 @@
...
+when a test has started and finished. For these cases, it is up to the user agent to send the message
+"test-started" and "test-finished", respectively. In addition, when recording the result for 
+a `browser.test.assert...()` method, it is up to the user agent to send the message "assert"
+or "assert-equality". 
+
+Because these tests don’t leverage `testharness.js` directly, we’ve introduced a new
+wrapper type, `.extension.js`, that will create the necessary boilerplate to
+convert these messages into the corresponding assertions in the `testharness.js`.
+   
+A proposed patch is available at https://github.com/web-platform-tests/wpt/pull/50648.
```
OrKoN, line 32: Is the PR in sync with the RFC? I wonder if RFC should also discuss how the runner would be resolving path to the extension. It appears that one option is to resolve extension path relative to the test path and another option is to require tests to provide path relative to the WPT root folder?

rfcs/web_extensions.md
kiaraarose: In Safari, we expose the `browser.test` namespace when in a testing environment. This is true for WPT tests. Chrome and Firefox have agreed to do the same.

rfcs/web_extensions.md
kiaraarose: In the PR, I have it set to get the extension path relative to the test path [here](https://github.com/web-platform-tests/wpt/pull/50648#discussion_r1957778696). i can add a comment in the RFC detailing that

rfcs/web_extensions.md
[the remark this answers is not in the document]
kiaraarose: That is the goal. There are still APIs within this namespace that would need documentation. I only added documentation for the APIs used in the PR.

rfcs/web_extensions.md
kiaraarose: > The only to-do for that is just moving the large inline script in https://github.com/web-platform-tests/wpt/blob/18384b4af964a99e2f2138ab3bb7bbee1b6eb4b3/tools/serve/serve.py#L328 into a file of its own.

I updated the PR to do this.

rfcs/web_extensions.md
```diff
@@ -0,0 +1,119 @@
...
+Asserts that a given condition is `false`. If the condition is `true`, the test fails.
+
+**Parameters**
+
+* `condition` (boolean) – The condition to assert. If `true`, the test fails.
+* `message` (string, optional) – A message describing the assertion.
+
+\
+**`browser.test.assertEq(actual, expected, message)`**\
+Asserts that the actual value is equal to the expected value. If not equal, the test fails.
```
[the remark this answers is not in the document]
kiaraarose: It matches ES's `===`. I haven't tested this yet, but I believe our implementation has `NaN === NaN` resulting in `false`. we intend to have the implementation of this API follow ES SameValue so I can discuss this with the other browsers and ensure that we align on having `NaN` equal `NaN` and `-0` not equal `+0`.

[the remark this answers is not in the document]
kiaraarose: I can also update the documentation to reflect this.

zombie (approved):
I have confirmed this design would work for Gecko's implementation of WebExtensions, I have a stack of patches that are ready to be submitted for review once this is rfc is approved.

rfcs/web_extensions.md
OrKoN: Does it mean that the extension-specific globals would be exposed for non-extension WPT tests? Is the `browser` global variable extension-specific or is it used for other web features too? I wonder if it could cause problems for other WPT tests by exposing an unexpected global? given there the proposal now includes a feature flag for extensions perhaps browsers should only expose the globals if the feature=extensions is on? cc @oliverdunk if you have some more details on how Chromium would expose the test interface, it would be useful.

OrKoN (approved):
I think Chromium will be able to support these tests using the WebDriver BiDi implementation with the current proposal. Optional feedback and some notes:

- `feature=extensions` allows Chromium to switch over to WebDriver BiDi implementation but the PR might require some changes (I encountered some errors trying to run the PR so I could not test it). The feature flag could also serve as indication if the test globals need to be exposed to not potentially affect other tests.
- When discussing internally, we were not sure if a new wrapper type (.extension.js) is really beneficial given that the test code runs in a page similar to the `.window.js` tests and not in an extension. 
- I still do not think that there is a need for an alternative WebDriver Classic specification that is not meant to become the standard. Even if we end up linking to it, I think it'd be good to put methods under the `extensions` namespace in testdriver.js `test_driver.web_extension.install/test_driver.web_extension.uninstall` to add future WebDriver BiDi methods there too if needed (e.g., https://github.com/w3c/webdriver-bidi/issues/898). I do not have a strong opinion here but it was brought up in discussions. 

@oliverdunk I have not reviewed browser.test APIs but I assume Chromium will work with what is proposed in this doc.

rfcs/web_extensions.md
```diff
@@ -0,0 +1,137 @@
...
+
+**Parameters**
+
+* `condition` (boolean) – The condition to assert. If `true`, the test fails.
+* `message` (string, optional) – A message describing the assertion.
+
+\
+**`browser.test.assertEq(actual, expected, message)`**\
+Asserts that the actual value is equal to the expected value. If not equal, the test fails.\
+This matches the ES [SameValue](https://tc39.es/ecma262/#sec-samevalue) implementation.
```
rdcronin, line 75: This is not true in Chromium-based browsers.  In Chromium-based browsers, we have custom handling for anything with `typeof` equal to `"object"`, which then results in a "deep equals" comparison (code [here](https://source.chromium.org/chromium/chromium/src/+/main:extensions/renderer/resources/test_custom_bindings.js;l=248-252;drc=abe36fd704986c2e4f54b4bd35c291d6beb0e28c)).  This is important, since it enables things like:

```
browser.test.assertEq({foo: 3, bar: 2}, value);
```

rfcs/web_extensions.md
```diff
@@ -0,0 +1,137 @@
...
+**`browser.test.assertFalse(condition, message)`**\
+Asserts that a given condition is `false`. If the condition is `true`, the test fails.
+
+**Parameters**
+
+* `condition` (boolean) – The condition to assert. If `true`, the test fails.
+* `message` (string, optional) – A message describing the assertion.
+
+\
+**`browser.test.assertEq(actual, expected, message)`**\
```
rdcronin, line 73: It doesn't matter _too_ much, but note that in Chromium-based browsers, the ordering of these is "expected, actual", not "actual, expected"

rfcs/web_extensions.md
```diff
@@ -0,0 +1,137 @@
...
+* `message` (string, optional) – A message describing the assertion.
+
+\
+**`browser.test.runTests(tests)`**\
+Queues an array of test functions to run and returns a promise.
+The promise is rejected if one of the tests fails, otherwise resolves.
+
+**Parameters**
+
+* `tests` (array) –  An array of functions.
```
rdcronin, line 100: This doesn't define how we move from one test to another, which IIRC is something we a) do want to align on and b) would require work in Chromium.  Is that in-scope to include here?

rdcronin (commented):
Thank you, Kiara!

rfcs/web_extensions.md
```diff
@@ -0,0 +1,137 @@
...
+
+\
+**`browser.test.assertThrows(fn, message)`**\
+Asserts that a given function throws an error when executed. If the function does not throw,
+the test fails.
+
+**Parameters**
+
+* `fn` (function) – A function expected to throw an error.
+* `message` (string, optional) – A message describing the assertion.
```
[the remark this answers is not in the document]
xeenon: WebKit also has `expectedError` as a RegEx or String for the second parameter.

* Interface: https://github.com/WebKit/WebKit/blob/59247a7628dd45966bd71f2a00523ed891650aa7/Source/WebKit/WebProcess/Extensions/Interfaces/WebExtensionAPITest.idl#L77
 * Implementation: https://github.com/WebKit/WebKit/blob/59247a7628dd45966bd71f2a00523ed891650aa7/Source/WebKit/WebProcess/Extensions/API/Cocoa/WebExtensionAPITestCocoa.mm#L328-L339
  * Example usage: https://github.com/WebKit/WebKit/blob/59247a7628dd45966bd71f2a00523ed891650aa7/Tools/TestWebKitAPI/Tests/WebKitCocoa/WKWebExtensionAPITabs.mm#L108-L118

rfcs/web_extensions.md
kiaraarose: i agree that it does't matter too much. since "actual, expected" matches what WPT uses in `testharness.js`, i'll keep this as is

rfcs/web_extensions.md
zombie: Both we and Safari have a separate `test.assertDeepEq()` for this use case.

I don't have a strong reason to oppose changing the `assertEq()` behavior to match Chromium, but there's something to be said for less magic and more explicitness in API design.

rfcs/web_extensions.md
```diff
@@ -0,0 +1,137 @@
...
+by using the `test.path` before the protocol loads the extension.
+
+A proposed patch is available at https://github.com/web-platform-tests/wpt/pull/50648.
+
+##  `browser.test` API
+
+The `browser.test` API provides utilities for writing and running tests for WebExtensions.
+It includes assertion functions and a test harness to facilitate structured and reliable testing.
+
+### Methods
```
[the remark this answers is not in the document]
zombie: Agree that we want to end up with ~all of them, but there's currently little alignment between implementations beyond what's currently in this PR, even the one's you mention, Firefox and Safari name it `test.assertRejects()` and neither have the `assertNe()`.

I suggest not blocking on specifying everything in one go, going with the MVP set that's already agreed, and taking incremental steps to discuss further methods (probably in our own repo) and then bringing them to WPT.

rfcs/web_extensions.md
kiaraarose: For Safari's implementation, the tests run in the order they appear in the array. Once one test finishes, then we move onto the next test. does this match Firefox's behavior @zombie

rfcs/web_extensions.md
```diff
@@ -0,0 +1,125 @@
...
+Asserts that a given function throws an error when executed. If the function does not throw, the test fails. The test will also fail if `expectedError` is defined and the function does not throw an error with a message matching `expectedError`.
+
+**Parameters**
+
+* `fn` (function) – A function expected to throw an error.
+* `expectedError` (any, optional) – A string (which must match exactly) or a RegExp stating the expected error.
+* `message` (string, optional) – A message describing the assertion.
+
+**`browser.test.runTests(tests)`**
+Queues an array of individual test functions to run, executes them in the order they appear, and then returns a promise. The promise is rejected if one of the tests fails; otherwise, it resolves. A test will fail if an assertion fails or if an unexpected error is thrown. All test functions passed into the array will run, even if a previous test failed.
```
Rob--W, line 68: What should be the behavior of `runTests` when it has already been called before?

I think that Chrome's current expectation is for tests to not call it more than once, considering that it replaces all existing tests before starting it off: https://source.chromium.org/chromium/chromium/src/+/main:extensions/renderer/resources/test_custom_bindings.js;l=443;drc=abe36fd704986c2e4f54b4bd35c291d6beb0e28c

In Firefox's WIP implementation (https://phabricator.services.mozilla.com/D244091?id=1016587), the behavior is to append the tests and run them (in parallel!). This doesn't look like ideal behavior either.

@kiaraarose @zombie WDYT of only allowing `browser.test.runTests` to be called once per context?

Alternatively, if this is too restrictive (e.g. if there is a desire to queue more tests after completing existing ones), only allow it to be called when the list of tests is empty.

rfcs/web_extensions.md
```diff
@@ -0,0 +1,125 @@
...
+**"assert" `data` parameters**
+
+* `result` (boolean) – A boolean indicating whether the test assertion passed or failed.
+* `message` (object) – A message describing the assertion.
+
+**"assert-equality" `data` parameters**
+
+* `result` (boolean) – A boolean indicating whether the test assertion passed or failed.
+* `message` (object) – A message describing the assertion.
+* `expectedValue` (any) – The expected value of the object being evaluated.
```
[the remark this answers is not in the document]
kiaraarose: these values are passed to testharness.js assertion methods. so if we have this call
```
browser.test.assertEq(typeof browser.runtime.getURL(""), "string")
```
the expected value of "string" is passed back so that we can use it for the testharness.js assertion method.
e.g.
```
assert_true(data.result, `Expected: ${data.expectedValue}; Actual: ${data.actualValue}; Description: ${data.message}`)
```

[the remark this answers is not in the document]
Rob--W: Should we specify that it is a primitive value, or even always a string, for the purpose of diagnostics/debugging?

From your response it is clear that this is the intended usage.

rfcs/web_extensions.md
```diff
@@ -0,0 +1,137 @@
...
+the test fails.
+
+**Parameters**
+
+* `fn` (function) – A function expected to throw an error.
+* `message` (string, optional) – A message describing the assertion.
+
+\
+**`browser.test.runTests(tests)`**\
+Queues an array of test functions to run and returns a promise.
```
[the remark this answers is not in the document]
kiaraarose: @Rob--W let me know if the specification for `browser.test.runTests` is sufficient. if not, please provide any additional details you think would be necessary to add. thanks!

rfcs/web_extensions.md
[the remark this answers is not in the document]
rdcronin: +1 to having questions about this for any non-primitive types.  Also, what is in `message` that makes it an object?  (I would have expected this to be the strings that are passed into the assertion?)

Taking a step back, do we need this?  Or can the test API code be responsible for formatting messages and just passing them as the failure message?  In its current state, I don't know if this is sufficient to really help debugging, since additional information (e.g. line number) is really valuable.

If this is really necessary and is just for error messages, could we just have the test harness pass the message?

```
* `result` (boolean) – A boolean indicating whether the test assertion passed or failed.
* `assertionDescription` (string) – A string with a human-readable description of the assertion, such as "Assertion Failed: Expected: 'string'; Actual: 'function'"
* `message` (string) (optional) – An additional optional message provided by the test assertion.
```

rdcronin (changes requested):
Thanks, Kiara!

rfcs/web_extensions.md
```diff
@@ -0,0 +1,129 @@
...
+When `assertThrows` is invoked with two parameters, the second value should be interpreted as `expectedError`. This is explicitly mentioned because `expectedError` and `message` are both optional and accepting string values.
+
+**Parameters**
+
+* `fn` (function) – A function expected to throw an error.
+* `expectedError` (any, optional) – A string (which must match exactly) or a RegExp stating the expected error.
+* `message` (string, optional) – A message describing the assertion.
+
+**`browser.test.runTests(tests)`**
+Queues an array of individual test functions to run, executes them serially, and then returns a promise. The promise is rejected if one of the tests fails; otherwise, it resolves. A test will fail if an assertion fails or if an unexpected error is thrown. All test functions passed into the array will run, even if a previous test failed.
```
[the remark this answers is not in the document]
zombie: 3. Is different in Firefox, and from the last meeting I think in Safari as well.  Individual `browser.test.assertEq` assertions don't interrupt the flow of the test -- they are just reported as test failures.  Since the wpt test harness also throws on assertion failures, I'm willing to align on that.

4. Is currently different, we bail on the whole test run if one of the individual tests throws/returns a rejected promise, but I can easily change it since `runTests()` is new in Firefox and nothing depends on it yet.

rfcs/web_extensions.md
zombie: Yes, but see my comment below for more details.

rfcs/web_extensions.md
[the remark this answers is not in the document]
kiaraarose: it should be  ``message` (string)`. i'll update that. i don't feel strongly about having the assertion description message being constructed by the browser or the WPT test harness. if folks think it makes more sense for the browser to construct it, then i can make that change. in terms of passing along useful debugging info, that can be passed in the `message` here.

rfcs/web_extensions.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Details
+
+Using `testdriver.js`, we added support for testing the WebExtensions API by loading a web extension designed to test the functionality of a specific API.
+
+These tests won't leverage `testharness.js` directly since most of the test execution is handled within the extension via the `browser.test` API. We’ve elected to use this API since all participating browser vendors use `browser.test` internally and can easily port over existing tests to the web-platform-tests.
+
+Tests are driven from a web page like a usual wpt test. These web pages have access to a minimal `browser.test` API consisting of the `browser.test.onMessage.addListener` method, which allows the test to receive extension test progress notifications from the user agent. It is up to the user agent to send the messages "test-started" and "test-finished", respectively. These messages should be sent when a test that was passed into `browser.test.runTests` (in the extension) has started and finished. In addition, when recording the result for a `browser.test.assert...()` method, it is up to the user agent to send the messages "assert" or "assert-equality".
```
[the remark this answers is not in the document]
kiaraarose: these messages are being used by the WPT test to know what step to take i.e. start a new test, end the current test, or translate the `browser.test.assert...` method into the testharness.js assert method. this could be re-worked so that the only messages the browser is responsible for sending is "test-started" and "test-finished", and if there isn't a message sent, then call `assert_true` on the result of the `browser.test.assert...` method. the difference between sending "assert" and "assert-equality" was mostly about figuring out if we should construct the assertion description e.g. `"Assertion Failed: Expected: 'string'; Actual: 'function'"`. but if we think that this message should be constructed by the browser, then there wouldn't be a need for this distinction. here is what the code looks like in WPT for reference

```
browser.test.onMessage.addListener((message, data) => {
        switch(message) {
          case "assert":
            test.step(() => {
              assert_true(data.result, data.message)
            })

            break
          case "assert-equality":
            test.step(() => {
              assert_true(data.result, `Expected: ${data.expectedValue}; Actual: ${data.actualValue}; Description: ${data.message}`)
            })

            break
          case "test-started":
            test = async_test(data.testName)

            break
          case "test-finished":
            test.done()

            if (data.result === false)
              test.set_status(test.FAIL)

            if (!data.remainingTests)
                test_driver.uninstall_web_extension(result.extension)
                    .then(() => {
                        done()
                    })

            break
        }
      })
```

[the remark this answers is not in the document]
rdcronin: I think it would make more sense to have all these asserts handled by the extensions test framework, rather than translating them to WPT.  The extensions test framework is the one that will be running the code and have the context (exact values, line numbers, call stacks, etc), which would all be mostly lost within WPT.  Given that, I think having the detailed logs emitted from the extensions test framework, and just a generic message from WPT for "Test failed" (maybe with the final log) would make the most sense, rather than trying to proxy them all through WPT.

rfcs/web_extensions.md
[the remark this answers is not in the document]
kiaraarose: 3. Similar to Firefox, we don't end the current test and move onto the next if an assertion fails. I'd be willing to align with your suggestion here since WPT stops reporting the assertion results once one test fails

rfcs/web_extensions.md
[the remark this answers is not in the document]
kiaraarose: I think that's fair, and doable for Safari.

rfcs/web_extensions.md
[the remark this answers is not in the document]
kiaraarose: 3. Actually, how does this approach work in practice? If the method is already executing, how can it be halted mid execution? is this something that Chrome does today for `runTests` @rdcronin ?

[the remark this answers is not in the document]
zombie: We can throw an assert exception from individual `browser.test.assertX()` methods, I believe that's what other test runners/harnesses do.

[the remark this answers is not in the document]
kiaraarose: that makes sense. thanks!

rfcs/web_extensions.md
kiaraarose: I think it should be allowed to be called more than once and the tests should be queued. although, i'm not sure why the queue would need to be empty in order for the function to be called again. wouldn't a second call just append the tests to the queue?

zombie: I believe Chromium's implementation overwrites the task list on second call, so for simplicity and easier alignment, I suggested allowing calling of `runTests()` again after the previous run resolves, which should be easy to just `await`.

rfcs/web_extensions.md
```diff
@@ -0,0 +1,164 @@
...
+A test is considered **passed** if:
+- It returns `undefined` (for synchronous tests), or
+- It returns a resolved promise (for asynchronous tests).
+
+A test is considered **failed** if:
+- It throws an exception,
+- It returns a rejected promise, or
+- It triggers an assertion failure via the `browser.test.assert...` methods.
+
+Uncaught exceptions and unhandled promise rejections that occur during the execution of a test
```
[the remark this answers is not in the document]
zombie: I expect you mean "uncaught asynchronous exception", like if a `setTimeout()` callback or promise handler throws/rejects (because main call-stack exceptions/awaited rejected promises would bubble-up as a rejected promise from the test function)?

Yeah, we currently don't fail tests in Firefox for those, so this sounds fine (though I wouldn't be opposed if we wanted to later start monitoring them for the purpose of more reliable tests).

rfcs/web_extensions.md
```diff
@@ -0,0 +1,187 @@
...
+
+\
+**`browser.test.assertThrows(fn, message)`**\
+Asserts that a given function throws an error when executed. If the function does not throw,
+the test fails.
+
+**Parameters**
+
+* `fn` (function) – A function expected to throw an error.
+* `message` (string, optional) – A message describing the assertion.
```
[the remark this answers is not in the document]
zombie: > In Chrome, this is the expected error, not a description of the assertion. It sounds like that's the same in Safari, but differs in Firefox.

No, in Firefox the (now optional) second argument is also the expected error (string for exact or regex for matching), and third (always optional) is a string description.  I think we're all aligned here, just a miscommunication in the text.

rfcs/web_extensions.md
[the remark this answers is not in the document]
zombie: I think that's already covered by
> A test is considered **failed** if:
> - It throws an exception,

rfcs/web_extensions.md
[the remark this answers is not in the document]
kiaraarose: will correct that in the text

rfcs/web_extensions.md
```diff
@@ -0,0 +1,187 @@
...
+
+* `testName` (string) – The name of the test being run.
+
+**"test-finished" `data` parameters**
+
+* `testName` (string) – The name of the test that ran.
+* `result` (boolean) – A boolean indicating whether the test passed or failed.
+* `remainingTests` (number) – The number of remaining tests to run.
+* `message` (string, optional) – An additional message provided by the test assertion.
+* `assertionDescription` (string) – A human-readable description, such as "Assertion Failed: Expected: 'string'; Actual: 'function'".
```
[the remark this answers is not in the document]
kiaraarose: 3. using the message sent along can be a distinguisher (i.e. "test-started" or "test-finished")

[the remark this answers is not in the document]
xeenon: I'm in favor of `browser.test.onTestStarted` and `browser.test.onTestEnded`.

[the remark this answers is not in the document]
zombie: Other than redundancy of `test.onTest*`, that sounds good to me.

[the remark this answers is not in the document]
kiaraarose: @rdcronin @zombie Updated the RFC with these new listeners

rfcs/web_extensions.md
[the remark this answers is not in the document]
gsnedders: I think from a Core Team point of view this is our only real concern with the RFC as it stands; splitting out this section into some CG document would be an improvement.

If that can be done quickly, that would be ideal to do before merging.

rfcs/web_extensions.md
```diff
@@ -0,0 +1,199 @@
...
+**Parameters**
+
+* `actual` (any) – The actual value.
+* `expected` (any) – The expected value.
+* `message` (string, optional) – A message describing the assertion.
+
+\
+**`browser.test.assertThrows(fn, expectedError, message)`**
+Asserts that a given function throws an error when executed. If the function does not throw, the test fails.
+The test will also fail if `expectedError` is defined and the function does not throw an error with a message matching `expectedError`.
```
jgraham, line 93: I think for web-platform-tests `expectedError` must not be used unless the error messages are more standardized than normal JS/DOM exceptions, since otherwise it's embedding browser-specific expectations into the test.

rfcs/web_extensions.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Summary
+
+The WebExtensions API extends the capabilities of the browser. Adding support for WebExtensions in
+the web-platform-tests will increase interoperability and will help drive the standardization of
+this API.
+
+This RFC proposes adding a new `testharness.js` wrapper type, `.extension.js`, to handle testing this
+API, in addition to using `testdriver.js` to load and unload extensions. A new `feature=extensions`
+was also introduced to support enabling `BiDi` specifcally for user agents who use it for the
```
jgraham, line 11: Just as a style note, these are usually written in the present or future tense rather than the past, as a description of what the proposal is and what changes it will make.

rfcs/web_extensions.md
```diff
@@ -0,0 +1,199 @@
...
+order they appear in the array. If multiple tests are passed, a new test begins once the previous
+tests completes. If a test fails—either due to a failed assertion or an unexpected exception—execution
+of that test stops, and the next test in the queue begins.
+
+After the returned promise is settled, this method can be called again to run additional tests.
+Calling it while test are still running throws or returns a rejected promise.
+
+The returned promise:
+- **Resolves** if all tests complete successfully.
+- **Rejects** if any test fails.
```
jgraham, line 118: This is obviously your existing API, but for web-platform-tests you really want to structure things so that tests are as independent as possible i.e. later tests running doesn't depend on earlier tests passing. We generally have the semantics that asserts fail the test and stop further execution of that one test, but as far as possible don't prevent future tests running. Otherwise you can get into a situation where a browser that would pass all but one of the subtests in a specific file actually stops after the first (failing) subtest.

jgraham (approved):
This is broadly fine, and although I have some feedback, it's not blocking.

I also think that it would be ideal to move the test API to an actual spec in the WebExt WG (or at least a note in the CG), rather than trying to document it here. But in the absence of that hopefully this is enough to allow people to build more or less interoperable implementations.

rfcs/web_extensions.md
kiaraarose: This documentation was moved [here ](https://github.com/w3c/webextensions/blob/main/proposals/browser_test_api.md) but the `runTests` method runs all of the tests, regardless if a previous test failed. The promise returned will reject if any test fails, but each test is guaranteed to run and the results for each test will be reported.

zombie: Yes, that's our intention as well, we clarified the wording in https://github.com/w3c/webextensions/blob/main/proposals/browser_test_api.md

rfcs/web_extensions.md
zombie: Thanks, yeah we don't expect to use this often, or only with a targeted regex for specific keywords.

jcscottiii:
Discussed in today's meeting. This RFC is good to go.

PR #221, opened 2025-04-07
wptrunner to print in economy mode by default

mstensho:
wptrunner pull request:
https://github.com/web-platform-tests/wpt/pull/51825


rfcs/print_economy_mode_by_default.md
```diff
@@ -0,0 +1,41 @@
...
+do by default.
+
+There's a CSS property `print-color-adjust` specified at
+https://drafts.csswg.org/css-color-adjust-1/#print-color-adjust , whose initial
+value is `economy`, which means to print without backgrounds. Setting it to
+`exact` will include backgrounds. If the browser is set up to always print
+backgrounds, this property will have no effect, since the spec says that the
+user preference must be respected more strongly than the hint provided by
+`print-color-adjust`. And now we're getting to the point: Without this change,
+it's impossible to test this property in web platform tests.
```
gsnedders, line 24: If the motivation is "it should be possible to test this property in web-platform-tests", then that presumably means that WPT cannot provide something as the user preference level unconditionally, given the spec says a user preference overrides it.

If the current WPT behaviour — which doesn't appear to be documented anywhere, c.f. https://github.com/web-platform-tests/wpt/blob/a09cf15dff82d741996db438e5804f1c04f46d3b/docs/writing-tests/print-reftests.md — is that, then an alternative fix would be to just define the behaviour as "WPT inserts an `html:style` element as the first child of the root element, whose text content is `:root { print-color-adjust: exact }`", because this can then trivially be overridden.

.gitignore
```diff
@@ -0,0 +1 @@
+*~
```
gsnedders, line 1: Let's not add text editor specific ignores, because there's a near endless quantity of text editors. You should just set a `gitignore` file, pointed to by `core.excludesFile` in your global/system gitconfig, containing things for whatever text editor you so desire, rather than relying on every git repository you touch having such ignores set.

gsnedders (commented):
While I haven't spoken to other WebKit contributors, this change makes me uneasy because it makes all print tests relying on a should-level carte-blanche, rather than clearly defined behaviour.

rfcs/print_economy_mode_by_default.md
mstensho: Yes, I want us to be able to write reftests for `print-color-adjust:economy`. Correct, this behavior is currently undefined for WPT. All the browsers that I know of print without backgrounds in economy mode (and, if needed, adjust text color to retain some contrast) - i.e. they all follow the informative example in the spec. Browsers have a checkbox in their print preview UI to enable backgrounds, and thus effectively disable support for `print-color-adjust:economy`. Currently, WPT behaves as if this box is always checked.

So it's impossible to add a reftest like this:
https://chromium-review.googlesource.com/c/chromium/src/+/6408671/5/third_party/blink/web_tests/external/wpt/css/css-color-adjust/print-color-adjust-001-print.html

So as long as the spec says do whatever you want to save ink for `print-color-adjust:economy`, I guess the feature is untestable. Make it normative?

jcscottiii:
Hi @mstensho,

FYI: This RFC was discussed during the recent infra [meeting](https://docs.google.com/document/d/1N9wdlGCOIlBo7_Dr7dcsgO4GIFPC8dnlb8qo_MaF97s/edit?tab=t.r6midqqkjacj).

To summarize from the notes:

```
There was concern from @gsnedders and @jgraham  that this was implementing unspecified behavior. And this should not happen by default. These types of tests should be optional/tentative. There might be changes required to implement that and that would require a RFC. But that RFC would be more likely to pass.
```

mstensho:
I will stop pursuing this for WPT for now. I'll add internal non-WPT Chromium tests for it instead, as I see no other option, as long as the text in the spec isn't normative.

PR #224, opened 2025-05-14
RFC for removing WebDriver implicit session creation

gsnedders:


rfcs/webdriver-implicit-session.md
```diff
@@ -0,0 +1,43 @@
+# RFC XXX: Remove implicit session creation for webdriver client
```
jgraham, line 1:
```suggestion
# RFC 224: Remove implicit session creation for webdriver client
```

jgraham (commented):
I think this should be fine; I'd probably want to run the patch against Gecko CI before landing, just in case any of our internal tests are doing anything unexpected.

gsnedders:
This RFC is now three weeks old, and it was briefly discussed in the meeting with rough support from @jgraham (see also above) and @jcscottiii; is there anything blocking this from getting approved?

jgraham (approved):
This seemed to be fine in our CI, so no objections here.

PR #80, opened 2021-04-30
RFC 80: Rename master branch to main in all repositories

foolip:
[Rendered](https://github.com/web-platform-tests/rfcs/blob/main_branch/rfcs/main_branch.md)

PR #225, opened 2025-05-15
Rename `master` branch to `main` in non-test repositories

gsnedders:
[Rendered](https://github.com/gsnedders/rfcs/blob/main_branch-side-repos/rfcs/main_branch_side_repos.md).

This is more limited in scope than #80, because it excludes `wpt` and `wpt-metadata`, which is where most of the risk/complexity comes from. This doesn't mean to imply we should not do #80, but we should take the easier wins where we can.

### DIFF
--- a/rfcs/main_branch_side_repos.md
+++ b/rfcs/main_branch_side_repos.md
@@ -1,0 +1,73 @@
+# RFC #XX: Rename `master` branch to `main` in non-test repositories
+
+## Summary
+
+Rename `master` branch to `main` in 
+[web-platform-tests/data-migration](https://github.com/web-platform-tests/data-migration),
+[web-platform-tests/wpt.live](https://github.com/web-platform-tests/wpt.live),
+[web-platform-tests/rfcs](https://github.com/web-platform-tests/rfcs),
+[web-platform-tests/editor](https://github.com/web-platform-tests/editor),
+[web-platform-tests/code-of-conduct-moderations](https://github.com/web-platform-tests/code-of-conduct-moderations), and
+[web-platform-tests/wpt-notes](https://github.com/web-platform-tests/wpt-notes).
+
+
+## Details
+
+Find non-archived repositories whose default branch is not main with:
+
+```sh
+gh api /orgs/web-platform-tests/repos | jq -r '.[] | select(.default_branch != "main" and (.archived | not)) | .html_url + ": " + .default_branch'
+```
+
+Yields:
+
+```
+https://github.com/web-platform-tests/wpt: master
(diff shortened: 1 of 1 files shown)


Ms2ger (approved):
sgtm

jgraham (approved):
Although `wpt.live` needs approval from whoever owns the deployment of that, since it would probably be broken by this change.

foolip (approved):
Yay, I look forward to `main` in all repos I work with :)

@jcscottiii I tried to find where wpt.live is deployed but couldn't. Do you know where it is, and is "master" hardcoded anywhere in that setup?

jcscottiii (approved):
LGTM from the wpt.live perspective. Deployments for the wpt.live infra are performed manually. As mentioned in the RFC, there are three references to `master` in the repo. -

- `src/supervisord.conf`: Tells the service to update it's local copy of the  `master` branch for [wpt](https://github.com/web-platform-tests/wpt.live/blob/e5833546d4365c847bb75312a159209a8d7cd84c/wpt-server-tot.Dockerfile#L64). But since wpt's default branch is not updated, **no change needed here**
- `infrastructure/web-platform-tests/compute.tf`: While it is a only a comment, we should point the comment to GCP official [docs](https://cloud.google.com/load-balancing/docs/https#firewall-rules)
- `.github/workflows/lint.yml`: We will need to change this from master to main

Looking forward to this change. I can make the above changes once this RFC is merged.

I had a good laugh at the Risk section of the RFC.

gsnedders:
> I had a good laugh at the Risk section of the RFC.

Literally a copy/paste of @foolip's #80's risk section. 🙃

> I can make the above changes once this RFC is merged.

FWIW, I was planning on doing this, changing `src/supervisord.conf` to fetch `HEAD` from upstream `wpt` (which eliminates one more thing we need to change when `wpt` finally renames its default branch); and changing `.github/workflows/lint.yml` to run on both `master` and `main`.

jgraham:
I have now done this for all the listed repos other than `wpt.live`.

gsnedders:
(For those unaware, `git remote set-head origin --auto` will reset the `origin/HEAD` symbolic link based on upstream.)

gsnedders:
Current status of `gh api /orgs/web-platform-tests/repos | jq -r '.[] | select(.default_branch != "main" and (.archived | not)) | .html_url + ": " + .default_branch'`:

https://github.com/web-platform-tests/wpt: master
https://github.com/web-platform-tests/wpt.live: master
https://github.com/web-platform-tests/wpt-metadata: master

The first and last are #80, and I've filed https://github.com/web-platform-tests/wpt.live/issues/85 for the middle one.

PR #226, opened 2025-06-03
Tentative testdriver methods

jcsteh:
[Rendered](https://github.com/jcsteh/wpt-rfcs/blob/tentative/rfcs/tentative_testdriver_methods.md).
CC @zcorpan, @cookiecrook.

### DIFF
--- a/rfcs/tentative_testdriver_methods.md
+++ b/rfcs/tentative_testdriver_methods.md
@@ -13,6 +13,7 @@ To answer these questions, it would be helpful if browser vendors could collabor
 ## Details
 1. There must be publicly documented (e.g. as part of an Interop Investigation Area) consensus between at least two browser vendors and intent to work towards a specification.
 2. Any tests directly or indirectly calling tentative testdriver methods must be [marked tentative](https://web-platform-tests.org/writing-tests/file-names.html#:~:text=.tentative,-:%20).
+3. These testdriver methods should assert that the test is tentative (URL path includes `.tentative' or '/tentative/'). 
 
 ## Alternatives considered
 1. Avoid any changes to the WPT repository until the specification is finalised.


zcorpan (approved):

rfcs/tentative_testdriver_methods.md
```diff
@@ -0,0 +1,29 @@
+# RFC 226: Tentative testdriver methods
+
+## Summary
+Allow tentative methods to be added to testdriver.js that are not yet included in the WebDriver or WebDriver BiDi specifications.
```
gsnedders, line 4: Let's be more explicit; we already have plenty of non-tentative things in testdriver which are not in either spec — #127 ([final RFC](https://github.com/web-platform-tests/rfcs/blob/main/rfcs/rfc_exemption_testdriver_method.md)) merely requires a WebDriver endpoint, not that it is in one of those specs.

rfcs/tentative_testdriver_methods.md
```diff
@@ -0,0 +1,29 @@
...
+This should only be allowed where there is consensus between multiple browser vendors and where there is clear progress towards a specification.
+
+## Background
+As part of the [Interop 2025 Accessibility Investigation Area](https://github.com/web-platform-tests/interop-accessibility/issues/148), the intention is to extend browsers and WPT to allow additional accessibility properties to be tested by web platform tests.
+However, there are open questions regarding the shape of the API for exposing these properties, what properties should be exposed, how the tests should be written, etc.
+While the end goal is to extend the WebDriver specification with the required new endpoints, these open questions need to be answered before this is feasible.
+To answer these questions, it would be helpful if browser vendors could collaborate on a tentative but working implementation.
```
gsnedders, line 11: For things like the Accessibility Investigation, is there any intention for there to be any, even high-level, explainer of what the proposed endpoints are? I'm mostly concerned about the risk of ending up somewhere where it is hard for anyone else to start passing the tests written using the tentative endpoint.

rfcs/tentative_testdriver_methods.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Details
+1. There must be publicly documented (e.g. as part of an Interop Investigation Area) consensus between at least two browser vendors and intent to work towards a specification.
```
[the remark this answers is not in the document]
gsnedders: I'd argue that tentative testdriver API pose a greater risk, because removing tests that rely on undocumented API is painful and hard — see https://github.com/web-platform-tests/rfcs/issues/172.

rfcs/tentative_testdriver_methods.md
```diff
@@ -0,0 +1,29 @@
...
+2. Any tentative methods added to testdriver.js should be prefixed with `tentative_`.
+3. Any utility methods which call tentative testdriver methods should also be prefixed with `tentative_`.
```
gsnedders, line 16: I wonder if we want there to be a `tentative` object, or whether we just want to prefix everything like this?

rfcs/tentative_testdriver_methods.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Alternatives considered
+1. Avoid any changes to the WPT repository until the specification is finalised.
+  This makes it very difficult for browser vendors to collaborate and to learn about problems that are much easier to discover and understand "in practice".
+  In contrast, if the specification were developed without a working implementation, there is a much higher risk of fundamental design problems in the specification which are much harder to fix later.
+2. Develop the implementation as vendor specific tests with methods in testdriver-vendor.js.
+  While this does allow a single vendor to iterate on a working implementation, it is still very difficult for multiple browser vendors to collaborate on this.
+  At the very least, other vendors may wish to contribute tests which exercise the implementation to ensure it fits their needs before agreeing to a final specification.
+
+## Risks
```
gsnedders, line 27: Also:

* That the API sufficiently changes before it becomes finalised, that we have sufficiently many tests written using the tentative API, that is becomes a significant effort to migrate them to the final API.

* That other vendors are interested in ensuring there is interoperability with the tested feature but have to reverse engineer what the tentative testdriver API is.

gsnedders (dismissed):
Most of my comments below are really "let's make this more explicit", rather than any real objection.

And mostly a question for the Core Team, but I do wonder what standard we thing we're holding things to based on #127: is an explainer enough for a `testdriver` API, or do we want a spec? Because if an explainer is enough, I'd question how often we'd actually add `tentative` API, because we probably do want _something_ defining what the behaviour of the endpoint is, even if only at a high-level.

rfcs/tentative_testdriver_methods.md
jcsteh: I wonder if having a tentative object would make it trickier to "graduate" things from tentative once the spec is final (and my thinking is that we want to incentivise finalising tentative things as soon as is feasible). With prefixing, you can basically just search/replace once the spec is final. With an object, things will need to be moved around. That's obviously not a show stopper if an object is preferred, just something I thought worth considering.

Ms2ger (commented):
I have no fundamental concerns with this, assuming we'd use this for experimentation, not things that stick around indefinitely.

rfcs/tentative_testdriver_methods.md
```diff
@@ -0,0 +1,32 @@
...
+However, there are open questions regarding the shape of the API for exposing these properties, what properties should be exposed, how the tests should be written, etc.
+While the end goal is to extend the WebDriver specification with the required new endpoints, these open questions need to be answered before this is feasible.
+To answer these questions, it would be helpful if browser vendors could collaborate on a tentative but working implementation.
+
+## Details
+1. There must be publicly documented (e.g. as part of an Interop Investigation Area) consensus between at least two browser vendors and intent to work towards a specification.
+2. Any tentative methods added to testdriver.js should be prefixed with `tentative_`.
+3. Any utility methods which call tentative testdriver methods should also be prefixed with `tentative_`.
+4. Any tests directly or indirectly calling tentative testdriver methods must be marked tentative.
```
cookiecrook, line 17:
```suggestion
4. Any tests directly or indirectly calling tentative testdriver methods must be marked tentative by including `.tentative.` in the file name.
```

rfcs/tentative_testdriver_methods.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Alternatives considered
+1. Avoid any changes to the WPT repository until the specification is finalised.
+  This makes it very difficult for browser vendors to collaborate and to learn about problems that are much easier to discover and understand "in practice".
+  In contrast, if the specification were developed without a working implementation, there is a much higher risk of fundamental design problems in the specification which are much harder to fix later.
```
cookiecrook, line 22: Perhaps the initial cross-engine prototype of the property bag implementation should include only the properties already exposed by other WebDriver methods, like `computedrole` and `computedlabel`.

rfcs/tentative_testdriver_methods.md
gsnedders: For example, https://webbluetoothcg.github.io/web-bluetooth/#automated-testing defines a whole load of endpoints.

The policy is that it requires a spec for the endpoint, not that that spec be in any specific standard.

rfcs/tentative_testdriver_methods.md
gsnedders: I'd rather we just cross-reference https://web-platform-tests.org/writing-tests/file-names.html#:~:text=.tentative,-:%20 rather than state that specifically — we aren't trying to require that the tests have `.tentative.` in their name specifically, merely that they match our normal tentative semantics.

rfcs/tentative_testdriver_methods.md
jcsteh: This would make it easier to standardise on a first pass, for sure. However, that first pass would have no practical benefit, since we can already get those properties using the existing methods. The problem I see with this is that we're less likely to spot problems because the implementation is thus entirely "theoretical": we won't spot the kinds of problems we're likely to see in testing new properties, which is the entire point of this work. In contrast, the current prototype supports a few simple ARIA properties and parent/children, which allows us to test the kinds of things we really want to test with this work and thus spot problems we might hit as we move forward.

rfcs/tentative_testdriver_methods.md
jcsteh: Thanks. Now I better understand your original question regarding whether an explainer is sufficient. I agree that at least having an explainer makes sense, so if the consensus is that an explainer is acceptable, we probably don't need this RFC. [RFC 127](https://github.com/web-platform-tests/rfcs/blob/main/rfcs/rfc_exemption_testdriver_method.md) just says "an existing WebDriver endpoint", which isn't explicit about a formal spec, though I guess that could be inferred. So perhaps we still need an RFC making it clear that an explainer is sufficient.

That said, one difference between a spec and an explainer is that an explainer is far more likely to change than a spec. So that might still be one reason to mark these methods as tentative: to indicate that the details are still being ironed out.

zcorpan:
@gsnedders Can you take another look at this?

rfcs/tentative_testdriver_methods.md
jcsteh: One open question here is whether we need this RFC at all or whether an explainer is sufficient. See https://github.com/web-platform-tests/rfcs/pull/226#discussion_r2224692810. If an explainer is sufficient, is it sufficient for that explainer to be a GitHub issue (as is the case now) or does it need to be a document (to make it easier to follow)?

rfcs/tentative_testdriver_methods.md
jcsteh: This should probably be discussed elsewhere, but now that we've ascertained that we don't need to worry about the WebElement-style abstraction for accessibility ids, I think the way in which accessibility ids should be exposed is easy to define: a simple unique string such as a UUID. DOM nodes should be exposed as WebElements just as they are everywhere else. It's just a matter of implementing it. There might be tree structure differences, but that's not relevant to WebDriver exposure; that's a higher level question of which parts of the tree we can standardise (and working around that in tests if need be).

zcorpan:
An alternative to having `tentative_` in the method names is to have an assertion for these methods that checks that the test itself is tentative (by checking the test's URL).

rfcs/tentative_testdriver_methods.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Background
+As part of the [Interop 2025 Accessibility Investigation Area](https://github.com/web-platform-tests/interop-accessibility/issues/148), the intention is to extend browsers and WPT to allow additional accessibility properties to be tested by web platform tests.
+However, there are open questions regarding the shape of the API for exposing these properties, what properties should be exposed, how the tests should be written, etc.
+While the end goal is to extend the WebDriver specification with the required new endpoints, these open questions need to be answered before this is feasible.
+To answer these questions, it would be helpful if browser vendors could collaborate on a tentative but working implementation.
+
+## Details
+1. There must be publicly documented (e.g. as part of an Interop Investigation Area) consensus between at least two browser vendors and intent to work towards a specification.
+2. Any tests directly or indirectly calling tentative testdriver methods must be [marked tentative](https://web-platform-tests.org/writing-tests/file-names.html#:~:text=.tentative,-:%20).
```
cookiecrook: Need to add number 3 here I think, so outline the expectation of an assertion in any tentative methods.

zcorpan:
```suggestion
2. Any tests directly or indirectly calling tentative testdriver methods must be [marked tentative](https://web-platform-tests.org/writing-tests/file-names.html#:~:text=.tentative,-:%20).
3. These testdriver methods should assert that the test is tentative (URL path includes `.tentative' or '/tentative/'). 
```

cookiecrook (approved):
r+ with the addition of the new assertion expectation in tentative methods.

cookiecrook:
> Most of my comments below are really "let's make this more explicit", rather than any real objection.

I think it's okay to dismiss @gsnedders review since it's 5 months old, was self-described as not a "real objection", and I believe has fully addressed in the latest iteration.

PR #232, opened 2026-01-06
Update minimum supported Python to 3.9.

### LINKED ISSUE #200: Update minimum supported Python to 3.9
As of now the minimum supported version of Python for web-platform tests is Python 3.8. Given that this release is supported until October 2024 we should start planning to get rid of it and use at least version 3.9 which gets security updates until October 2025.

Right now such a bump would be blocked at least from Mozilla side because the wpt tests are still running on workers which have Python 3.8 installed. But over the last months newer images have been created with Python 3.10 installed and [those are getting slowly rolled out](https://bugzilla.mozilla.org/show_bug.cgi?id=wayland-ci). There is specifically [bug 1865353](https://bugzilla.mozilla.org/show_bug.cgi?id=1865353) for web-platform tests.

I'm filing this as issue for now to get feedback from other vendors regarding their requirements and possible blockers. I assume that we want to keep support for all available Python versions or should we even consider bumping the minimum version to 3.10? Feedback is welcome.

CC @jgraham, @foolip, @jcscottiii, @gsnedders, @Ms2ger

whimboo:
Fixes #200.

@jgraham, @gsnedders and @WeizhongX please review. Thanks

### DIFF
--- a/rfcs/python_39.md
+++ b/rfcs/python_39.md
@@ -12,7 +12,7 @@ support for it. As a result, we have been unable to update several dependencies
 that now require Python 3.9 or newer.
 
 Historically, continued Python 3.8 support was required due to downstream CI
-integrations with Apple’s and Google’s infrastructure. These systems have since
+integrations with vendor infrastructure. These systems have since
 been updated and no longer depend on Python 3.8, removing the primary blocker
 for increasing our minimum supported version.
 


rfcs/python_39.md
```diff
@@ -0,0 +1,43 @@
...
+Historically, continued Python 3.8 support was required due to downstream CI
+integrations with Apple’s and Google’s infrastructure. These systems have since
+been updated and no longer depend on Python 3.8, removing the primary blocker
+for increasing our minimum supported version.
+
+Implementing this change will require several concrete steps:
+
+- Update CI jobs currently running on Python 3.8 to use Python 3.9.
+- Refresh vendored dependencies that were previously constrained by
+  Python 3.8 compatibility.
```
gsnedders: Note with https://github.com/web-platform-tests/wpt/pull/49752 unlanded we don't have a particularly easy way to do this. :'(

rfcs/python_39.md
```diff
@@ -0,0 +1,43 @@
...
+- Re-run CI for Dependabot pull requests to allow dependency updates to proceed.
+  Some dependencies may now require Python 3.10 or newer; in those cases, we
+  should manually update to the latest release that remains compatible with
+  Python 3.9.
```
gsnedders: I did question in https://github.com/web-platform-tests/wpt/issues/55608 whether we should just outright disable dependabot; it's unclear to me we're gaining much except noise with our minimum version?

gsnedders (commented):
This clearly needs review from someone at Chromium more than anyone else, but:

gsnedders:
> @gsnedders mind checking again now that Google approved this RFC? Thanks.

I'd just like to not claim that Apple required Python 3.8 support for a while, because we stopped needing support in early 2023. I believe it was also back in 2023 when other WebKit ports (WebKitGTK notably) stopped requiring Python 3.8 support.

By contrast, #200 has both Mozilla and Google needing support in 2024.

PR #233, opened 2026-01-08
safe-printable-inset META value proposal

mstensho:
Spec: https://drafts.csswg.org/css-page-3/#page-margin-safety

Spec discussion: https://github.com/w3c/csswg-drafts/issues/11395

### DIFF
--- a/rfcs/safe-printable-inset.md
+++ b/rfcs/safe-printable-inset.md
@@ -18,7 +18,7 @@ areas.
 One rather straight-forward solution would be a META value that sets the width
 of the unprintable area on all four sides. For instance:
 
-`<meta name="safe-printable-inset" content="[inset-specifier]"`
+`<meta name="safe-printable-inset" content="[inset-specifier]">`
 
 where `inset-specifier` is a numeric value. The unit could be CSS pixels or
 points. Using centimeters for anything here isn't a great idea, since they don't
@@ -31,4 +31,4 @@ oversimplification of reality, printers may rotate the print output at their own
 discretion. The user agent may therefore not be able to make assumptions about
 which edge (long or short?) will be fed first into the printer, or what
 orientation the sheet of paper has. Therefore using just one value (which should
-represent the largest of the four) seems reasonable.
+represent the larger of the four) seems reasonable.
--- a/rfcs/safe-printable-inset.md
+++ b/rfcs/safe-printable-inset.md
@@ -1,0 +1,34 @@
+# RFC #xyz: Simulate safe printable inset
+
+## Summary
+
+We need a way of testing the [safe-printable-inset](https://github.com/w3c/csswg-drafts/pull/13190/files) property.
+
+## Details
+
+Most printers have a small region along each edge of the paper edges that's not
+reliably printable, usually due to the printer's paper handling mechanism.
+Authors can steer clear of such unprintable areas using the
+`safe-printable-inset` property, which applies in `@page` and `@page` margin
+contexts.
+
+There should be a way for print reftests to test this, by simulating unprintable
+areas.
+
+One rather straight-forward solution would be a META value that sets the width
+of the unprintable area on all four sides. For instance:
+
+`<meta name="safe-printable-inset" content="[inset-specifier]"`
+
+where `inset-specifier` is a numeric value. The unit could be CSS pixels or
+points. Using centimeters for anything here isn't a great idea, since they don't
+convert nicely into CSS pixels (unlike inches).
+
+Why just one value for all four edges? Although many printers indeed don't
+necessarily have a uniform unprintable area width along each of the four paper
+edges (although many do), so that just providing one value for all is an
+oversimplification of reality, printers may rotate the print output at their own
+discretion. The user agent may therefore not be able to make assumptions about
+which edge (long or short?) will be fed first into the printer, or what
+orientation the sheet of paper has. Therefore using just one value (which should
+represent the largest of the four) seems reasonable.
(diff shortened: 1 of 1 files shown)


rfcs/safe-printable-inset.md
```diff
@@ -0,0 +1,34 @@
...
+areas.
+
+One rather straight-forward solution would be a META value that sets the width
+of the unprintable area on all four sides. For instance:
+
+`<meta name="safe-printable-inset" content="[inset-specifier]"`
+
+where `inset-specifier` is a numeric value. The unit could be CSS pixels or
+points. Using centimeters for anything here isn't a great idea, since they don't
+convert nicely into CSS pixels (unlike inches). I suggest using CSS pixels.
```
jcscottiii: Question: Is it worth it to enforce the usage of pixels here? (Or rather what's the risk if we don't enforce it?)

When new metadata is introduced, there's a risk of incorrect usage (e.g., typos, invalid values, or inappropriate units). For example, a value like `content="2in"` might be provided, even though the RFC suggests CSS pixels or points and discourages inches. The best way to mitigate this is to add automated checks. Other places that handled this:
- **RFC 163 (web_features)** added a lint to ensure its new metadata didn't get out of sync with test files
- **RFC 158 (Variant name should be a non-zero length string)** added a lint to forbid a problematic authoring pattern
- **RFC 211 (Support testdriver.js in other test types)** proposed a lint to enforce correct usage with `reftest-wait`.

jcscottiii (approved):
LGTM. I only have one question.

gsnedders (commented):
Mostly my question is "how would we hook this up to the style engine", because somehow we have to go from the meta element to this impacting print style and output.

Most of the time we're using WebDriver to communicate with the browser; does it therefore make sense to add something to the WebDriver print command that specifies the safe printable inset? My gut is yes?

gsnedders (changes requested):
> @gsnedders Are you asking whether we should add something to https://w3c.github.io/webdriver/#print-page ?

Yes, which appears to be what the draft CL does. Can you please make a PR against WebDriver, and get consensus there first?

> I think that might be a good idea. It could be specified as a single inset value, or it could be a rectangle within `pageWidth` x `pageHeight`.

It should allow both separately, I'd expect.
 
> Is everything in centimeters here? That's rather bad, since they don't convert nicely to CSS pixels. But I suppose we need to follow suit.

Yes, which is… deeply unfortunate, I agree. Though to be fair, A4 paper doesn't convert nicely to CSS pixels either. :)

mstensho:
The css-page spec changes have landed, and I've updated the description here accordingly.

jcscottiii:
We talked about this today. @gsnedders is commenting on https://github.com/w3c/webdriver/pull/1950 and https://github.com/w3c/csswg-drafts/issues/11395

rfcs/safe-printable-inset.md
```diff
@@ -0,0 +1,35 @@
...
+since everything else is already in centimeters, better follow suit.
+
+Why just one value for all four edges? Although many printers indeed don't
+necessarily have a uniform unprintable area width along each of the four paper
+edges (although many do), so that just providing one value for all is an
+oversimplification of reality, printers may rotate the print output at their own
+discretion. The user agent may therefore not be able to make assumptions about
+which edge (long or short?) will be fed first into the printer, or what
+orientation the sheet of paper has. Therefore using just one value (which should
+represent the larger of the four) seems reasonable.
```
gsnedders, line 35: This does leave us unable to test all the different at-rules the descriptor applies to — `@page`, `@top-left-corner`, `@top-left`, `@top-center`, `@top-right`, `@top-right-corner`, `@right-top`, `@right-middle`, `@right-bottom`, `@bottom-right-corner`, `@bottom-right`, `@bottom-center`, `@bottom-left`, `@bottom-left-corner`, `@left-bottom`, `@left-middle`, `@left-top`.

The spec says, along these lines:

> Some printers don’t have a uniform unprintable area width along each of the four paper edges, and the printer may rotate the print output at their own discretion. The user agent may therefore not be able to make assumptions about which edge will be fed first into the printer, or what orientation the sheet of paper has. If the user agent cannot make such assumptions, only one will be provided (to be used on all 4 sides): The larger of these four values. Otherwise, if the user agent can trust that the four values are usable individually, and that no rotation is going to take place, and so on, each side of the page may have individual values.

It's not entirely clear to me from the spec how you go from potentially seventeen values (if specified in `@page` and all the margin at-rules) to only one, because the "larger of these four values" makes an assumption that you've already gone from seventeen to four.

Maybe starting with taking a single value is fine — but I could easily see us wanting to extend the metadata to take four values (potentially with semantics similar to the `margin` property?) to be able to actually test the case where the user agent *can* make assumptions about what edge will be fed first.

mstensho: We're not unable to test any of those at-rules. The safe printable inset will just be the same along each of the four edges.

`@page` is adjacent to all four edges of a paper sheet, so all of `margin-top`, `margin-right`, `margin-bottom` and `margin-left` may be affected by page margin safety. `@blah-corner` is adjacent to two of the edges. The remaining at-rules are adjacent to only one edge. For e.g. `@top-center`, only `margin-top` may be affected by page margin safety. For `@bottom-right-corner`, `margin-right` and `margin-bottom` may be affected.

Here's a test (not yet landed): https://chromium-review.googlesource.com/c/chromium/src/+/7415846/4/third_party/blink/web_tests/external/wpt/css/css-page/safe-printable-inset-003-print.html

Ms2ger:
Please add some details on how this would work into the RFC document - I see references to webdriver in the comments, but not in the RFC itself

mstensho:
> Please add some details on how this would work into the RFC document - I see references to webdriver in the comments, but not in the RFC itself

What can I add at this time, given that https://github.com/w3c/webdriver/pull/1950 hasn't landed yet?

jcscottiii:
@mstensho We discussed this. Still waiting on the web driver PR to land before we land this.

PR #235, opened 2026-02-19
RFC 234: support 5 WebTransport h3 servers with different certs

jesup:


gsnedders (changes requested):
My biggest questions here are about how this interacts with the WebDriver Classic/BiDi `acceptInsecureCerts`.

Currently, looking at https://github.com/search?q=repo%3Aweb-platform-tests%2Fwpt+acceptInsecureCerts+path%3A%2F%5Etools%5C%2F%2F&type=code, we use `acceptInsecureCerts` for:

 * Firefox does when there's no `certutil` binary on the path (which if you're using a published release build depends on if you've otherwise installed it; for the sake of our Docker images, Ubuntu has this in a separate [libnss3-tools](https://packages.ubuntu.com/resolute/libnss3-tools) package).
 * Safari does unconditionally.
 * Chrome uses its `--ignore-certificate-errors-spki-list` argument instead.

Realistically, we're going to need to keep the `acceptInsecureCerts` case working, and have it work for WebTransport, which makes it hard to unconditionally certificate validity, which AIUI is part of the goal here.

To me, this suggests we need some flag (like we have for h2) for tests which require certificate validation.

PR #237, opened 2026-03-26
RFC 237: Exclusions in WEB_FEATURES.yml files

jugglinmike:
[Rendered](https://github.com/bocoup/wpt-rfcs/blob/web-features-exclusions/rfcs/web_features_exclusions.md)

### DIFF
--- a/rfcs/web_features_exclusions.md
+++ b/rfcs/web_features_exclusions.md
@@ -1,4 +1,4 @@
-# RFC 226: Exclusions in WEB_FEATURES.yml files
+# RFC 237: Exclusions in WEB_FEATURES.yml files
 
 ## Summary
 
--- a/rfcs/web_features_exclusions.md
+++ b/rfcs/web_features_exclusions.md
@@ -197,7 +197,7 @@ features:
 - name: alerts
   patterns:
   - files:*
-  - !feature:print
+  - "!feature:print"
 - name: print
   patterns:
   - files:print-*
@@ -247,7 +247,7 @@ features:
   - "!print"
 - name: print
   files:
-  - ./print-*
+  - print-*
 ```
 
 While this disambiguation heuristic would be trivial to implement in software,
--- a/rfcs/web_features_exclusions.md
+++ b/rfcs/web_features_exclusions.md
@@ -1,0 +1,306 @@
+# RFC 226: Exclusions in WEB_FEATURES.yml files
+
+## Summary
+
+Extend the schema of the `WEB_FEATURES.yml` files to support excluding one set
+of tests from another.
+
+## Background
+
+[Contributors classified a large number of web-features over the final months
+of
+2025](https://github.com/search?q=repo%3Aweb-platform-tests%2Fwpt++is%3Apr+author%3Achrisc+author%3Ajugglinmike+author%3Astalgiag+author%3Ahoward-e++author%3Agnarf+created%3A%3E2025-09-01+&type=pullrequests),
+and their experience surfaced many opportunities for improving the
+maintainability of WPT's `WEB_FEATURES.yml` files[^1]. One of them is a
+shorthand for excluding one set of tests from another set.
+
+The problem starts with the application of filename-matching patterns. On its
+face, this is a helpful feature. It helps maintainers avoid some of the
+brittleness of the out-of-band classification scheme. However, in many cases,
+there are subsets of tests which are better-classified to distinct
+web-features. It's currently possible to exclude tests by filename by prefixing
+a pattern with an exclamation mark (`!`), similar to [the convention in Git's
+`.gitignore` files](https://git-scm.com/docs/gitignore). In the example below
+(inspired by [the current form of
+`html/webappapis/user-prompts/WEB_FEATURES.yml`](https://github.com/web-platform-tests/wpt/blob/e08956ee1ae70503cfbbc160716081768c9630cd/html/webappapis/user-prompts/WEB_FEATURES.yml)),
+the entry for the `alerts` web-feature includes all files except those
+belonging to the `print` web-feature:
+
+```yaml
+features:
+- name: alerts
+  files:
+  - "*"
+  - "!print-*"
+- name: print
+  files:
+  - print-*
+```
+
+Maintaining this separation requires duplicating membership information in two
+places. This duplication is challenging to maintain, and it obscures the
+semantic relationship between the sets. (For larger excluded sets, the
+challenge is more pronounced--the example used in this document was selected
+for brevity, but there exist many files with far more duplication, e.g.
+[`css/CSS2/generated-content/WEB_FEATURES.yml`](https://github.com/web-platform-tests/wpt/blob/e08956ee1ae70503cfbbc160716081768c9630cd/css/CSS2/generated-content/WEB_FEATURES.yml).)
+
+Although in-line comments can alert contributors to the intention and implore
+them to maintain it, a structural fix would alleviate everyone of this chore.
+
+## Details
+
+We propose allowing any entry for one web-feature to reference another from the
+`files` list by prefixing the web-feature ID with the number sign (`#`). The
+matched files can be excluded from the referencing set by further prefixing
+with the exclamation point (`!`), just as with filename patterns. Filename
+patterns should be distinguished with a leading sequence of `./`.
+
+The example above, altered according to this proposed change, would appear as
+follows:
+
+```yaml
+features:
+- name: alerts
+  files:
+  - ./*
+  - "!#print"
+- name: print
+  files:
+  - ./print-*
+```
+
+## Alternatives Considered
+
+### Changing the name of the `files` key
+
+The patterns will no longer directly describe files, so "files" may not be the
+more accurate name. Names like `patterns` or `tests` may be more appropriate.
+
+```yaml
+features:
+- name: alerts
+  patterns:
+  - ./*
+  - "!#print"
+- name: print
+  patterns:
+  - ./print-*
+```
+
+However, file-matching continues to be the patterns' purpose. This might be
+worth re-visiting when/if we decide to allow for matching tests as distinct
+from files (e.g. matching specific expansions of [`.any.js`
+tests](https://web-platform-tests.org/writing-tests/testharness.html#tests-for-other-or-multiple-globals-any-js)).
+
+### Exclusion as the default (and only) behavior
+
+Currently, we only have need to exclude sets. The `!` prefix, borrowed from
+existing file-pattern syntax, insinuates non-existent "inclusion" use-case.
+Omitting it would avoid any such insinuation and reduce the number of
+characters necessary to support the desired use-case.
+
+```yaml
+features:
+- name: alerts
+  files:
+  - ./*
+  - "#print"
+- name: print
+  files:
+  - ./print-*
+```
+
+That said, it seems somewhat confusing for the semantics of the two types of
+patterns to vary so widely. (This could be mitigated by placing exclusions
+under a separate key since. A name like "exclusions" would make it much easier
+to recognize the "exclude by default" semantic of these patterns. However,
+introducing a separate key by any name has its own drawbacks--see the next
+section.) There's also reason to believe that future extensions will make use
+of feature *inclusion* (namely: the ability to explicitly express intentional
+overlaps).
+
+### Expressing inclusions under a dedicated key
+
+Introducing a new key like `features` (and placing all feature exclusion rules
+there) would negate the need for "pattern type" prefixes altogether and obviate
+all the design considerations which follow this one. Since we expect most
+patterns to describe files directly, any prefix for those entries could be
+considered noisy. (They will also dramatically increase the size of the patch
+which enacts this RFC.)
+
+For example:
+
+```yaml
+features:
+- name: alerts
+  files:
+  - "*"
+  features:
+  - "!print"
+- name: print
+  files:
+  - print-*
+```
+
+And for completeness, here's how that could look with the "exclude by default"
+semantic contemplated above:
+
+```yaml
+features:
+- name: alerts
+  files:
+  - "*"
+  exclude:
+  - print
+- name: print
+  files:
+  - print-*
+```
+
+However, splitting patterns across two keys, whatever their name, has
+concerning implications for pattern precedence. The sequence for applying two
+distinct sets of patterns isn't immediately obvious (especially considering
+that this feature may one day be used to express intentional overlaps between
+sets). Whatever design we might select, the need for a dedicated design effort
+suggests that the behavior may not be intuitive for contributors.
+
+A more complex schema could have it both ways: a "list of maps of lists" would
+reduce the repetition of a "files" designation while preserving the author's
+control over the order of operations. For example, the rules expressed in the
+following sketch have a much more clear sequence:
+
+```yaml
+features:
+- name: alerts
+  patterns:
+  - files:
+    - "*"
+  - features:
+    - "!print"
+- name: print
+  patterns:
+  - files:
+    - print-*
+```
+
+While clear precedence is important (especially for future extensions), the
+complexity of this structure seems preferable to the noise of pattern prefixes.
+
+### More explicit pattern prefixes
+
+Keyword prefixes like `files:` and `feature:` would be more self-documenting
+(and would likely justify the key-renaming change explored above):
+
+```yaml
+features:
+- name: alerts
+  patterns:
+  - files:*
+  - !feature:print
+- name: print
+  patterns:
+  - files:print-*
+```
+
+The shorter prefixes (`./` and `#`) are inspired by conventions that many
+developers will recognize (filesystem paths and HTML IDs, respectively), so
+they strike a balance between verbosity and understandability.
+
+### Omitting the `files:` prefix
+
+The prefix is not technically necessary because tooling could assume entries
+are filename patterns if they lack an explicit "feature" prefix (whatever that
+prefix may be).
+
+```yaml
+features:
+- name: alerts
+  files:
+  - "*"
+  - "!#print"
+- name: print
+  files:
+  - print-*
+```
+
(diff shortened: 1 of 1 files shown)


jugglinmike:
Hi @jcscottiii! Thanks for your feedback!

One novel aspect of your proposal is that it scopes exclusions to individual path patterns. While I think that could be tenable, it's not a capability that we've specifically felt a need for.

It sounds like we're aligned on prioritizing human authors/readers. To that end, I think "scoped" exclusions may make these rules more difficult to understand since they would effectively introduce a grouping operator that hasn't been motivated by experience. (Well, not our experience, anyway. I'd be happy to hear about any instances where you wanted it!)

How would you feel about expressing exclusions with a standalone object so that each list item could be either a string value or a dict with a single key (namely, `exclude_ids`)? For example:

```diff
 features:
   - name: alerts
     files:
-      - path: ./*
-        exclude_ids:       # Standard YAML list, zero ambiguity
+      - ./*
+      - exclude_ids:       # Standard YAML list, zero ambiguity
           - print
           - logging
   - name: print
     files:
       - ./print-*         # Simple file path can stay a simple string!
```

Anecdotally, I've only observed a small number (read: 1 to 3) of exclusions per feature entry, so a nested list like `exclude_ids` might be more structure than we truly need. I'm curious about simplifying it further to a string value for a key named `exclude_id` (despite the bit of repetition it adds to the running example):

```diff
 features:
   - name: alerts
     files:
-      - path: ./*
-        exclude_ids:       # Standard YAML list, zero ambiguity
-          - print
-          - logging
+      - ./*
+      - exclude_id: print   # Standard YAML string, zero ambiguity
+      - exclude_id: logging # Standard YAML string, zero ambiguity
   - name: print
     files:
       - ./print-*         # Simple file path can stay a simple string!
```

...but I don't feel this flattening would have a huge impact on ergonomics, so I could go either way!

In any case (in your proposal and in my suggested amendments), the `./` prefix is not technically necessary. Are you suggesting it should become mandatory?

jcscottiii:
@jgraham mentioned that this should be a set of rules to web-feature-ids. Something like this:

```
	file-1.html: css-flexbox
	file-2.html: css-grid
	*: css-multicol
```

I'll let @jgraham provide a more thorough review.

jugglinmike:
A bit more context for the benefit of tomorrow's RFC triage call:

Since this RFC is about streamlining the metadata, [here's an approximation of its impact](https://gist.github.com/jugglinmike/1bd28f6c6f6af252a930679fb4ce067c). While the syntax is still under consideration, that patch shows the order of magnitude: "130 files changed, 275 insertions(+), 791 deletions(-)"

Until recently, WPT didn't formally document the `WEB_FEATURES.yml` files. [We landed some documentation last week](https://github.com/web-platform-tests/wpt/pull/59170), so you can now find it online at https://web-platform-tests.org/writing-tests/out-of-band-metadata.html

Looking forward to discussing more tomorrow!

jgraham:
> @jgraham mentioned that this should be a set of rules to web-feature-ids

Yes. 

Mapping feature ids to filename patterns makes sense if you are trying to work out "which files are part of this feature". You find the id you care about and then evaluate the given rules against all the files in the directory. That is indeed a common use case for this data, but critically it's a use case that's basically always automated.

On the other hand if you have one file and want to know what features it will correspond to you have to evaluate the full set of rules. That's a use case you have when adding a file when you want to know if it will be correctly labelled, or if a rule update is required. Typically that use case isn't automated.

So if we want to optimise for the latter, we should look to make it as easy as possible to figure out what features correspond to a given file. In that case having the rules be a list of patterns that could apply and stopping on the first that does apply seems likely to be much easier to work with e.g.

```
*-grid-*: [css-grid, css-multicol]
*-flex-*: [css-flexbox, css-multicol]
*: css-multicol
```

There is arguably a bug there that a file which has both `-grid-` and `-flex` in it would only be labelled as `css-grid` and `css-multicol`. In theory that seems bad. In practice it seems likely to be fine to fix it by adding another rule like:
```
*-grid-flex-*: [css-grid, css-flexbox, css-multicol]
```
and just assuming that people can follow a naming convention rather than requiring a perfectly general syntax. In particular, I think that once you start making things additive you end up back at the starting point for this RFC which is "we need to invent an exclusion mechanism so that some rules don't apply in some cases" and then you're back at it being really hard for a human to correctly deduce the impact of the rules.

jugglinmike:
> we should look to make it as easy as possible to figure out what features correspond to a given file.

I believe that we can achieve that goal without any changes to the schema. Since all file-matching patterns are strictly ordered in the current design (via lists of lists), we can simply change the way the files are interpreted (namely by adding strict precedence). Just like in the flat structure that @jgraham sketched out above, this will obviate the need for the `!` prefix, reducing total entry count.

This might be worth considering because the flat structure would introduce more repetition in entries that hold a lot of patterns. The pathological case is [`./css/css-conditional/container-queries/WEB_FEATURES.yml`](https://github.com/web-platform-tests/wpt/blob/master/css/css-conditional/container-queries/WEB_FEATURES.yml) with its 128 patterns for container-queries. We shouldn't design for the pathological case, though, so I've collected some stats to give a better sense for the number of file patterns typically used (taking care to ignore the `!`-prefixed entries).

Mean: 2.3628988642509463  
Standard deviation: 5.748726896571267

```mermaid
xychart
    title "File patterns per web-feature entry"
    x-axis "# of web-features" [1, 2, 3, 4, 5, 6, 7, 8, 9+]
    y-axis "# of file patterns" 1 --> 869
    bar [869, 748, 78, 43, 27, 17, 12, 4, 51]
```

Grist for the mill, in any case!

<details>
  <summary>Source code</summary>

    #!/usr/bin/env python3
    
    import itertools
    import os
    import yaml
    
    MAX_BUCKET = 9
    
    def find(root):
        for dirpath, dirnames, filenames in os.walk(root):
            if 'WEB_FEATURES.yml' not in filenames:
                continue
    
            filename = os.path.join(dirpath, 'WEB_FEATURES.yml')
            with open(filename, 'r') as handle:
                yield (filename, yaml.safe_load(handle)['features'])
    
    def int_list(ints):
        return ', '.join(map(str, ints))
    
    def render(mean, standard_deviation, grouped):
        x_axis = int_list(range(1, MAX_BUCKET)) + f', {MAX_BUCKET}+'
        max_value = max(grouped.values())
        values = int_list(grouped.values())
        return f'''
    Mean: {mean}  
    Standard deviation: {standard_deviation}
    
    ```mermaid
    xychart
        title "File patterns per web-feature entry"
        x-axis "# of web-features" [{x_axis}]
        y-axis "# of file patterns" 1 --> {max_value}
        bar [{values}]
    ```
        '''
    
    def main(root):
        file_entry_counts = []
        grouped = {x: 0 for x in range(1, MAX_BUCKET + 1)}
        for filename, features in find(root):
            for feature in features:
                # Negation entries will be obviated by the new semantics
                count = len([x for x in feature['files'] if not x.startswith('!')])
                grouped[min(MAX_BUCKET, count)] += 1
    
                file_entry_counts.append(count)
    
        size = len(file_entry_counts)
        mean = sum(file_entry_counts) / size
        variance = sum([(count - mean) ** 2 for count in file_entry_counts]) / size
        standard_deviation = variance ** 0.5
        return render(mean, standard_deviation, grouped)
    
    if __name__ == '__main__':
        print(main('.'))

</details>

jugglinmike:
@jcscottiii @jgraham @gsnedders I've updated the proposal to reflect the consensus of the latest RFCs & Infrastructure meeting.

rfcs/web_features_exclusions.md
```diff
@@ -0,0 +1,192 @@
...
+
+In the example above, to include the file named `bar.html` in both the `print`
+web-feature *and* the `alerts` web-feature, one would write:
+
+```yaml
+features:
+- foo.html: NULL
+- bar.html: [print, alerts]
+- print-*: print
+- "*": alerts
```
jcscottiii: Instead of the value being of type `string | array[string] | NULL`, couldn't we just simplify it to be array[string]?

```suggestion
```yaml
features:
- foo.html: []
- bar.html: ["print", "alerts"]
- print-*: ["print"]
- "*": ["alerts"]
```

rfcs/web_features_exclusions.md
```diff
@@ -0,0 +1,192 @@
...
+This proposal's strength comes from making file-matching rules mutually
+exclusive by default. Because these rules have a deterministic order under the
+existing schema, the main benefits could be enjoyed without any schema changes
+whatsoever.
+
+However, less structure is generally easier for human contributors to read and
+understand. A simpler structure will tend to promote maintenance by those who
+have less awareness of WPT infrastructure.
+
+## Risks
```
jcscottiii, line 162: Are we getting rid of the "**" syntax? I'm fine with keeping it or getting rid of it. But I think it would be worth adding an example if we decide to keep it so we can discuss the implications.

If we drop "**" , the RFC should explicitly clarify whether subdirectory path searches are allowed in the pattern rules (e.g., - "webappapis/user-prompts/*": ["alerts"]) so maintainers can centralize rules in parent directories and minimize sparse files. If not, we should explicitly highlight in the RFC that a high volume of sparse files is an expected and accepted trade-off.

rfcs/web_features_exclusions.md
```diff
@@ -0,0 +1,192 @@
...
+`NULL` to signify files which should not belong to any particular web-feature.
+
+In the example above, to exclude the file named `foo.html` from *any*
+web-feature, one would write:
+
+```yaml
+features:
+- foo.html: NULL
+- print-*: print
+- "*": alerts
```
jcscottiii: The old schema used explicit properties (name and files). Because of this, it was trivial to add new metadata fields in the future (e.g., bug_url, reason) without breaking the parser.

We could change it be something like: 
```suggestion
```yaml
features:
- pattern: "print-*"
  features: ["print"]
- pattern: "*"
  features: ["alerts"]
- pattern: "foo.html"
  features: []
  reason: "Excluded due to flakiness" # Easily extensible in the future. And not just rely on yaml comments that don't make it into the metadata
```

jcscottiii (commented):
This looks a lot better. I have some questions. Feel free to post your thoughts on them.

rfcs/web_features_exclusions.md
jugglinmike: I'm on board with nixing `NULL` in favor of `[]`. It reduces the number of ways to express the same thing (`NULL`, `[]`, or even `[NULL]`). `[]` also happens to be fewer characters without (in my opinion) being any less clear.

I'm on the fence about requiring arrays, though. While it can be slightly more concise (no need to wrap the strings in quotes), it's still non-zero cognitive/ergonomic overhead for what is (currently) the overwhelmingly more common use-case. 99% of the tests belong to just one web-feature as of this writing:

    $ python3 -m json.tool < WEB_FEATURES_MANIFEST.json | grep "\"/" | sort | uniq -c | sort -n | grep -Eo '^ *[0-9]+' | uniq -c
      54229       1
        416       2
          1       3

rfcs/web_features_exclusions.md
jugglinmike: I'm in favor of keeping it for this iteration. I could see how that change might be positive, but I think it would require a bit of dedicated research to justify. It will also be easier to make that change in a dedicated change set. I've added an example to this RFC.

rfcs/web_features_exclusions.md
jugglinmike: If we feel that future extensions are sufficiently likely (and if we think that another schema revision would be sufficiently disruptive), then I would vote for keeping the existing design as-is. As noted in this latest version of the RFC, the schema currently in use is expressive enough to implement the new semantics. And as you note, the schema currently in use also supports extension. In addition, it is the most concise out of all the designs we've considered thus far.

That said, I've read in @jgraham's feedback a general interest in promoting usability for contributors today. The latest proposal seems optimized for that. It might be easier to justify a design that favors future work if we had some indication about the likelihood/timeline for that work.

jgraham (approved):

rfcs/web_features_exclusions.md
```diff
@@ -0,0 +1,208 @@
...
+In order to support the case of cross-cutting tests, it should be possible to
+express membership in multiple web-features via a list of web-feature IDs in
+place of a single web-feature ID.
+
+In the example above, to include the file named `foo.html` in both the `print`
+web-feature *and* the `alerts` web-feature, one would write:
+
+```yaml
+features:
+- foo.html: [print, alerts]
```
jgraham, line 100: As a consumer "always a list" is easier to work with than "sometimes a list, sometimes a string", but it's clearly a tradeoff for authors and I don't feel strongly.

rfcs/web_features_exclusions.md
jgraham: If we wanted to extend it we could do it like
```yaml
features:
  - print-*:
      features: ["print"]
      reason: "All the print reftests"
```

That's slightly less verbose in any case and would continue to allow the optimisation where the right hand side can just be a string or a list and that's interpreted as identical to an object with just a "features" key.

jcscottiii: I like @jgraham's version of this. Thoughts on adopting this way? @jugglinmike. I am in favor of this. For now, `features` would be the only key.

rfcs/web_features_exclusions.md
jcscottiii: I am in favor of "always a list". Do you mind making that change @jugglinmike ?

Prior art: In the Web DX community, we previously had `string | string []`  for some fields because we thought it would help. But in the end, it made things more complicated. https://github.com/web-platform-dx/web-features/pull/3184

rfcs/web_features_exclusions.md
jcscottiii: Thanks for calculating that. I would still vote in favor of always requiring arrays. See my comment here: https://github.com/web-platform-tests/rfcs/pull/237#discussion_r3421781457

jcscottiii (commented):
@jugglinmike Sorry it took so long on the follow up review. PTAL at the unresolved conversations.

rfcs/web_features_exclusions.md
jugglinmike: @jgraham's suggestion is an extension to syntax currently proposed by the RFC, not an alternative to it. I'm in favor of supporting an extension because it would mean we would only pay the price of overhead when we need it, and because I still don't have any insight into the likelihood/timeline of additional metadata.

Although making that syntax the only option would simplify the parser, it would still increase verbosity well beyond what our current needs require. I feel that the existing syntax strikes a much stronger balance between parser simplicity, maintainability, and concision.

In any case, I think all extant proposals which include a property name for the list of web-feature IDs should be revised. In the case of the existing syntax, the property name is "name". It should instead be "ids" both to reflect the plurality of the value and to align with the WebDX terminology ([where a web-feature's "name" is distinct from its "ID"](https://github.com/web-platform-dx/web-features/blob/a5d5fc6e3688a66a434833779fa8a1878276d654/docs/guidelines.md)). The name "ids" is especially appealing in this latest alternative because it avoids using the same property for two distinct parts of the data structure.

It seems like we're close; here are the options that are on the table right now:

1. keep the existing syntax (but replace `"name"` with `"ids"`):
   ```yaml
   features:
     - ids: []
       files:
       - foo.html
     - ids: [print]
       files:
       - print-*
   ```
2. use the abbreviated syntax with an optional extension:
   ```yaml
   features:
   - foo.html: []
   - print-*:
       ids: [print-*]
   ```
3. use the extended form of the abbreviated syntax only:
   ```yaml
   features:
   - foo.html:
       ids: []
   - print-*:
       ids: [print-*]
   ```

My preference is option 2 followed by option 1 followed by option 3. What do you folks think?

rfcs/web_features_exclusions.md
jugglinmike: Sounds good to me!

rfcs/web_features_exclusions.md
jugglinmike: Done!

rfcs/web_features_exclusions.md
jcscottiii: We can go with option 2.

For background: Initially I was in favor of option 3 because I was looking ahead to increased automation (e.g., tooling like `wpt-gen` or LLMs). Consistent schemas (always objects) are generally easier for tools to write and modify programmatically without bugs.

Since these YAML files are internal to WPT infrastructure and changes here won't impact the final consumers of `WEB_FEATURES_MANIFEST.json`, I felt we had the flexibility to prioritize that tool-friendliness.

But after thinking about it, I agree with what you said that keeping the barrier to entry low for human contributors is really important (which option 2 does). 

One thing I wanted to point out. I like the fact that you mentioned changing name to ids. Maybe for option 2, we could do the same for the high level key and change to `files` or `rules`. Either is okay. But, the fact that the underlying list is essentially pointing to files or rules makes more sense than features now.

```yaml
files: # or rules here.
  - foo.html: [print, alerts]
  - print-*:
      ids: [print]
      reason: "All the print reftests"
```

jugglinmike: Awesome!

Swapping out "features" makes a lot of sense to me! I like "rules" more than "files" because the values describe more than just files.

This conversation about extensibility is worth highlighting. I've tried to do that by framing it as a risk within the RFC itself: how the proposal appears to be less extensible, but how we've considered the direction that future additions can make. This will help guide our grandchildren when they find the time to add metadata.

Thanks @jcscottiii! Thanks @jgraham!

rfcs/web_features_exclusions.md
```diff
@@ -201,6 +201,29 @@ web-feature in a given directory[^2] is approximately 2.362 with a standard
 deviation of approximately 5.7487. This reduces the benefit of optimizing for
 the case of web-features with many associated file-matching patterns.
 
+### Extendability
```
jcscottiii, line 204: Thanks for adding this section! I get that this is a risk. But supporting both the short hand and extended form out of the box will be good long term!

jcscottiii (approved):

PR #238, opened 2026-05-13
RFC 238: Websocket over HTTP/3 test server

Patil-Mayur:
Hi web-platform-tests,

We propose introducing a new test server for WebSocket over HTTP/3 to enable writing corresponding WPTs.
[Preview version](https://github.com/web-platform-tests/rfcs/blob/12476590ebe2c13eb217410a984ed812e5aa20d6/rfcs/websockets_http3.md)

Prototype implementation for reference only: https://chromium-review.googlesource.com/c/chromium/src/+/7835151 

FYI 
@bashi , @ricea, @nidhijaju

### DIFF
--- a/rfcs/websockets_http3.md
+++ b/rfcs/websockets_http3.md
@@ -1,0 +1,72 @@
+# RFC 85: WebSocket over HTTP/3 Test Server
+
+## Summary
+
+Add a [WebSocket over HTTP/3](https://datatracker.ietf.org/doc/html/rfc9220) server to wpt for [WebSocket Web API](https://websockets.spec.whatwg.org/) tests. The server requires Python 3 and reuses existing pywebsocket3 handlers where possible.
+
+This follows the [WebTransport over HTTP/3 test server](https://github.com/bashi/rfcs/blob/08cd404ee62df4dd93b11175310d44fa98187c96/rfcs/webtransport_h3_test_server.md) model and extends the [WebSocket over HTTP/2](https://github.com/bashi/rfcs/blob/08cd404ee62df4dd93b11175310d44fa98187c96/rfcs/websockets_http2.md) approach to WebSocket over HTTP/3.
+
+## Details
+
+### Implementation
+
+The server will live under WPT's `tools/websocket_h3` directory and use `aioquic` for QUIC and HTTP/3. It advertises `SETTINGS_ENABLE_CONNECT_PROTOCOL = 1` and handles RFC 9220 Extended CONNECT requests with `:protocol` set to `websocket`.
+
+After a successful handshake, DATA frames on the CONNECT stream carry RFC 6455 WebSocket frames.
+
+### Handshake
+
+The server validates that:
+
+* `:method` is `CONNECT`;
+* `:protocol` is `websocket`;
+* an authority value is present; and
+* a handler exists for the requested `:path`.
+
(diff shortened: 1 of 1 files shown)


rfcs/websockets_http3.md
```diff
@@ -0,0 +1,72 @@
...
+// META: variant=?wpt_flags=h3
+```
+
+Helpers can then select `{{ports[h3][0]}}` and use `wss://` for the WebSocket URL. The HTTP/3 server remains opt-in so WPT does not start an extra QUIC server for unrelated runs.
+
+### Integration
+
+`wptrunner` should recognize `h3` tests but keep them disabled unless H3 support is explicitly enabled, for example with `--enable-h3`. When enabled, `wptrunner` includes those tests, and `wptserve` launches and monitors the WebSocket over HTTP/3 server.
+
+Chromium should use the configured H3 port and force QUIC for that origin.
```
gsnedders: I don't think how Chromium should be configured it really a relevant part of the question from a WPT point of view, unless it's something particularly complex?

rfcs/websockets_http3.md
```diff
@@ -0,0 +1,72 @@
...
+
+This RFC does not change the WebSocket API, add JavaScript API surface, or replace existing WebSocket servers. It only adds infrastructure to run existing WebSocket API tests over HTTP/3.
+
+Support for WebSocket over HTTP/1.1 and WebSocket over HTTP/2 remains unchanged.
+
+## Risks
+
+The main risks are the extra opt-in daemon, the adapter between asynchronous HTTP/3 streams and pywebsocket3's synchronous handler API, and keeping static-resource behavior aligned with `wptserve`.
+
+The Web Networking and Ecosystem Infra owners should maintain the integration point and server behavior.
```
gsnedders: Most of the actual WPT tooling is maintained, in principle, by WPT organisationally — knowing there are people willing to help is significant, but not a major point.

And also: this should be clear that these are Chromium teams.

gsnedders (approved):
Especially given this doesn't actually add any new dependencies, and when this shouldn't be a significant amount of code, I'm not even sure this is really RFC worthy.

> Experimental CL for reference which is following this doc changes: https://chromium-review.googlesource.com/c/chromium/src/+/7835151

I do think it might be worthwhile going through the review within the WPT project, rather than doing it in Chromium and relying on upstreaming, because I think the interesting questions are really all specific implementation ones?

rfcs/websockets_http3.md
Patil-Mayur: Thanks, removed the Chromium-specific part. Also, the integration section now only describes the WPT-side behavior when the h3 variant is included.

rfcs/websockets_http3.md
Patil-Mayur: Done, clarified that WPT owns this, and the Chromium teams can help support it.

Patil-Mayur:
> Especially given this doesn't actually add any new dependencies, and when this shouldn't be a significant amount of code, I'm not even sure this is really RFC worthy.
> 
> > Experimental CL for reference which is following this doc changes: https://chromium-review.googlesource.com/c/chromium/src/+/7835151
> 
> I do think it might be worthwhile going through the review within the WPT project, rather than doing it in Chromium and relying on upstreaming, because I think the interesting questions are really all specific implementation ones?

That makes sense. I updated the description to make the Chromium CL just a prototype/reference, and I’ll take the actual implementation review through WPT.

DanielRyanSmith (approved):

Patil-Mayur:
Is this change ready to be merged?
Right now, I don’t have the access needed to merge them.

Issue #202, opened 2024-07-08
Set policy for LLM-generated tests


jgraham:
CC @web-platform-tests/wpt-core-team 

I was recently asked about the policy for using LLMs to generate tests that are submitted to wpt. Currently we don't have any explicit policy on this, but I think we need one, so it's clear to test authors what's permissible, and to downstream consumers whether wpt is in line with any other policies they might have with regard to LLMs.

To be clear, I expect that any policy here would not affect review requirements i.e. we'd still require human review for all tests to ensure they're correct and easy to follow.

Ms2ger:
Given their tendency to create output that looks plausible but isn't actually correct, I'd lean towards "don't allow" or at least increase review requirements on them. We've been relatively permissive in accepting tests based on the fact that a human has already reasoned their way through the spec while writing; if that's no longer a valid assumption, we should take that into account.

tabatkins:
Yes, I'm overall on the "disallow officially" stance. If they *are* LLM-generated, but are good/correct enough we can't tell, then fine, but in my experience people who use LLMs like this are not particularly concerned about the quality in the first place.

jgraham:
We also need to consider cases where the test is paritally generated but human-modified (e.g. with copilot-like tools).

The practical arguments I've heard in favour of allowing (partially) LLM-generated tests are basically:
* People are going to do it anyway, and it will be difficult to police.
* For people who do follow the policy being forbidden from using productivity tooling for wpt will make them more likely to write non-wpt tests.

Speaking personally, I'm definitely worried about the potential for lower-quality tests being written using these technologies, but as a practical matter I don't know how much we can enforce, so I'd only want to _ban_ them if there's some strong non-technical reason (e.g. vendors who would be unable to use wpt). However I'd be interesting in ideas for marking such tests (assuming good faith contributors) so that we could verify whether they cause more problems in the longer term compared to purely human-generated tests.

WeizhongX:
Maybe we need clarify the license issue associate with code generated with LLM? One of the reason LLM is banned from some corporation is because the license of LLM-generated code can be unclear, thus not safe to use.

What is the policy regarding the license now for WPTs contributed by a human?

jgraham:
I think broadly it's the submitter's responsibility to ensure that they're able to contribute it under the terms of the [3-Clause BSD License](https://github.com/web-platform-tests/wpt/blob/master/LICENSE.md).

zcorpan:
[...]

PR #239, opened 2026-05-20
RFC 239: Policy on LLM assistance in contributions

jugglinmike:
This initial draft takes a maximalist approach (and a permissive stance) to promote a robust and grounded discussion.

[Rendered](https://github.com/bocoup/wpt-rfcs/blob/llm-policy/rfcs/llm_policy.md)

### DIFF
--- a/rfcs/llm_policy.md
+++ b/rfcs/llm_policy.md
@@ -65,10 +65,6 @@ referenced both from the project's `README.md` file and the
 > Every pull request must be initiated by one human. That person must author
 > the pull request description, understand every change proposed, and be
 > prepared to engage in technical discussion regarding those changes.
->
-> #### Discussion
->
-> Discussion
 
 ## Risks
 
--- a/rfcs/llm_policy.md
+++ b/rfcs/llm_policy.md
@@ -1,0 +1,85 @@
+# RFC 239: Policy on LLM assistance in contributions
+
+## Summary
+
+Introduce guidelines for acceptable use of large-language models when
+contributing to web-platform-tests.
+
+## Background
+
+[#202 Set policy for LLM-generated
+tests](https://github.com/web-platform-tests/rfcs/issues/202) includes evidence
+for public interest in a formal policy for LLM usage in authoring contributions
+to WPT.
+
+The Chrome team is exploring applications of LLMs for detecting coverage gaps
+and for filling those gaps with generated code. ([Project
+repository](https://github.com/GoogleChromeLabs/wpt-gen), [April 2026
+presentation](https://www.youtube.com/watch?v=9r0PBbJFLoM))
+
+A few examples of policies on LLM use in FOSS contributions:
+
+- permissive
+  - [ghostty/AI_POLICY.md at main · ghostty-org/ghostty](https://github.com/ghostty-org/ghostty/blob/main/AI_POLICY.md)
+  - [Policy about LLM generated code from PRs · Issue #28335 · opencv/opencv](https://github.com/opencv/opencv/issues/28335)
+  - [CONTRIBUTING.md: Guidelines relevant to AI-assisted contributions by gasche · Pull Request #14052 · ocaml/ocaml](https://github.com/ocaml/ocaml/pull/14052)
+  - [LLVM AI Tool Use Policy — LLVM 23.0.0git documentation](https://llvm.org/docs/AIToolPolicy.html)
+- prohibitive
+  - [Code of Conduct ⚡ Zig Programming Language](https://ziglang.org/code-of-conduct/#strict-no-llm-no-ai-policy)
+  - [Getting Started - The Servo Book](https://book.servo.org/contributing/getting-started.html#ai-contributions)
+
+## Details
+
+Proposed text:
+
+> ### For Individual Contributors
+>
+> #### Disclosure
+>
+> Contributions that contain substantial amounts of tool-generated content must
+> be labeled as such.
+>
+> #### Attribution
+>
+> Commits generated entirely by an LLM must be attributed to the LLM in the
+> "Author" field.
+>
+> #### Understanding
+>
+> Every pull request must be initiated by one human. That person must author
+> the pull request description, understand every change proposed, and be
+> prepared to engage in technical discussion regarding those changes.
+>
+> ### For Trusted External Review
+>
+> Some external projects conduct review which the WPT maintainers recognize as
+> authoritative. From rendering engines like Gecko to dedicated test suites
+> like WASM, patches merged in these projects are incorporated into WPT without
+> further review. The policy outlined by this document does not apply to these
+> contributions; the external projects are trusted to determine their own
+> mechanisms for quality assurance.
+
+## Risks
+
+### Discouraging volunteers
+
+All but the most permissive policy is effectively another hurdle to
+contributing to the project. Friction in the contribution process could deter
+people who might otherwise volunteer their time to help improve the project.
+
+In some sense, adding friction is the goal of this policy. New technology has
+removed barriers which previously restricted unqualified individuals from
+participation. Rather than introducing more restrictions on good-faith actors,
+an ideal policy will buttress eroded structural barriers with more intentional
+social ones.
+
+### Encouraging low-value contributions
+
+All but the most restrictive policy could be interpreted as an invitation to
+take shortcuts which undermine the quality of contributions.
+
+However, it will not be possible to strictly enforce any policy. It inevitably
+falls on contributors to follow rules and for administrators to police
+transgressions. Respect in public works projects is never guaranteed; policies
+exist only to make expectations clear (this is the same dynamic that guides the
+design and enforcement of codes of conduct).
--- a/rfcs/llm_policy.md
+++ b/rfcs/llm_policy.md
@@ -24,25 +24,41 @@ A few examples of policies on LLM use in FOSS contributions:
   - [Policy about LLM generated code from PRs · Issue #28335 · opencv/opencv](https://github.com/opencv/opencv/issues/28335)
   - [CONTRIBUTING.md: Guidelines relevant to AI-assisted contributions by gasche · Pull Request #14052 · ocaml/ocaml](https://github.com/ocaml/ocaml/pull/14052)
   - [LLVM AI Tool Use Policy — LLVM 23.0.0git documentation](https://llvm.org/docs/AIToolPolicy.html)
+  - [Chromium AI Coding Policy](https://chromium.googlesource.com/chromium/src/+/4f44016dfbd4fcd890694c00d7f9ec6dcefe4955/agents/ai_policy.md)
+  - [Firefox AI Coding Policy](https://github.com/mozilla-firefox/firefox/blob/1f7030c8de8f2b349c7d91d7b5a3253c109a1cc1/docs/contributing/ai-coding.md)
 - prohibitive
   - [Code of Conduct ⚡ Zig Programming Language](https://ziglang.org/code-of-conduct/#strict-no-llm-no-ai-policy)
   - [Getting Started - The Servo Book](https://book.servo.org/contributing/getting-started.html#ai-contributions)
 
 ## Details
 
-Proposed text:
+The following text describes the policy in full and will be maintained in a
+dedicated document within WPT's `docs/writing-tests/` directory (which will be
+referenced both from the project's `README.md` file and the
+`docs/writing-tests/index.md` file):
 
-> ### For Individual Contributors
+> ### Guidelines for acceptable LLM use
+>
+> The use of large language models (LLMs) as tools to help author commits to
+> this repository is allowed with the stipulations described below.
+> Contributors who repeatedly fail to adhere to these guidelines may be banned
+> from contributing to this project.
 >
 > #### Disclosure
 >
-> Contributions that contain substantial amounts of tool-generated content must
-> be labeled as such.
+> If LLMs are used as a significant input to a commit, authors are encouraged
+> to include details about how they were used as part of the commit message in
+> order to help review and future understanding of the code.
+>
+> Human-authored code discourse (e.g. issue descriptions, pull request
+> descriptions, and responses to discussion threads) should not include
+> LLM-generated content in the main text; any such content must be clearly
+> labelled and placed inside a `<details>` element.
 >
 > #### Attribution
 >
-> Commits generated entirely by an LLM must be attributed to the LLM in the
-> "Author" field.
+> All commits must be attributed to the human who is taking responsibility for
+> them, regardless of LLM use.
 >
 > #### Understanding
 >
@@ -50,14 +66,9 @@ Proposed text:
 > the pull request description, understand every change proposed, and be
 > prepared to engage in technical discussion regarding those changes.
 >
-> ### For Trusted External Review
+> #### Discussion
 >
-> Some external projects conduct review which the WPT maintainers recognize as
-> authoritative. From rendering engines like Gecko to dedicated test suites
-> like WASM, patches merged in these projects are incorporated into WPT without
-> further review. The policy outlined by this document does not apply to these
-> contributions; the external projects are trusted to determine their own
-> mechanisms for quality assurance.
+> Discussion
 
 ## Risks
 
@@ -76,7 +87,10 @@ social ones.
 ### Encouraging low-value contributions
 
 All but the most restrictive policy could be interpreted as an invitation to
-take shortcuts which undermine the quality of contributions.
+take shortcuts which undermine the quality of contributions. Any permissive
+policy might be taken as encouragement to rely on fallible tools (LLMs are
+particularly susceptible to certain kinds of test-writing errors, such as
+over-fitting and fabrication).
 
 However, it will not be possible to strictly enforce any policy. It inevitably
 falls on contributors to follow rules and for administrators to police
--- a/rfcs/llm_policy.md
+++ b/rfcs/llm_policy.md
@@ -53,7 +53,9 @@ referenced both from the project's `README.md` file and the
 > Human-authored code discourse (e.g. issue descriptions, pull request
 > descriptions, and responses to discussion threads) should not include
 > LLM-generated content in the main text; any such content must be clearly
-> labelled and placed inside a `<details>` element.
+> labelled and placed inside a `<details>` element. A `<blockquote>` element
+> may instead be used to designate text if it is no longer than one paragraph
+> in length.
 >
 > #### Attribution
 >


gsnedders:
Fixes https://github.com/web-platform-tests/rfcs/issues/202

rfcs/llm_policy.md
```diff
@@ -0,0 +1,85 @@
...
+A few examples of policies on LLM use in FOSS contributions:
+
+- permissive
+  - [ghostty/AI_POLICY.md at main · ghostty-org/ghostty](https://github.com/ghostty-org/ghostty/blob/main/AI_POLICY.md)
+  - [Policy about LLM generated code from PRs · Issue #28335 · opencv/opencv](https://github.com/opencv/opencv/issues/28335)
+  - [CONTRIBUTING.md: Guidelines relevant to AI-assisted contributions by gasche · Pull Request #14052 · ocaml/ocaml](https://github.com/ocaml/ocaml/pull/14052)
+  - [LLVM AI Tool Use Policy — LLVM 23.0.0git documentation](https://llvm.org/docs/AIToolPolicy.html)
+- prohibitive
+  - [Code of Conduct ⚡ Zig Programming Language](https://ziglang.org/code-of-conduct/#strict-no-llm-no-ai-policy)
+  - [Getting Started - The Servo Book](https://book.servo.org/contributing/getting-started.html#ai-contributions)
```
gsnedders, line 31: I feel like there's three others which are notably relevant here: [Chromium's](https://chromium.googlesource.com/chromium/src/+/4f44016dfbd4fcd890694c00d7f9ec6dcefe4955/agents/ai_policy.md), and [Firefox's](https://github.com/mozilla-firefox/firefox/blob/1f7030c8de8f2b349c7d91d7b5a3253c109a1cc1/docs/contributing/ai-coding.md), given they are two of the five repos which have approval to land changes in WPT without further review. (WebKit and Test262 do not currently have policies — [TC39's](https://github.com/tc39/how-we-work/blob/68d1201c1afb4a4155d1ff572d41bb8028f82b3e/AI_POLICY.md) explicitly does not apply to code.)

rfcs/llm_policy.md
```diff
@@ -0,0 +1,85 @@
...
+>
+> #### Disclosure
+>
+> Contributions that contain substantial amounts of tool-generated content must
+> be labeled as such.
+>
+> #### Attribution
+>
+> Commits generated entirely by an LLM must be attributed to the LLM in the
+> "Author" field.
```
gsnedders: This feels problematic. If we attribute a PR to Claude, Gemini, or OpenAI's GPT, if I try and contact the author… well, I don't think Anthropic, Google, or OpenAI are going to be very helpful?

Both Chromium and Firefox's policies are crystal clear that humans are still the authors and must self-review before submitting.

Therefore, when there's still a human very much in the loop who is required to self-review, it does not seem reasonable to consider the LLM the author — and the Chromium policy is explicit that, "Authors must attest that the code they submit is their original creation, regardless of whether AI tooling was used".

rfcs/llm_policy.md
```diff
@@ -0,0 +1,85 @@
...
+## Details
+
+Proposed text:
+
+> ### For Individual Contributors
+>
+> #### Disclosure
+>
+> Contributions that contain substantial amounts of tool-generated content must
+> be labeled as such.
```
gsnedders: Neither Chromium nor Firefox require this today, and it's entirely plausible we've already had commits land into WPT via exports which don't meet this bar.

That said, Chromium's policy here is currently:

> To aid reviewers, authors should flag areas that they are not confident about that had AI assistance. 

This is maybe a weaker form, and hopefully something more in line with existing contributions.

rfcs/llm_policy.md
```diff
@@ -0,0 +1,85 @@
...
+> prepared to engage in technical discussion regarding those changes.
+>
+> ### For Trusted External Review
+>
+> Some external projects conduct review which the WPT maintainers recognize as
+> authoritative. From rendering engines like Gecko to dedicated test suites
+> like WASM, patches merged in these projects are incorporated into WPT without
+> further review. The policy outlined by this document does not apply to these
+> contributions; the external projects are trusted to determine their own
+> mechanisms for quality assurance.
```
gsnedders: This feels like it should probably be at least in part in another RFC that tries to define our existing policies?

As far as I'm aware, there's currently five repos which have approval to incorporate based on downstream review — Chromium, Firefox, Servo, Test262, and WebKit.

My understanding of the unwritten policy is we trust downstream reviewers; I can't even find the various places where we've elucidated parts of the policy over the years.

rfcs/llm_policy.md
```diff
@@ -0,0 +1,85 @@
...
+  - [Policy about LLM generated code from PRs · Issue #28335 · opencv/opencv](https://github.com/opencv/opencv/issues/28335)
+  - [CONTRIBUTING.md: Guidelines relevant to AI-assisted contributions by gasche · Pull Request #14052 · ocaml/ocaml](https://github.com/ocaml/ocaml/pull/14052)
+  - [LLVM AI Tool Use Policy — LLVM 23.0.0git documentation](https://llvm.org/docs/AIToolPolicy.html)
+- prohibitive
+  - [Code of Conduct ⚡ Zig Programming Language](https://ziglang.org/code-of-conduct/#strict-no-llm-no-ai-policy)
+  - [Getting Started - The Servo Book](https://book.servo.org/contributing/getting-started.html#ai-contributions)
+
+## Details
+
+Proposed text:
```
gsnedders: Proposed for where?

rfcs/llm_policy.md
```diff
@@ -0,0 +1,85 @@
...
+> ### For Trusted External Review
+>
+> Some external projects conduct review which the WPT maintainers recognize as
+> authoritative. From rendering engines like Gecko to dedicated test suites
+> like WASM, patches merged in these projects are incorporated into WPT without
+> further review. The policy outlined by this document does not apply to these
+> contributions; the external projects are trusted to determine their own
+> mechanisms for quality assurance.
+
+## Risks
```
gsnedders, line 71: I think it's worthwhile including at least a few more technical risks:

1. Contributions of tests generated by an LLM closely looking at a specific implementation's code, matching that implementation, rather than the spec. (This is, of course, already an issue — but could inevitably become more of a problem if we get more, larger contributions.)
2. Contributions not matching the spec at all. I've seen this mostly with trying to generate tests to assert ordering of things which end of using [HTML's parallelism](https://html.spec.whatwg.org/multipage/infrastructure.html#parallelism) and [HTML's event loops](https://html.spec.whatwg.org/multipage/webappapis.html#event-loop); that case is especially annoying because it can lead to flaky tests.

gsnedders (changes requested):
Thanks for trying to tackle this!

I think we should also write something about use of LLMs for review, if nothing else.

rfcs/llm_policy.md
tabatkins: Strong agree on this point for all the reasons you give. The Author field's purpose is to give a contact for problems/question, not assign blame. Listing an LLM is worthless there.

And also, yeah, assigning authorship to an LLM is abrogating your responsibility as an engineer to commit useful code that you understand.

martinthomson:
In terms of a survey, [this](https://redmonk.com/kholterhoff/2026/02/26/generative-ai-policy-landscape-in-open-source/) is a very useful resource.  It's fairly comprehensive and it provides a useful rubric for classification across quality, copyright, and ethics.

rfcs/llm_policy.md
jgraham: +1 Commits that attribute authorship to a tool rather than the person using the tool is something that the policy should explicitly ban, not require.

jgraham (changes requested):
My view is that the overall policy should be something like:
* The use of LLMs as tools to help author commits to this repository is allowed
* All contributions must be attributed to a person. That person is expected to fully understand and take responsibility for their contribution, irrespective of any tooling used.
* If LLMs are used as a significant input to a commit, authors are encouraged to include details about how they were used as part of the commit message, in order to help review and future understanding of the code (similar to the way that a commit created using a script might include details of the script).
* People who habitually submit low quality PRs, PRs they don't understand, or otherwise make contributions that are deemed to waste the time of other project contributors or downstream users of the project may be banned.

jgraham:
https://github.com/BurntSushi/ripgrep/blob/master/AI_POLICY.md is also interesting, in particular the "no AI in issues or issue comments" part, with a specific carve out for use cases like translation.

DanielRyanSmith:
I think I'm mostly in agreement with the discussions here so far. I do have one scenario where I think the policy could be expanded on.

> All contributions must be attributed to a person. That person is expected to fully understand and take responsibility for their contribution, irrespective of any tooling used.

I think it would be reasonable to allow a test to be proposed or generated, and although the original author does not have a strong fundamental understanding of the change, it would be feasible to have that change reviewed thoroughly by a domain expert before submission.

For example, using the experimental tool [WPT-Gen](https://github.com/GoogleChromeLabs/wpt-gen) can uncover potential gaps in test coverage reading the spec, and then generate tests to cover those gaps. Hypothetically, we could occasionally run this automatically for certain features. If we have domain expert reviewers who are willing to thoroughly review generated tests, it seems like it would be useful to have an option in the LLM policy to allow for this. Naturally, this does put a lot of burden on the already thankless job of being a reviewer, but it still seems like a viable contribution avenue.

I imagine this will not be terribly common, but it is something that doesn't sound unrealistic.

tabatkins:
As a reviewer, I am not interested in reviewing tests that the author does not understand. There are a lot more test authors than test reviewers. Someone using wpt-gen to create tests should then review those tests and make sure they're doing reasonable things before tossing them at a reviewer. (Generally speaking, I'm not happy with things that further unbalance the effort expectations in the submitter/reviewer dynamic. I'm happy to do thorough review *to teach a person* how to handle an area; I am not happy to do the same to approve/reject a large batch of unreviewed tests when the effort will not meaningfully improve future interactions.))

jgraham:
To expand a bit on my proposal for issue/PR comments:
* The project may choose to use tooling that produces LLM-generated comments. Such comments must always be explicitly associated with a bot account. The core team gets final discretion about which bots / tooling are permitted.
* Human authored comments should not include LLM-generated content in the main text; any such content must be clearly labelled and placed inside a `<details>` element.

DanielRyanSmith:
> As a reviewer, I am not interested in reviewing tests that the author does not understand. There are a lot more test authors than test reviewers. Someone using wpt-gen to create tests should then review those tests and make sure they're doing reasonable things before tossing them at a reviewer. (Generally speaking, I'm not happy with things that further unbalance the effort expectations in the submitter/reviewer dynamic. I'm happy to do thorough review *to teach a person* how to handle an area; I am not happy to do the same to approve/reject a large batch of unreviewed tests when the effort will not meaningfully improve future interactions.))

I don't disagree with any of this, and I should clarify that I'm not advocating for reviewers now being responsible for reviewing AI-generated PRs that authors do not understand - that would absolutely not be beneficial to the project.

I do think that there's a scenario where someone is working to identify and fill gaps in test coverage using LLMs for specific features, and if that person has already interacted with a specific reviewer who is in agreement and willing to review these tests, then that should not be disbarred from contribution if these tests are being reviewed by a competent reviewer.

This is mostly a corner case, but it is something that I wanted to bring up as well.

rfcs/llm_policy.md
jugglinmike: To me, those look like specific failure modes of LLMs. I think they'd be more helpful to elucidate what we mean by "low-value contributions" rather than as additional risks that are distinct from low-value contributions. I've just pushed a commit incorporating them in that way.

jugglinmike:
Thanks, folks! During today's meeting, I offered to incorporate all the feedback that hasn't received explicit pushback on this discussion thread. While carrying that out, I've found one small conflict between the suggestions above.

@gsnedders [gestured toward](https://github.com/web-platform-tests/rfcs/pull/239#discussion_r3284469366) a statement in the Chromium project's policy:

> That said, Chromium's policy here is currently:
>
> > To aid reviewers, authors should flag areas that they are not confident about that had AI assistance.
>
> This is maybe a weaker form, and hopefully something more in line with existing contributions.

However, @jgraham [later made a suggestion](https://github.com/web-platform-tests/rfcs/pull/239#pullrequestreview-4380258465) which undermines that text:

> - All contributions must be attributed to a person. That person is expected to fully understand and take responsibility for their contribution, irrespective of any tooling used.
> - If LLMs are used as a significant input to a commit, authors are encouraged to include details about how they were used as part of the commit message, in order to help review and future understanding of the code (similar to the way that a commit created using a script might include details of the script).

If a person must "fully understand and take responsibility for their contribution," then there doesn't seem to be a need to guide folks who "are not confident" about their work. I've incorporated the part from @jgraham's exclusively because @gsnedders has not pushed back on it.

gsnedders:
> I do think that there's a scenario where someone is working to identify and fill gaps in test coverage using LLMs for specific features, and if that person has already interacted with a specific reviewer who is in agreement and willing to review these tests, then that should not be disbarred from contribution if these tests are being reviewed by a competent reviewer.

I think there's two meaningful cases here:

1. The specific reviewer doesn't have access to the same models as the other contributor, and
2. The specific reviewer is delegating the task of actually generating the tests to the other contributor.

In either case, I think I'd rather push the other contributor to provide the tests to the specific reviewer in some other way — e.g., by pushing the branch to their fork without opening a PR so the specific reviewer can take over and do any initial fixup work before opening a PR.

I, like others here, don't want to deal with PRs that the author hasn't themselves reviewed.

> If a person must "fully understand and take responsibility for their contribution," then there doesn't seem to be a need to guide folks who "are not confident" about their work. I've incorporated the part from @jgraham's exclusively because @gsnedders has not pushed back on it.

While I recognise @jgraham was clearly deriving text from the Firefox policy, I do wonder if "fully understand" is a somewhat unreasonable bar! I think most of us working on the infrastructure have touched things without fully understanding the code?

That said, I don't have a particularly strong opinion here!

rfcs/llm_policy.md
```diff
@@ -0,0 +1,99 @@
...
+> #### Disclosure
+>
+> If LLMs are used as a significant input to a commit, authors are encouraged
+> to include details about how they were used as part of the commit message in
+> order to help review and future understanding of the code.
+>
+> Human-authored code discourse (e.g. issue descriptions, pull request
+> descriptions, and responses to discussion threads) should not include
+> LLM-generated content in the main text; any such content must be clearly
+> labelled and placed inside a `<details>` element.
```
annevk: This feels too onerous. What if people use it for translation? What if they copy-and-pasted a tiny snippet?

rfcs/llm_policy.md
```diff
@@ -0,0 +1,99 @@
...
+>
+> #### Attribution
+>
+> All commits must be attributed to the human who is taking responsibility for
+> them, regardless of LLM use.
+>
+> #### Understanding
+>
+> Every pull request must be initiated by one human. That person must author
```
annevk, line 67: Should this be at least one? Don't we sometimes get team contributions?

rfcs/llm_policy.md
```diff
@@ -0,0 +1,99 @@
...
+> the pull request description, understand every change proposed, and be
+> prepared to engage in technical discussion regarding those changes.
+>
+> #### Discussion
+>
+> Discussion
```
annevk: Is this meant to be part of the policy?

rfcs/llm_policy.md
jgraham: The ripgrep policy linked above has specific text about translation, which after noting its usefulness says the same policy applies: 
> If using AI for translation, we recommend writing in your native language and including the AI translation in a quote block.

That makes a lot of sense to me. If you only auto-translated the content then providing the source language as canonical allows others to redo the translation potentially using a better tool.

The tiny snippet thing feels very edge-casy, but we can try to reword. Are there specific scenarios you have in mind? Personally I'm not sure why someone would take just a few words directly and intermix it with otherwise handwritten text. If it's not substantive (e.g. you're just copying phrasing) then I'd prefer that you didn't do that, if it is substantive (e.g. you're relying on an LLM for a point of fact) I'd prefer that you are explicit about where you got it from (or rewrite it in your own words, which at least suggests that you put a modest amount of thought into it).

I do think there's a case for allowing `<blockquote>` for including shorter sections of LLM output (up to one paragraph).

rfcs/llm_policy.md
jgraham: All PRs are created by one user account which almost always means one person. Do you have an example of a PR that's attributed to a team rather than an individual?

rfcs/llm_policy.md
annevk: E.g., the LLM whipped up a comparison table to illustrate some design trade-offs. It just seems excessive to have a "must" on this.

rfcs/llm_policy.md
annevk: That part makes sense, but I don't think the human uploading always knows about every line, if they are open sourcing a set of tests for instance that were developed internally over some period of time. E.g., I think some Opera contributions in the past were of this nature.

rfcs/llm_policy.md
jgraham: That feels like a case where it's very useful to know that it was LLM output that was copied and pasted.

rfcs/llm_policy.md
jgraham: That seems like a case where we're trusting a third party repository as a source of commits (as we do for Gecko, Chromium, WebKit and Servo repositories today). It is true the policy there isn't quite made explicit anywhere, but I think we should deal with that separately compared to the policy for direct contributions to this repository.

rfcs/llm_policy.md
annevk: I personally don’t really mind where it comes from as long as a human pasted it and knows it’s correct. If they are unsure, then yes.

Being overtly strict doesn’t seem like the right trade-off. Especially since I don’t think we even have that many bad contributions? They first have to figure out how to get reviewed. 😅

rfcs/llm_policy.md
gsnedders: > Do you have an example of a PR that's attributed to a team rather than an individual?

Maybe irrelevant detail, but you can create PRs whose head points at an organization's fork, though it is still a user account actually opening the PR.

rfcs/llm_policy.md
gsnedders: While I think the (2) might be reasonable to class as "low value", I think (1) is really a different failure mode, and one that becomes a higher-risk with LLMs (by virtue of sheer volume).

rfcs/llm_policy.md
jugglinmike: @annevk I think such an exception would probably have to be a little more precise than "a tiny snippet." That might also help folks engage with the way you're considering risks. What quantity of generated text needs a label?

rfcs/llm_policy.md
jugglinmike: @annevk If you're satisfied by James' latest response, could you mark this as resolved?

rfcs/llm_policy.md
jugglinmike: It is not! I've removed it.

rfcs/llm_policy.md
annevk: I'm not sure I'm entirely satisfied, but I'm also not sure it matters much. I don't think we have to necessarily have a relationship with a third-party repository for someone to contribute tests from a third-party repository. Like if someone had an independent implementation of the URL standard and they and other contributors created a bunch of tests for it, I would not want to enforce that the person upstreaming those tests can defend and speak to them all themselves.

rfcs/llm_policy.md
annevk: If you have a human written summary and want to defend that with a bunch of LLM text then I think it needs a label and it probably helps if it's hidden by default as there's not a lot of care put into it. But if you create a response that stands on its own it's probably fine. I'm not sure we have enough experience to judge one way or another.

jugglinmike:
My understanding is that while we expect this policy may evolve with experience, there are no objections to implementing it in its current form. To that end, @jgraham or @gsnedders: could you formalize your approval so that the acceptance countdown can begin?

rfcs/llm_policy.md
jugglinmike: I agree that (1) is distinct from (2), but I still don't see how it represents a meaningfully-distinct failure mode of this policy. Both (1) and (2) are second-order risks from defining a policy. They both flow from the premise that any such policy will encourage the use of tools that have these weaknesses. And they both have the same impact on our risk calculation here: we can't technically stop people, but an explicit policy formalizes a social contract.

That's what's keeping me from creating a new sub-section for over-fitting. It's easy to frame that as a separate risk, but it seems like we can only come to the same conclusion as we do for the "low-value contribution" risk: we accept it because we expect people to act in good faith and follow the policy.

rfcs/llm_policy.md
```diff
@@ -0,0 +1,95 @@
...
+> #### Disclosure
+>
+> If LLMs are used as a significant input to a commit, authors are encouraged
+> to include details about how they were used as part of the commit message in
+> order to help review and future understanding of the code.
+>
+> Human-authored code discourse (e.g. issue descriptions, pull request
+> descriptions, and responses to discussion threads) should not include
+> LLM-generated content in the main text; any such content must be clearly
+> labelled and placed inside a `<details>` element.
```
jgraham: I think perhaps we should allow "blockquote" for short snippets (say: up to one paragraph).

jgraham (approved):

rfcs/llm_policy.md
jugglinmike: Done!

jgraham:
@gsnedders are your concerns still blocking here?

jcscottiii (approved):

rfcs/llm_policy.md
gsnedders: I think this might be slightly too strict, but on balance I think having a policy is better than not — let's keep this as is, and we can revisit if people find it onerous.

gsnedders (approved):

PR #241, opened 2026-06-09
Codify the `requestAnimationFrame` delay in `test-wait.js` as the test-completion timing

gsnedders:
[Rendered](https://github.com/gsnedders/rfcs/blob/test-wait-double-rAF/rfcs/test-wait-double-rAF.md)

### DIFF
--- a/rfcs/test-wait-double-rAF.md
+++ b/rfcs/test-wait-double-rAF.md
@@ -1,0 +1,236 @@
+# RFC: Codify the `requestAnimationFrame` delay in `test-wait.js` as the test-completion timing
+
+## Summary
+
+Codify the double `requestAnimationFrame` call in wptrunner's `test-wait.js` as the test-completion timing,
+updating the documentation to match the implementation.
+
+This delay was added in 2017 as a short-term workaround for a Chromium rendering bug,
+but the discrepancy between the documented and actual behaviour has caused recurring flakiness in test runners that implement the documented (no-delay) semantics.
+
+## Background
+
+### Documentation
+
+The [reftest documentation][reftests-docs] describes the screenshot timing as follows:
+
+> By default, reftest screenshots are taken after the following
+> conditions are met:
+>
+> * The `load` event has fired
+> * Web fonts (if any) are loaded
+> * Pending paints have completed
+>
+> In some cases it is necessary to delay the screenshot later than this,
+> for example because some DOM manipulation is required to set up the
+> desired test conditions. To enable this, the test may have a
+> `class="reftest-wait"` attribute specified on the root element. In
+> this case the harness will run the following sequence of steps:
+>
+> * Wait for the `load` event to fire and fonts to load.
+> * Wait for pending paints to complete.
+> * Fire an event named `TestRendered` at the root element, with the
+>   `bubbles` attribute set to true.
+> * Wait for the `reftest-wait` class to be removed from the root
+>   element.
+> * Wait for pending paints to complete.
+> * Screenshot the viewport.
+
+The [crashtest documentation][crashtests-docs] describes similar but distinct behaviour:
+
+> The simplest crashtest is a single HTML file with any content. The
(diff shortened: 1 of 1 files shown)


jcscottiii:
I talked to Xianzhu about this. They said:

> I agree with https://issues.chromium.org/issues/41311503#comment3 that the WebDriver should be able to reliably capture an image reflecting the document's latest status, without the client needing to do anything.

> I don't remember many details, but [crrev.com/c/4505400](http://crrev.com/c/4505400) now looks like a temporary solution to make the two web test runners do the same thing. It was probably a workaround for the issue where some ref tests had [double-]rAF etc., which required waiting, but the tests didn't declare `reftest-wait`, causing flaky premature finishes (e.g. [crbug.com/40899923](http://crbug.com/40899923), [crbug.com/](http://crbug.com/) [crrev.com/c/40901018](http://crrev.com/c/40901018)). Now I think we should find and fix the tests instead of working around them in the test runner.

gsnedders:
> I agree with https://issues.chromium.org/issues/41311503#comment3 that the WebDriver should be able to reliably capture an image reflecting the document's latest status, without the client needing to do anything.

But there's not really any good way to define this — nothing really specifies when a frame gets displayed to the user — and that makes it hard to update the WebDriver spec to require this.

The reality is any executor implementation using WebDriver is going to have a notably longer delay than one within the browser process, and I'd expect having a longer delay than is necessary would make it more likely that it has settled fully.

The challenge with removing it is we basically have to run every reftest numerous times to validate that we aren't adding new instability into the suite, noting that that could easily depend on hardware performance. I'm not complete opposed to trying — but there is definitely risk here.

Per discussion with @jgraham in the infra meeting earlier in the week, there's also concerns about changing the internal Gecko implementation (which just waits for `MozAfterPaint`).

There's a few bits of experimentation we can do:

1. We could try removing the double rAF from wptrunner, and run full runs of Chrome and Safari with it, and see how many tests it changes the results of.
1. We could try adding the double rAF to WebKit, and see how many tests it changes the results of.
2. We could try removing the double rAF from `content_shell`, and see how many tests it changes the results of.
3. We could make the Firefox internal reftest implementation wait for MozAfterPaint or the second rAF, whichever comes later, and see how many tests it changes the results of.

None of these are foolproof, and anything that moves the timing forward (1, 3) has a risk of exposing more flakiness we don't yet know about.

gsnedders:
> * We could try adding the double rAF to WebKit, and see how many tests it changes the results of.

https://github.com/WebKit/WebKit/pull/71591

Ms2ger:
> > * We could try adding the double rAF to WebKit, and see how many tests it changes the results of.
> 
> [WebKit/WebKit#71591](https://github.com/WebKit/WebKit/pull/71591)

@gsnedders have you had a chance to look at the results?

gsnedders:
> > > * We could try adding the double rAF to WebKit, and see how many tests it changes the results of.
> > 
> > 
> > [WebKit/WebKit#71591](https://github.com/WebKit/WebKit/pull/71591)
> 
> @gsnedders have you had a chance to look at the results?

I think the only relevant change is [`imported/w3c/web-platform-tests/css/css-view-transitions/no-painting-while-render-blocked.html`](https://github.com/WebKit/WebKit/blob/07d83d0d699eedc4e80908d99470a5fca06acdc8/LayoutTests/imported/w3c/web-platform-tests/css/css-view-transitions/no-painting-while-render-blocked.html), which in most runs has gone from PASS to FAIL, which does admittedly match what wpt.fyi gets: https://wpt.fyi/results/css/css-view-transitions/no-painting-while-render-blocked.html?run_id=5079123960791040&run_id=5093206386802688&run_id=6272330065248256&run_id=4920473405358080

That said, I think the WebKit was the least likely to show up results — delaying the timing is less likely to change results, because it's mostly likely to just make things that were flaky now non-flaky, and we (hopefully) have them marked as flaky already.

PR #242, opened 2026-07-29
RFC 242: Wide-gamut and HDR screenshot reftests

Minju-kim-igalia:
This RFC proposes opt-in `reftest-color-space` metadata for wide-gamut and HDR screenshot reftests.

For opted-in tests, `wptrunner` requests PNG output in the requested color space through WebDriver BiDi and compares the decoded samples without converting them to sRGB. The same request applies to the root test and its reference graph.

The WebDriver BiDi protocol extension is tracked separately in w3c/webdriver-bidi#1114. The intended first end-to-end target is `display-p3` / `image/png`.

For implementation feasibility, I also uploaded a small Chromium WIP CL stack for that initial path:

- [Add a destination color space to Viz copy output](https://chromium-review.googlesource.com/c/chromium/src/+/8174224)
- [Prototype top-level Display-P3 screenshot capture](https://chromium-review.googlesource.com/c/chromium/src/+/8181686)

The second CL depends on the first. These CLs are supporting feasibility prototypes rather than final landing proposals. The WebDriver BiDi contract and the upstream `chromium-bidi` and WPT implementation changes remain separate work.

Related:
- web-platform-tests/wpt#44320
- web-platform-tests/interop-jpegxl#2
- web-platform-tests/interop-jpegxl#5

### DIFF
--- a/rfcs/wide_gamut_hdr_screenshot_reftests.md
+++ b/rfcs/wide_gamut_hdr_screenshot_reftests.md
@@ -1,0 +1,74 @@
+# RFC ##: Wide-Gamut and HDR Screenshot Reftests
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


Ms2ger (approved):
Seems like a reasonable approach

PR #243, opened 2026-08-18
Change test status race conditions in design of web_extensions.md

justinlulejian:
Updated the RFC to mandate that browser.test status listeners are registered before the test extension is loaded. This eliminates test result race conditions caused by cross-browser variations in extension loading speeds, and simplifies browser implementations by removing the need for complex queuing logic.

We are aware that Safari will need to adjust its implementation to accommodate this change. Safari is currently unable to run WPT tests unless the test status listeners are registered after the extension has loaded. However, we strongly believe that removing the queuing requirement is the best long-term choice for a simpler, improved implementation across all browsers. When designing this system, it is important to consider not only current implementations but also future browsers that want to participate in WPT testing, ensuring their onboarding is as straightforward as possible.


justinlulejian:
@rdcronin @oliverdunk for initial review

justinlulejian:
WECG review: @kiaraarose @xeenon @zombie @Rob--W @rdcronin

gsnedders:
See also: https://github.com/web-platform-tests/wpt/issues/60670

jgraham:
We typically don't update RFCs in-place but file new RFCs, although I see why you wanted to in this case since it's a small change to the implementation requirements. However ideally the actual documentation would be somewhere other than the RFC.

justinlulejian:
@jgraham does that mean that it would be more ideal to strip the RFC of implementation details and instead refer out to a separate document that has implementation details for the RFC? If so, are there any examples we can follow of this?

justinlulejian:
Ping for @jgraham.

Rob--W:
If we cannot update the RFC doc, would it make sense to include relevant changes in the test API doc, that you (Justin)  are already updating at https://github.com/w3c/webextensions/pull/1057 ?

